PR #460, opened 2017-08-09
PP-2006 - Standardised and clarified validation messages

jonheslop:
and added validation to forgotten password page


jonheslop:
rebuild

PR #468, opened 2017-08-11
PP-2497 decrease docker image size

maxcbc:
## WHAT
Switching to a 2 image build process.

1. Build the first image, which installs all test deps
2. Run the first image with workspace mounted as a volume, the `CMD` then copies in the deps, compiles the assets, runs the tests and finally prunes the deps ready for deployment
3. Build the 2nd image, which just copies the project into itself (including the node_modules and compiled assets obtained from running the first image).

## WHY
* Results in a build image with a variable layer size of ~5MB (assuming no changes to package.json), as only the `npm test` and `npm run compile` commands are fresh layers between most builds.
* Results in a deployed images ~90MB instead to the current of <600MB, with a variable layer size (the layer that changes each build and thus needs to be pulled each time) of ~40MB.

This is because we don't have layers that include test/build only dependencies in the 2nd image. We have a v.thin image that just includes the files and dependencies necessary to run on an environment.


maxcbc:
Updated commit messages

Jenkinsfile
```diff
@@ -9,7 +9,7 @@ pipeline {
   }
 
   libraries {
-    lib("pay-jenkins-library@master")
+    lib("pay-jenkins-library@PP-2497_decrease_frontend_image_size")
```
rauligs: So I believe this needs to be changed now?

PR #486, opened 2017-09-07
PP-2499 sign in and new account fixes

jonheslop:
There were loads of inconsistencies and inefficiencies with the way the grid was marked up and styled. This clears out some unnecessary declarations in favour of the default stuff.

Also added the propositional nav to the logged out states along with the beta badge to the header


jonheslop:
rebuild

PR #490, opened 2017-09-11
PP-2618 - Added the autofocus attributes to the input field

jonheslop:


jonheslop:
Thanks to @alexbishop1 for teaching me something new!

PR #496, opened 2017-09-20
PP-2575 remove .only from tests

maxcbc:
- remove .only erroneously committed in tests


maxcbc:
@jonheslop  I'm writing a grep command to fail builds if we commit them

jonheslop:
@maxcbc 👏

PR #509, opened 2017-10-02
PP-2446 product client

kakumara:
## CONTEXT
Product is a selfservice representation of the `Product` concept from products microservice. Its identical except for not including a `pay_api_key` (purposely).

## WHATS INCLUDED
- Similar to other external integrations introduced `products_client` which is a wrapper for `products` microservice interactions.
- `products_service.createProduct` orchestrates between `publicAuth` and `products` in order to create a product successfully
- The client of this method must make sure to perform necessary validations in `productOptions` to include all mandatory fields required

Other than that
  - added Pact based integration tests
  - added product_fixtures

## NOT DONE
- the controller and UI's for creating prototype payments in not included in this PR. (next to come)


app/models/Product.class.js
```diff
@@ -0,0 +1,47 @@
+'use strict'
+
+/**
+ @class Product
+ */
+class Product {
```
maxcbc: Noice

app/services/clients/products_client.js
```diff
@@ -0,0 +1,70 @@
+'use strict'
+
+const q = require('q')
```
maxcbc: We're switching to using native Promises on frontend for performance. 

We ought to do the same here. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

app/models/Product.class.js
kakumara: 👍

app/services/clients/products_client.js
kakumara: Agree. But we'll have to look at this later as changing this is a bit `convoluted` and causes all other client changes causing bit heavy recfactoring. Prehaps should do this as a separate effort. I'll adopt it in the new node microservice

maxcbc: OK

PR #510, opened 2017-10-02
[WIP] Nunjucks

jonheslop:
Mustache makes my dev life hard, also it's old and not that quick

Nunjucks is like Jinja and GDS like Jinja also the new GOV.UK Frontend thing coming soon uses Nunjucks because it's nice


jonheslop:
This was just me having a play, closing

PR #513, opened 2017-10-04
PP-2543 filter by multiple states

maxcbc:
# PP-2543 filter by multiple states

## WHAT
- places multiple state filter logic behind feature flag
- adds feature flag access in templates through change to `display_converter.js`
- adds basic support for multiple states via `multiple` attribute added to state `<select></select>`
- added support for multiple states in `filters.js`
- Implemented a multi-select dropdown as a progressive enhancement of a `<select></select>` that has a `data-enhance-multiple` attribute
- Added hoganify to our client-side js grunt-browserify config, this means that if you require a `.html` file browserify will pre-compile this for you via hogan.js.
- Updated state `<select></select>` on transactions filter view to have `data-enhance-multiple` attribute
- Updated filters description to embolden variables and not push out the results length toggle if it gets too long.
- Also updated tests where this had broken them


app/client-side-scripts/multi-select.js
```diff
@@ -0,0 +1,97 @@
+'use strict'
+// TODO: we should probably do some browser testing in this project to prove this all works as intended
```
mrlumbu: 😄

app/assets/sass/modules/_multi_select.scss
```diff
@@ -0,0 +1,102 @@
+
+
+div.multi-select {
+  position: relative;
+  @extend %contain-floats;
+  width:100%;
```
jonheslop: Pedantic, but there should be spaces between the `:` and the value i.e. `width: 100%`. There's a lot of occurrences in this file. There is a sass-lint file in the repo, if you use a plugin it will highlight these and you can prolly run an autofix thing

app/assets/sass/modules/_multi_select.scss
```diff
@@ -0,0 +1,102 @@
...
+    label {
+      padding: 10px 4px 4px;
+      width:100%;
+    }
+    input {
+      z-index: 12;
+      width:100%;
+    }
+  }
+  .multiple-choice [type=checkbox] + label::before,
```
jonheslop: should be `[type='checkbox']`

app/assets/sass/modules/_transactions_list.scss
```diff
@@ -11,3 +13,11 @@
     @include core-14;
   }
 }
+
+h2#total-results {
```
jonheslop: Don’t need the h2 (sass lint would complain about this too)

app/utils/filters.js
```diff
@@ -9,37 +9,40 @@ function validateFilters (filters) {
...
   return (pageSizeIsNull || pageSizeInRange) &&
-         (pageIsNull || pageIsPositive)
+    (pageIsNull || pageIsPositive)
 }
 
 function describeFilters (filters) {
   let description = ``
-  if (filters.fromDate) description += ` from ${filters.fromDate}`
-  if (filters.toDate) description += ` to ${filters.toDate}`
+  if (filters.fromDate) description += ` from <b>'${filters.fromDate}'</b>`
```
jonheslop: Is there a reason to use `<b>` as opposed to `<strong>` or something else?

test/ui/transaction_search_ui_tests.js
```diff
@@ -73,21 +74,27 @@ describe('The transaction list view', function () {
     }
 
     var body = renderTemplate('transactions/index', templateData)
+    var $ = cheerio.load(body)
 
-    body.should.containSelector('#download-transactions-link').withAttribute('href', templateData.downloadTransactionLink)
+    expect($('#download-transactions-link').attr('href')).to.equal(templateData.downloadTransactionLink)
+
+    const totalResultsText = $('h2#total-results').text()
```
jonheslop: That `h2` again

app/client-side-scripts/multi-select.js
```diff
@@ -0,0 +1,107 @@
...
+
+// Local Dependencies
+const multiSelect = require('../views/includes/multi-select.html')
+
+// Variables
+const MAXIMUM_VISIBLE_ITEMS = 5.5 // Maximum amount of items to show in dropdown
+// Selectors
+const DOCUMENT_SELECTOR = window.document
+const ENHANCEMENT_SELECTOR = 'select[data-enhance-multiple]'
+const TOP_LEVEL_SELECTOR = 'div.multi-select'
```
jonheslop: Don't use a qualifying tag name for selectors, it’s slows jQuery down. Make the class more specific is necessary

app/utils/filters.js
maxcbc: I've always used `<b>` for bold. No other reason than that.

jonheslop: `<b>` makes things bold
`<strong>` makes things bold and marks them as semantically important

Not sure which is more applicable in this case — https://www.html-5-tutorial.com/strong-and-b-elements.htm

jonheslop: I tend to use strong because its likely to be important if it needs to be in bold

maxcbc: I'm gonna leave it as `<b>` for now, we should bring this up next time we do an accessibility review. I'm not sure if they are more semantically important than the rest of the sentence.

test/ui/transaction_search_ui_tests.js
maxcbc: fixed

app/client-side-scripts/multi-select.js
maxcbc: fixed

app/assets/sass/modules/_multi_select.scss
maxcbc: fixed

PR #518, opened 2017-10-23
PP-2655 - Reusable client side form validation

jonheslop:
Usually we write validation into the controllers
which ends up meaning we have a lot of duplication
which is a shame and is loooong.

So this is a generic browserified validation engine.
To use it anywhere you need to do a couple of things

- add `data-validate="true"` to the form you want to validate
- add `data-validate` to each of the `input/select/textarea`'s you want validated
- by default it will check that the value is not blank but you can add a space separated list of validations e.g. `data-validate="required currency"`

Currently I have only written validators for `isCurrency`
and `isEmpty`. You can add new ones to `field-validation-checks.js`

The validator expects them to return a `string` with the error message or else `false`

### DIFF
--- a/app/browsered.js
+++ b/app/browsered.js
@@ -2,8 +2,10 @@
 // NPM dependencies
 const $ = require('jquery')
 const multiSelects = require('./client-side-scripts/multi-select')
+const fieldValidation = require('./browsered/field-validation')
 
 // This adds jquery globally for non-browserified contexts
 window.$ = window.jQuery = $
 
 multiSelects.enableMultiSelects()
+fieldValidation.enableFieldValidation()
--- a/app/browsered/field-validation-checks.js
+++ b/app/browsered/field-validation-checks.js
@@ -1,0 +1,23 @@
+'use strict'
+
+const validationErrors = {
+  required: 'This is field cannot be blank',
+  email: 'Please use a valid email address',
+  currency: 'Choose an amount in pounds and pence using digits and a decimal point. For example “10.50”'
+}
+
+module.exports.isEmpty = function (value) {
+  if (value === '') {
+    return validationErrors.required
+  } else {
+    return false
+  }
+}
+
+module.exports.isCurrency = function (value) {
+  if (!/^([0-9]+)(?:\.([0-9]{2}))?$/.test(value)) {
+    return validationErrors.currency
+  } else {
+    return false
+  }
+}
--- a/app/browsered/field-validation.js
+++ b/app/browsered/field-validation.js
@@ -1,0 +1,98 @@
+'use strict'
+
+// NPM Dependencies
+const lodash = require('lodash')
+
+// Local Dependencies
+const checks = require('./field-validation-checks')
+const validationErrorsTemplate = require('../views/includes/validation-errors.html')
+
+exports.enableFieldValidation = function () {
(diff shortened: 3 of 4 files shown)


jonheslop:
Things I wonder/worry about
- Have I got carried away and used `let` too much?
- Do in need the `for` loop [here](https://github.com/alphagov/pay-selfservice/pull/518/files#diff-251f310f43b3ec577b5dac8cdf6ca8f7R53)?
- Have a included it within `browsered.js` in the right way, do I need to use `module.export`?

app/browsered.js
```diff
@@ -7,3 +7,4 @@ const multiSelects = require('./client-side-scripts/multi-select')
 window.$ = window.jQuery = $
 
 multiSelects.enableMultiSelects()
+module.exports.fieldValidation = require('./browsered/field-validation')
```
maxcbc: I find it's generally better to export a function that has to be explicitly called rather than have a module change the global environment just by being required.
Only my opinion though.

app/browsered/field-validation.js
```diff
@@ -0,0 +1,96 @@
...
+  if (!formGroup.classList.contains('error')) {
+    formGroup.classList.add('error')
+    document.querySelector('label[for="' + field.name + '"]').insertAdjacentHTML('beforeend',
+      '<span class="error-message">' + result + '</span>')
+  }
+}
+
+function populateErrorSummary (form) {
+  let erroringFields = [...form.querySelectorAll('.form-group.error label')]
+  let configuration = {
```
maxcbc: not sure what this function does as it doesn't seem to alter an element on the page, or return anything

app/browsered/field-validation.js
```diff
@@ -0,0 +1,96 @@
...
+    return field.hasAttribute('data-validate')
+  })
+}
+
+function validateField (form, field) {
+  let result
+  let validationTypes = field.getAttribute('data-validate').split(' ')
+
+  for (let validationType of validationTypes) {
+    switch (validationType) {
```
maxcbc: should probably use an 'if'/'else' here as there are only 2 options

app/browsered/field-validation.js
```diff
@@ -0,0 +1,96 @@
...
+  return formFields.filter(field => {
+    return field.hasAttribute('data-validate')
+  })
+}
+
+function validateField (form, field) {
+  let result
+  let validationTypes = field.getAttribute('data-validate').split(' ')
+
+  for (let validationType of validationTypes) {
```
maxcbc: Could use Array.forEach(. Not sure. Just a thought. Ignore me.

app/browsered/field-validation-checks.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+
```
maxcbc: noice

app/browsered/field-validation.js
jonheslop: There will be loads eventually, I was just gonna and them as and when, for example isEmail, isLatinChars etc etc

app/browsered/field-validation.js
jonheslop: It renders the error summary above the inline errors.

app/browsered.js
jonheslop: Ok cool will change

PR #519, opened 2017-10-23
PP-2804 - Fixed inconsistencies in flash banner styles and placement

jonheslop:
The flash banner was placed in different places for layout and logged out layout
This caused style issues. I put it in the same place and then tweaked
spacing of elements it touches to make it more straight forward and
consistent when it is hidden or showing


maxcbc:
Can you add a 1 line description of where it's missing from please. Both to the PR comment and the commit message?

PR #520, opened 2017-10-23
PP-2800 - fixed missing icon

jonheslop:
The icon was missong from turn off email notifications screen

There was a load of CSS to replace it, that was no longer working, I don't know why so much was written when only the background-image: url() was causing the issue

There is a screenshot of the problem on Jira — https://payments-platform.atlassian.net/browse/PP-2800


maxcbc:
Can you add a 1 line description of where it's missing from please. Both to the PR comment and the commit message?

PR #522, opened 2017-10-24
PP 2655 Update Products Client

maxcbc:
# PP 2655 Update Products Client

## WHAT
- Renamed existing 'base_client' to 'old_base_client'
- Moved correlation_id middleware to separate file, added support for 'correlation_id' module
- Added new base_client and associated request wrapper from products_ui
- Added tests for base_client and wrapper

## WHY
New base client eliminates the need for:
- Passing correlation ids down function calls
- Calling log statements in every outgoing request
- Promisifying every call to request


maxcbc:
I think it is too much work for this PR. I will update the other clients in separate PRs.

PR #530, opened 2017-10-30
PP-2683 - Don't show gateway type if live

jonheslop:
Services only have one live account so it's not
necessary to show the account type. So added a
property to the object to let the template know
whether it is live or not.


mrlumbu:
is there a reason why we need the payment provider in test? Could we not just have sandbox|test|live or something?

jonheslop:
@mrlumbu I did question Sonia on why it wasn't necessary but she felt it was needless as no one ever does use more than one Gateway (which I worry could change in the future).

If you inspect the element however there is a class name that denotes whether it's worldpay/smartpay etc. So for testing purposes it can be derived which link is which

mrlumbu:
Ok - think this will need a bigger rethink at some point anyway depending on how we do direct debit reporting, and when we introduce new payment methods etc. This seems fine for now

PR #531, opened 2017-10-30
PP-2517 - Improved account creation validation messaging

jonheslop:
Error messaging about whether a password was strong enough
wasnt very helpful and was inconsistent. I've made it
explicitly tell you whether your password is too simple or
too short.

Whilst doing this I have noticed that the email validation
was entirely a duplication. So I have stripped it back and
updated all applications of it.

I also added the new client side validation to the sign up
forms to give more immediate feedback to users.

### DIFF
--- a/app/browsered/field-validation-checks.js
+++ b/app/browsered/field-validation-checks.js
@@ -1,9 +1,16 @@
 'use strict'
 
+// local dependencies
+const emailValidator = require('../utils/email_tools.js')
+
+// Constants
+const NUMBERS_ONLY = new RegExp('^[0-9]+$')
+
 const validationErrors = {
   required: 'This is field cannot be blank',
-  email: 'Please use a valid email address',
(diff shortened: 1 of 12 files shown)


app/browsered/field-validation-checks.js
```diff
@@ -1,9 +1,16 @@
 'use strict'
 
+// local dependencies
+const emailValidator = require('../utils/email_tools.js')
+
+// Constants
+const NUMBERS_ONLY = new RegExp('^[0-9]+$')
```
maxcbc: 👍

app/utils/email_tools.js
```diff
@@ -1,47 +1,13 @@
...
-      } else {
-        let domain = match[7]
-        return !(domain && domain.indexOf('.') === -1)
-      }
-    }
+module.exports = (email) => {
+  if (!rfc822Validator(email)) {
+    return false
+  } else {
+    let domain = email.split('@')[1]
```
maxcbc: 👍

PR #535, opened 2017-11-01
[REVIEW ONLY] PP-2688 Add edit merchant details form

simad:
## WHAT
This PR introduces a new form in the self-service settings, thanks to which a service user can specify the "Merchant Details". These details are: `name, address, city, postcode and country`.

The PR can be merged after https://github.com/alphagov/pay-adminusers/pull/142, which introduces the update merchant details endpoint in Adminusers.

## HOW 
- Create an admin users, and access the self service
- Go to the settings, there will be a "Merchant details" tab on the left
- The merchant details page will contain a form. All fields are mandatory, except the address line 2. 
- Submit valid details
- The form will be pre-populated with the chosen details and they will be visible in the frontend, at the bottom of the payment journey pages


app/controllers/edit_merchant_details_controller.js
```diff
@@ -0,0 +1,89 @@
...
+const lodash = require('lodash')
+
+// Local Dependencies
+const responses = require('../utils/response')
+const paths = require('../paths')
+const serviceService = require('../services/service_service')
+const ukPostcode = require('uk-postcode')
+const countries = require('../services/countries.js')
+
+const merchantNameField = 'merchant-name'
```
maxcbc: Generally we've been doing these sort of top-level `const` declarations in 'SCREAMING_SNAKE_CASE'.

Not sure this is a must.

app/controllers/edit_merchant_details_controller.js
```diff
@@ -0,0 +1,89 @@
...
+    address_line1: req.body[addressLine1Field],
+    address_city: req.body[addressCityField],
+    address_postcode: req.body[addressPostcodeField],
+    address_country: req.body[addressCountryField]
+  }
+  if (req.body[addressLine2Field]) {
+    requestPayload.address_line2 = req.body[addressLine2Field]
+  }
+  const errors = isValidForm(req)
+  if (lodash.isEmpty(errors)) {
```
maxcbc: This as far as I can tell will be an empty Object (i.e. `{}`) if there are no errors. If that is the case you can just do `!errors` as an empty Object is a falsey value.

test/unit/controller/edit_merchant_details_controller/edit_merchant_details_controller_test.js
```diff
@@ -0,0 +1,139 @@
...
+const mockResponses = {}
+const mockServiceService = {}
+const random = require('../../../../app/utils/random')
+const merchantDetailsCtrl = proxyquire('../../../../app/controllers/edit_merchant_details_controller', {
+  '../utils/response': mockResponses,
+  '../services/service_service': mockServiceService
+})
+let req, res
+
+describe('when the request body is correct', () => {
```
maxcbc: We can combine this test and the ui test as they duplicate similar behaviour. Can pair with you if you want on this.

app/controllers/edit_merchant_details_controller.js
simad: Blame javascript :D (or v8)

test/unit/controller/edit_merchant_details_controller/edit_merchant_details_controller_test.js
simad: Done! (Will cash in on the pairing offer another time :) )

app/controllers/edit_merchant_details_controller.js
```diff
@@ -0,0 +1,89 @@
...
+    errors[addressPostcodeField] = true
+  }
+  return errors
+}
+
+exports.post = (req, res) => {
+  const correlationId = lodash.get(req, 'correlationId')
+  const serviceExternalId = lodash.get(req, 'service.externalId')
+
+  const requestPayload = {
```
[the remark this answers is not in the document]
simad: 👍 we don't want to waste them memory

simad:
Addressed @maxcbc's wise comments

PR #537, opened 2017-11-06
PP-2688 Merchant details fixes

DanailMinchev:
## WHAT

When we were submitting merchant details (initial blank value) and there were a validation error, then the form data is not prefilled.
This PR fixes that.

### DIFF
--- a/test/ui/edit_merhant_details_ui_tests.js
+++ b/test/ui/edit_merhant_details_ui_tests.js
@@ -1,0 +1,56 @@
+'use strict'
+
+// Custom dependencies
+const renderTemplate = require('../test_helpers/html_assertions').render
+const paths = require('../../app/paths')
+const countries = require('../../app/services/countries')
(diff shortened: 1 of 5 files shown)


test/ui/edit_merhant_details_ui_tests.js
```diff
@@ -0,0 +1,56 @@
+'use strict'
```
simad: It would be good to apply the same patterns as with the previous PR (see https://github.com/alphagov/pay-selfservice/pull/535). Ui tests are "merged" in `get_test.js` and `post_test.js`. You can use cheerio to assert on ui elements.
The test to add is that when you submit something, it stays in the session (and that the form contains the session values after a post).

DanailMinchev: Thanks @simad . Addressed in [a113be0]

PR #545, opened 2017-11-13
PP-2543 - Multi select browser bug fixes and tweaks

jonheslop:
* Edited label to make it clearer it's a multi-select
* Moved to browsered as name is clearer
* Capitalise first letter of filter states, couldnt do with CSSas didnt work in FF
* removed merchant details as view-only users dont get to see that
* only constrain dropdownn height if there isnt enough room for it
* Updated aria and accessibility labels to make it easier for screen reader users how to use the multi-select


kakumara:
LGTM 👍

PR #548, opened 2017-11-14
PP-2655 Restrict demo and prototype payments to sandbox

maxcbc:
# PP-2655 Restrict demo and prototype payments to sandbox

## What
- Restricted demo and prototype payments to sandbox accounts
- Created `restrict_to_sandbox` middleware
- Added tests for new middleware


maxcbc:
The middleware test suite has a negative test. I was unsure as to whether we should add it to each controller or not. Can do, its no biggy

PR #550, opened 2017-11-14
A small fix to change pounds to pence before submitting it to products

SandorArpa:
## WHAT
`publicapi` expects pence as payment amount

## HOW 
replace `.` with no space before posting it to products

### DIFF
--- a/app/controllers/make_a_demo_payment/confirm_controller.js
+++ b/app/controllers/make_a_demo_payment/confirm_controller.js
@@ -15,6 +15,8 @@ module.exports = (req, res) => {
   let gatewayAccountId = auth.getCurrentGatewayAccountId(req)
 
   if (protoData) {
+    let paymentAmount = Math.trunc(protoData.protoPaymentAmount * 100)
+
     let params = {
       indexPage: paths.user.loggedIn,
       name: protoData.protoPaymentDescription,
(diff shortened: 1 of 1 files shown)


kakumara:
is the description correct. As far as I know selfservice does not interact with public api. I believe you mean products ???

app/controllers/make_a_demo_payment/confirm_controller.js
```diff
@@ -15,6 +15,8 @@ module.exports = (req, res) => {
   let gatewayAccountId = auth.getCurrentGatewayAccountId(req)
 
   if (protoData) {
+    let paymentAmount = protoData.protoPaymentAmount.replace(new RegExp('\\.', 'g'), '')
```
kakumara: Is this correct? Shouldn't you be multiplying the `protoData.protoPaymentAmount` * 100 ? 
Say if I type `10.000` it still means I expect £10 (1000 pence) not £100 (10000 pence)

SandorArpa: it is checked in previous controller before sent to this one

app/controllers/make_a_demo_payment/confirm_controller.js
```diff
@@ -15,6 +15,12 @@ module.exports = (req, res) => {
   let gatewayAccountId = auth.getCurrentGatewayAccountId(req)
 
   if (protoData) {
+    let paymentAmount = protoData.protoPaymentAmount * 100
```
kakumara: We should have a test for this

app/controllers/make_a_demo_payment/confirm_controller.js
```diff
@@ -15,6 +15,12 @@ module.exports = (req, res) => {
...
+    let index = protoData.protoPaymentAmount.indexOf('.')
+    if (index !== -1) {
```
kakumara: Once again, this smells of trouble.

SandorArpa:
Fix will be included in a different PR with relevant tests

PR #551, opened 2017-11-14
PP-2604 Remove vendored chamber & create single entry point

rhowe-gds:
## WHAT
Remove the vendored chamber binary and provide a single 'intelligent' entry point for the container, detecting whether or not to use chamber based on whether or not we're under ECS.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,4 +15,4 @@ ADD . /app
 
 RUN ln -s /tmp/node_modules /app/node_modules
 
-CMD npm start
+CMD docker-startup.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM node:6.11.1-alpine
 
+ARG CHAMBER_URL=https://github.com/segmentio/chamber/releases/download/v1.9.0/chamber-v1.9.0-linux-amd64
+
 RUN apk update &&\
     apk upgrade &&\
     apk add --update bash python make g++ libc6-compat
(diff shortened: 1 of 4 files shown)


update-chamber
```diff
@@ -0,0 +1,28 @@
...
+
+chamber_version=${1?Usage: $0 <chamber version>}
+
+chamber_url="https://github.com/segmentio/chamber/releases/download/v${chamber_version}/chamber-v${chamber_version}-linux-amd64"
+
+# Download & calculate the SHA-256 checksum of this version of chamber
+curl --fail --silent -L "$chamber_url" | shasum -a 256 | sed -e 's#-$#bin/chamber#' > chamber.sha256sum
+
+# Rewrite the download URL in the dockerfile
+sed -i '' -e "/^RUN .*wget.*chamber/s#https://.*#${chamber_url}#" Dockerfile
```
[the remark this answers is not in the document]
rhowe-gds: Good spot

Dockerfile
```diff
@@ -15,4 +15,10 @@ ADD . /app
 
 RUN ln -s /tmp/node_modules /app/node_modules
 
-CMD npm start
+RUN apk add openssl && \
+    mkdir -p bin && \
+    wget -qO bin/chamber https://github.com/segmentio/chamber/releases/download/v1.9.0/chamber-v1.9.0-linux-amd64 && \
```
DanailMinchev: Should we set `chmod +x bin/chamber` here or it is not needed?

rhowe-gds: I'm pretty sure it is! Done

rhowe-gds:
@london-coder Good idea, although I think in this case the ARG should just be the whole URL. Change pushed.

PR #552, opened 2017-11-14
Pp 2755 Removed syntactic sugar that has been used in field validation

ckalista:
## WHAT

- Removed syntactic sugar that has been used in the field validation that is unfortunately only supported in the more recent version of browsers which causes problems during end to end testing with our version of PhantomJs. The following has been replaced with a more backward compatible alternative:

```
- Spread: `[... ]` replaced by equivalent `Array.prototype.slice.call()`
- let... of replaced by equivalent `forEach`
```

- Added `id` attributes to certain elements to facilitate E2E.

### DIFF
--- a/app/browsered/field-validation.js
+++ b/app/browsered/field-validation.js
@@ -33,15 +33,15 @@ function initValidation (e) {
 }
 
 function clearPreviousErrors () {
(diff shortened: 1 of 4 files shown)


maxcbc:
It's odd that this fails as our client side scripts are browserfied. Maybe our browserfy config is wrong.

PR #553, opened 2017-11-14
PP-2655 Added unit tests for make_demo_payment controllers

maxcbc:
# PP-2655 Added unit tests for make_demo_payment controllers

## WHAT
- refactored confirm page so as to be a GET using session variables only
- added validation to index page on POST copying validation from prototype links pages
- added tests for index, confirm and edit pages

### DIFF
--- a/app/controllers/make_a_demo_payment/confirm_controller.js
+++ b/app/controllers/make_a_demo_payment/confirm_controller.js
@@ -15,6 +15,12 @@ module.exports = (req, res) => {
   let gatewayAccountId = auth.getCurrentGatewayAccountId(req)
 
   if (protoData) {
+    let paymentAmount = protoData.protoPaymentAmount * 100
+    let index = protoData.protoPaymentAmount.indexOf('.')
+    if (index !== -1) {
+      paymentAmount = paymentAmount.substr(0, index)
+    }
+
     let params = {
       indexPage: paths.user.loggedIn,
       name: protoData.protoPaymentDescription,
@@ -35,7 +41,7 @@ module.exports = (req, res) => {
         payApiToken: publicAuthData.token,
         gatewayAccountId,
         name: protoData.protoPaymentDescription,
-        price: protoData.protoPaymentAmount
+        price: paymentAmount
       }))
       .then(product => {
         params.prototypeLink = lodash.get(product, 'links.pay.href')
--- a/app/controllers/make_a_demo_payment/confirm_controller.js
+++ b/app/controllers/make_a_demo_payment/confirm_controller.js
@@ -11,47 +11,37 @@ const publicAuthClient = require('../../services/clients/public_auth_client')
 const auth = require('../../services/auth_service.js')
 
 module.exports = (req, res) => {
-  let protoData = lodash.get(req, 'session.pageData.protoData')
-  let gatewayAccountId = auth.getCurrentGatewayAccountId(req)
+  const gatewayAccountId = auth.getCurrentGatewayAccountId(req)
+  const {paymentAmount, paymentDescription} = lodash.get(req, 'session.pageData.makeADemoPayment')
 
-  if (protoData) {
-    let paymentAmount = protoData.protoPaymentAmount * 100
-    let index = protoData.protoPaymentAmount.indexOf('.')
-    if (index !== -1) {
-      paymentAmount = paymentAmount.substr(0, index)
-    }
-
-    let params = {
-      indexPage: paths.user.loggedIn,
-      name: protoData.protoPaymentDescription,
-      price: protoData.protoPaymentAmount,
-      gateway_account_id: gatewayAccountId
-    }
-
-    publicAuthClient.createTokenForAccount({
-      accountId: gatewayAccountId,
-      correlationId: req.correlationId,
-      payload: {
-        account_id: gatewayAccountId,
-        created_by: req.user.email,
-        description: `Token for Prototype: ${req.body['payment-description']}`
-      }
-    })
-      .then(publicAuthData => productsClient.product.create({
-        payApiToken: publicAuthData.token,
-        gatewayAccountId,
-        name: protoData.protoPaymentDescription,
-        price: paymentAmount
-      }))
-      .then(product => {
-        params.prototypeLink = lodash.get(product, 'links.pay.href')
-        return response(req, res, 'dashboard/demo-payment/confirm', params)
-      })
-      .catch(error => {
-        req.flash('genericError', `<h2>There were errors</h2> ${error}`)
-        return res.redirect(paths.prototyping.demoPayment.index)
-      })
-  } else {
-    res.redirect(paths.prototyping.demoPayment.index)
+  if (!paymentAmount || !paymentDescription) {
+    return res.redirect(paths.prototyping.demoPayment.index)
   }
+
+  publicAuthClient.createTokenForAccount({
+    accountId: gatewayAccountId,
+    correlationId: req.correlationId,
+    payload: {
+      account_id: gatewayAccountId,
+      created_by: req.user.email,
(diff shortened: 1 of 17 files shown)


test/unit/controller/make_a_demo_payment_controller/confirm_controller_test.js
```diff
@@ -0,0 +1,200 @@
...
+    describe(`when the API token is successfully created`, () => {
+      describe(`and the product is successfully created`, () => {
+        let result, $, session
+        before(done => {
+          nock(PUBLIC_AUTH_URL).post('', VALID_CREATE_TOKEN_REQUEST).reply(201, VALID_CREATE_TOKEN_RESPONSE)
+          nock(PRODUCTS_URL).post('/v1/api/products', VALID_CREATE_PRODUCT_REQUEST).reply(201, VALID_CREATE_PRODUCT_RESPONSE)
+          session = getMockSession(VALID_USER)
+          lodash.set(session, 'pageData.makeADemoPayment.paymentDescription', PAYMENT_DESCRIPTION)
+          lodash.set(session, 'pageData.makeADemoPayment.paymentAmount', PAYMENT_AMOUNT)
+          supertest(createAppWithSession(getApp(), session))
```
heathd: I can see why you've put this in the before() so that you can have separate tests making assertions about different aspects of the expected behaviour in this circumstance.

However, I would like if at all possible to follow the [ArrangeActAssert](http://wiki.c2.com/?ArrangeActAssert) pattern for tests, and in particular to avoid doing the 'act' step in `before` blocks. IMHO both "act" and "assert" should occur explicitly in each test. 

I would consider `createAppWithSession()` to be part of "arrange" but `get()` is "act".

maxcbc: I like this definition of Arrange, Act, Assert I'm not sure I agree with both "act" and "assert" being in each test (`it` block):
- putting the "act" in each `it` block would duplicate an awful lot code if you keep the separate tests for different aspects of the expected behaviour
- if you consolidate the `it` blocks, you end up with poor test feedback when you make future changes, as each `it` block ends up testing too much and you only see what assertion (i.e. `expected: true, actual: false`) has failed not what particular expected behaviour has failed to occur
- by having multiple asynchronous `it` blocks rather than handling all asynchronous code in a `before` block your tests can end up taking longer (particularly when doing something slower like pact testing)

If we really wanted to maintain separation between "arrange" and "act" an option could be to do two`before` blocks to effectively separate this code (arrange generally  is synchronous anyway). Effectively giving us:
`before(() => {...}) // Arrange`
`before(done => {...}) // Act`
`it(() => {...}) // Assert`

Mocha calls before blocks in order so this would work in practice.

app/controllers/make_a_demo_payment/confirm_controller.js
```diff
@@ -11,41 +11,37 @@ const publicAuthClient = require('../../services/clients/public_auth_client')
...
+      account_id: gatewayAccountId,
+      created_by: req.user.email,
+      description: `Token for Demo Payment`
+    }
+  })
+    .then(publicAuthData => productsClient.product.create({
+      payApiToken: publicAuthData.token,
+      gatewayAccountId,
+      name: paymentDescription,
+      price: paymentAmount * 100
```
kakumara: Isn't it safer to use a `Math.trunc()` or `Math.floor()` around the calculated payment amount. irrespective of what validations were taken care of by the caller modules.

maxcbc: Yeah, sorry raised the PR as a WIP. Was about to add this.

kakumara:
We might want to double check if we are doing the reverse transformation to £ (pounds) in 
https://github.com/alphagov/pay-selfservice/blob/0d6e693c042ca99e24838d4c298645574c3d733d/app/views/dashboard/demo-service/index.html#L53

app/controllers/make_a_demo_payment/confirm_controller.js
maxcbc: Actually I'm not sure it is, going by our existing validation model. 

If we we're to do this in the Node.js code, surely we should be doing this at the ProductsClient level. But by that logic, surely we should also be validating that it also a Number.

As this api expects an Integer for price, shouldn't the API return 400 if it's passed a float?

test/unit/controller/make_a_demo_payment_controller/confirm_controller_test.js
heathd: the aim is not to have each of arrange/act/assert is in a separate `it` block, but rather to have a pattern which clearly separates each of those things. In the c2 wiki they talk about having segments separated by a blank line. That would be sufficient. 

The important thing is that:
a) there's a consistent pattern used in all tests
b) the three parts are separated in a clear way

I do have a preference for having the 'act' and 'assert' parts within the `it` block. The reason for that preference is that when I'm reading a test, the first thing that I look for is *what* is being tested, and in order to find that I'm looking for the 'act' part. So I expect that to be clearly visible in each test step. 

Regarding the risk of duplication/repetition, that's absolutely fine. 

Test code should be [DAMP not DRY](https://stackoverflow.com/a/11837973).

app/controllers/make_a_demo_payment/confirm_controller.js
```diff
@@ -11,41 +11,37 @@ const publicAuthClient = require('../../services/clients/public_auth_client')
...
-      .catch(error => {
-        req.flash('genericError', `<h2>There were errors</h2> ${error}`)
-        return res.redirect(paths.prototyping.demoPayment.index)
+  publicAuthClient.createTokenForAccount({
+    accountId: gatewayAccountId,
+    correlationId: req.correlationId,
+    payload: {
+      account_id: gatewayAccountId,
+      created_by: req.user.email,
+      description: `Token for Demo Payment`
```
ckalista: This no needing substitution, we should probably use normal quotes instead of back tick.

app/controllers/make_a_demo_payment/index_controller.js
```diff
@@ -6,35 +6,38 @@ const lodash = require('lodash')
...
+  if (!paymentAmount || isCurrency(paymentAmount)) {
+    lodash.set(req, 'session.pageData.makeADemoPayment.paymentAmount', paymentAmount)
+    req.flash('genericError', `<h2>Use valid characters only</h2> ${isCurrency(paymentAmount)}`)
+    return res.redirect(paths.prototyping.demoPayment.editAmount)
   }
 
-  const currencyMatch = /^([0-9]+)(?:\.([0-9]{2}))?$/.exec(paymentAmount)
-
-  if (currencyMatch && !currencyMatch[2]) {
+  paymentAmount = paymentAmount.replace(/[^0-9.-]+/g, '')
```
ckalista: Good we fixed it!

app/controllers/test_with_your_users/submit_controller.js
```diff
@@ -32,12 +32,14 @@ module.exports = (req, res) => {
...
 
-  const amountFormatCheck = AMOUNT_FORMAT.exec(paymentAmount)
   if (lodash.get(req, 'session.flash.genericError.length')) {
     return res.redirect(paths.prototyping.demoService.create)
-  } else {
-    paymentAmount = parseInt(amountFormatCheck[1]) * 100
-    if (amountFormatCheck[2]) paymentAmount += parseInt(amountFormatCheck[2])
+  }
+
+  paymentAmount = paymentAmount.replace(/[^0-9.-]+/g, '')
```
ckalista: 👍

app/controllers/make_a_demo_payment/confirm_controller.js
maxcbc: done

PR #573, opened 2017-11-21
PP-2446 Added Max Amount Validator

maxcbc:
# PP-2446 Added Max Amount Validator

## WHAT
- Added Max Amount Validator and tests
- Added as inline client-side validator via html template
- Added server-side validator to demo and prototype controllers with amount inputs

### DIFF
--- a/test/unit/controller/test_with_your_users_controller/submit_controller_test.js
+++ b/test/unit/controller/test_with_your_users_controller/submit_controller_test.js
@@ -24,6 +24,9 @@ const VALID_PAYLOAD = {
   'payment-description': 'Test service name',
   'confirmation-page': 'https://www.example.com'
 }
+const VALID_MINIMAL_GATEWAY_ACCOUNT_RESPONSE = {
+  payment_provider: 'sandbox'
+}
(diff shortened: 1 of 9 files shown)


test/unit/controller/test_with_your_users_controller/submit_controller_test.js
```diff
@@ -188,44 +189,117 @@ describe('test with your users - submit controller', () => {
...
+        it('should redirect with a statusCode of 302', () => {
+          expect(result.statusCode).to.equal(302)
+        })
+        it('should redirect to the create prototype link page', () => {
+          expect(result.header).to.have.property('location').to.equal(paths.prototyping.demoService.create)
+        })
+        it('should add a relevant error message to the session \'flash\'', () => {
+          expect(session.flash).to.have.property('genericError')
+          expect(session.flash.genericError.length).to.equal(1)
+          expect(session.flash.genericError[0]).to.equal('<h2>Enter a valid amount</h2> Choose an amount under £10,000,000')
```
ckalista: Interesting point here, the assertion of the test relies on the local been setup correctly which I guess we can assume, so no biggy.

PR #582, opened 2017-11-30
PP-2993 migrate to nunjucks

jonheslop:
We’ve decided to migrate from Mustache to Nunjucks because Mustache is no
longer maintained or used by the Node.js community, Nunjucks on
the other hand is widely used. It’s also used widely within GDS
(and also shares it's syntax with Jinja2 which GDS Python projects use)

This first commit is a rebase an amalgamation of my initial spike
work to see how Nunjucks could work for Pay and what config was
needed to get things going.
- ca9c3f7 But basically, I’ve removed the old templating packages and added the Nunjucks package. I have also renamed and moved all the old 'token' templates into their own directory to be inline with all the other settings templates I also renamed API keys not 'dev tokens' as this was confusing as we don't talk about them as tokens anywhere but here.
- 8c7b7b4 Credentials templates
- c88cf42 Settings, My services and Transaction detail templates
- bff7ead Team member templates
- 8f86196 Dashboard and prototype link templates
- 01e5121 Forgot password templates
- 8cdb77b Merchant details template
- 095f529 Self create service templates
- 2c52384 User registration templates
- 06b7420 Misc. error and other templates
- ae11188 Adding Nunjucksify, so we can do dynamic stuff in browser (like error messaging)
- 8522185 Moving all the converted templates to `views`


Gruntfile.js
```diff
@@ -47,9 +47,12 @@ module.exports = function (grunt) {
           },
           {
             expand: true,
-            cwd: 'node_modules/govuk_template_mustache/',
+            cwd: 'node_modules/govuk_template_jinja/',
             src: '**',
-            dest: 'govuk_modules/govuk_template/'
+            dest: 'govuk_modules/govuk_template/',
+            rename: function (dest, src) {
```
maxcbc: can be `rename: (dest, src) => dest + src.replace('html', 'njk'),`

app/controllers/api_keys_controller.js
```diff
@@ -7,8 +7,8 @@ var auth = require('../services/auth_service.js')
 const publicAuthClient = require('../services/clients/public_auth_client')
```
maxcbc: Love the rename

app/views-nunjucks/api-keys/_api-keys.njk
```diff
@@ -0,0 +1,28 @@
+<div class="tab-content current">
+    <h2 id="{{header}}" class="heading-medium remove-top-margin">
+        {% if tokens.length %}
+            {% if tokens_singular %}
```
maxcbc: could be `if tokens.length === 1`

server.js
```diff
@@ -72,14 +76,39 @@ function initialiseProxy (app) {
...
-function initialiseAppVariables (app) {
-  app.set('view engine', 'html')
-  app.set('vendorViews', path.join(__dirname, '/govuk_modules/govuk_template/views/layouts'))
-  app.set('views', path.join(__dirname, '/app/views'))
-}
-
 function initialiseTemplateEngine (app) {
-  app.engine('html', require(path.join(__dirname, '/lib/template-engine.js')).__express)
+  // Define app views
+  const appViews = [
```
maxcbc: should be declared outside function scope a la products-ui

server.js
```diff
@@ -72,14 +76,39 @@ function initialiseProxy (app) {
...
+  const nunjucksConfiguration = {
+    express: app, // the express app that nunjucks should install to
+    autoescape: true, // controls if output with dangerous characters are escaped automatically
+    throwOnUndefined: false, // throw errors when outputting a null/undefined value
+    trimBlocks: true, // automatically remove trailing newlines from a block/tag
+    lstripBlocks: true, // automatically remove leading whitespace from a block/tag
+    watch: false, // reload templates when they are changed (server-side). To use watch, make sure optional dependency chokidar is installed
+    noCache: false // never use a cache and recompile templates each time (server-side)
+  }
+  if ((!nodeEnv) || (nodeEnv !== 'production')) {
```
maxcbc: should perhaps just be in the nunjuncs configuration:
```
nocache: NODE_ENV !== 'production',
watch: NODE_ENV !== 'production
```

maxcbc: Also if you inline these you could inline the whole config as you wouldn't need to declare `const nunjucksConfiguration`. You could probable inline the appViews then too.

server.js
```diff
@@ -29,6 +30,9 @@ const errorHandler = require(path.join(__dirname, '/app/middleware/error_handler
 // Global constants
 const port = (process.env.PORT || 3000)
 const unconfiguredApp = express()
+const nodeEnv = process.env.NODE_ENV
```
maxcbc: should probably be screaming snake case

nickcolley:
Woah mammoth work, so good 🚀 💥 👍 

thumbs up on a rocket

PR #586, opened 2017-12-04
PP-2967 Updating Docker packages

daniele-occhipinti:
Updating Docker packages


daniele-occhipinti:
rebuild

daniele-occhipinti:
@heathd I pulled the dependency explicitly from edge because the ones in 3.4 were vulnerable. So, before this change, those packages were already in the container, but at an older version (which had sec vuls). The change explicitly installs newer versions of those packages from edge.

PR #587, opened 2017-12-05
PP-2836 - Strip whitespace from email and phone numbers

jonheslop:
emails with leading/trailing whitespace were rejected, now we trim it so they're not

Also phone numbers with spaces in were rejected too and that seems a shame, no we strip them server side

### DIFF
--- a/app/controllers/register_service_controller.js
+++ b/app/controllers/register_service_controller.js
@@ -40,7 +40,7 @@ module.exports = {
   submitRegistration: (req, res) => {
     const correlationId = req.correlationId
     const email = req.body['email']
-    const telephoneNumber = req.body['telephone-number']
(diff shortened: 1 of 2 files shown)


app/controllers/register_service_controller.js
```diff
@@ -40,7 +40,7 @@ module.exports = {
   submitRegistration: (req, res) => {
     const correlationId = req.correlationId
     const email = req.body['email']
-    const telephoneNumber = req.body['telephone-number']
+    const telephoneNumber = req.body['telephone-number'].replace(/\s/g, '')
```
georgeracu: There is no trim() function for this? Or something similar? 😕

jonheslop: `trim()` just removes whitespace from either end, this removes all the whitespace

PR #591, opened 2017-12-11
PP-3039 Upgrades Node.js version in package.json as well as test/prod Dockerfiles

brid:
## WHAT
Upgrades Node.js in both the package.json 'engines' (if deployed to Heroku) as well as the build/test and production Dockerfiles to use the 6.12.2 release of Node.js which fixes an Open SSL vulnerability released on the 8th December.


maxcbc:
Also worth changing `.nvmrc`

PR #592, opened 2017-12-11
BAU: Fix logging

tlwr:
Getting a million lines of

```
warn: Google Analytics Tracking ID [ANALYTICS_TRACKING_ID] is not set
```

in logs.

@jonheslop

solo @tlwr


server.js
```diff
@@ -52,7 +58,6 @@ function initialiseGlobalMiddleware (app) {
     res.locals.asset_path = '/public/'
     res.locals.routes = router.paths
     if (typeof process.env.ANALYTICS_TRACKING_ID === 'undefined') {
```
maxcbc: this shouldn't be done in middleware really. And it should be on `app.local` rather than `res.local` as it is global to the express app and not just to this particular request.
Just set `app.locals.analyticsTrackingId = process.env.ANALYTICS_TRACKING_ID || ''` globally here
It can then be accessed in the request via `res.app.locals.analyticsTrackingId`

PR #595, opened 2017-12-12
bumped busybox from r6 to r7 as r6 no longer available on repo

maxcbc:
bumped busybox from r6 to r7 as r6 no longer available on repo


brid:
As an aside, it looks like using the edge repo means we are going to be chasing issues relating to package version mismatches on an ongoing basis.

PR #597, opened 2017-12-12
PP-2877 payment totals on dashboard

jonheslop:
- 0a20710 - Templates for activity dashboard
- 99adff2 - refactored login controller into separate files ready for the fact the loggedIn function will become the dashboard controller
- 72a9ade - Added in activity endpoints to connector client 
- 635009e - Custom Nunjucks filter for formatting currency 
- e0cd89a 
  - Updated connector client to make it easy to retrieve results
  - Added a controller to fetch and display recent transactions
  - Updated markup to replace placeholders for real variables
  - Wrote new date/time filter for Nunjucks so can convert from ISO
time strings to readable formats
  - Added in small JS to reload page on period change, but with
noscript fallback
  - fixed escaping in flash messaging
- 51fa6ee & a330041 - Nunjucks custom filter tests 
- 85c2b96 - moved dashboard controller to its own directory 
- 40630ed - wrote tests for dashboard activity controller with @maxcbc


app/controllers/login/index.js
```diff
@@ -0,0 +1,14 @@
+'use strict'
+
+exports.logout = require('./logout-controller')
```
DanailMinchev: I think we should use `module.exports` instead of `exports` as we always return `module.exports` from the module

app/views/dashboard/_activity.njk
```diff
@@ -0,0 +1,77 @@
...
+        <option {% if period == 'previous-thirty-days' %}selected{% endif %} value="previous-thirty-days">Previous 30 days</option>
+      </select>
+      <noscript>
+        <button class="button nojs-update-button" type="submit">Update</button>
+      </noscript>
+    </fieldset>
+  </form>
+</div>
+
+{% set transactions_period_string %}
```
DanailMinchev: Should we put that variable in the controller? Or it is actually "local view" variable and belongs more to the view logic?

app/views/includes/scripts.njk
```diff
@@ -1,2 +1,6 @@
 <script src="{{ js_path }}"></script>
-
+<script src="/public/javascripts/govuk/details.polyfill.js"></script>
+<script>
+  GOVUK.details.init()
+  console.log('hello')
```
DanailMinchev: Could you remove `console.log` from here

app/utils/nunjucks-filters/currency.js
```diff
@@ -0,0 +1,5 @@
+const currencyFormatter = require('currency-formatter')
+
+module.exports = amountInPence => {
```
DanailMinchev: Nice!

PR #599, opened 2017-12-14
PP-2963 - Switched phase tag for live from blue to grey

jonheslop:
Switched phase tag for live from blue to grey as its confusing as the beta tag above is blue

### DIFF
--- a/app/assets/sass/template/header.scss
+++ b/app/assets/sass/template/header.scss
@@ -2,10 +2,10 @@
   max-width: 1020px;
 }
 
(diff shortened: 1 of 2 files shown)


app/assets/sass/template/header.scss
```diff
@@ -2,10 +2,10 @@
   max-width: 1020px;
 }
 
-/* Allow the menu link to sit on the same line as the logo */
+// Allow the menu link to sit on the same line as the logo
```
DanailMinchev: Great!

PR #617, opened 2018-01-08
New analytics tracking function expected analytics functions to be present

jonheslop:
But on test analytics is purposefully not set up and so `ga` doesn’t exist so the click events triggered an error in the console.

Also updated dependencies in `dockerfile` to fix breaking build

### DIFF
--- a/app/views/dashboard/demo-service/index.njk
+++ b/app/views/dashboard/demo-service/index.njk
@@ -8,7 +8,7 @@ Test with your users - {{currentServiceName}} {{currentGatewayAccount.full_type}
   <a href="/" class="link-back">Back to dashboard</a>
   <h1 class="heading-large prototyping__heading">Test with your users</h1>
   <aside class="info-box">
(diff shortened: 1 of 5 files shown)


app/views/dashboard/demo-service/index.njk
```diff
@@ -8,7 +8,7 @@ Test with your users - {{currentServiceName}} {{currentGatewayAccount.full_type}
   <a href="/" class="link-back">Back to dashboard</a>
   <h1 class="heading-large prototyping__heading">Test with your users</h1>
   <aside class="info-box">
-    <p>Prototype link urls will be updated on 11 January 2018. You will need to create new links and update your prototypes.</p>
+    <p>Prototype link urls will be updated on 16 January 2018. You will need to create new links and update your prototypes.</p>
```
maxcbc: See comments on PP-3078. Actually should be 17th sorry.

maxcbc: Should infact be 16th after all

PR #618, opened 2018-01-09
New date for the the update

soniaturcotte:
## WHAT
Blue Badge and they have a service assessment on the 16th, which will require them to have a working prototype. We're going to wait until the 17th to make the smallest impact to the team.


jonheslop:
This will be done when #617 is merged

PR #619, opened 2018-01-12
PP-3048_add_update_service_name_to_products_client

maxcbc:
- added update service name to products client
    - added synchronised code with products client in self-service

### DIFF
--- a/app/services/clients/products_client.js
+++ b/app/services/clients/products_client.js
@@ -20,7 +20,7 @@ module.exports = {
   product: {
     create: createProduct,
     disable: disableProduct,
-    updateProductServiceName,
+    updateServiceNameOfProductsByGatewayAccountId: updateServiceNameOfProductsByGatewayAccountId,
     getByProductExternalId: getProductByExternalId,
     getByGatewayAccountId: getProductsByGatewayAccountId
   },
@@ -97,23 +97,23 @@ function getProductsByGatewayAccountId (gatewayAccountId) {
 }
 
 /**
- * @param {String} productExternalId: the external id of the product you wish to retrieve
+ * @param {String} gatewayAccountId: the id of the gateway account whose service name you wish to update
  * @returns {Promise<Product>}
  */
-function updateProductServiceName (productExternalId, serviceName) {
+function updateServiceNameOfProductsByGatewayAccountId (gatewayAccountId, serviceName) {
   return baseClient.patch({
     headers,
     baseUrl,
-    url: `/products/${productExternalId}`,
+    url: `/gateway-account/${gatewayAccountId}`,
     json: true,
     body: {
-      op: 'update',
+      op: 'replace',
       path: 'service_name',
       value: serviceName
     },
     description: `update a product's service name`,
     service: SERVICE_NAME
-  }).then(product => new Product(product))
+  })
 }
 
 /**
--- a/app/services/clients/products_client.js
+++ b/app/services/clients/products_client.js
@@ -20,6 +20,7 @@ module.exports = {
   product: {
     create: createProduct,
     disable: disableProduct,
+    updateProductServiceName,
     getByProductExternalId: getProductByExternalId,
     getByGatewayAccountId: getProductsByGatewayAccountId
   },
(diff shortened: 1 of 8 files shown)


app/services/clients/products_client.js
```diff
@@ -96,9 +96,29 @@ function getProductsByGatewayAccountId (gatewayAccountId) {
   }).then(products => products.map(product => new Product(product)))
 }
 
+/**
+ * @param {String} productExternalId: the external id of the product you wish to retrieve
+ * @returns {Promise<Product>}
+ */
+function updateProductServiceName (productExternalId, serviceName) {
```
kakumara: This looks wrong. its based on gateway account (not per product). So you need to pass the gateway account id and the new service name. The sync code will look identical to the connector name sync call.

maxcbc: damn you're right.

SandorArpa: he is always right... :(

maxcbc: sorted :-)

app/services/clients/products_client.js
```diff
@@ -96,9 +96,29 @@ function getProductsByGatewayAccountId (gatewayAccountId) {
...
+    url: `/gateway-account/${gatewayAccountId}`,
+    json: true,
+    body: {
+      op: 'update',
+      path: 'service_name',
+      value: serviceName
+    },
+    description: `update a product's service name`,
+    service: SERVICE_NAME
+  }).then(product => new Product(product))
```
kakumara: You won't get a product back in this case. technically you are not calling product resource. It is an implicit gateway account resource you are invoking. So there is no way to return a single product.

test/fixtures/product_fixtures.js
```diff
@@ -14,6 +14,33 @@ module.exports = {
...
+    }
+    return {
+      getPactified: () => {
+        return pactProducts.pactify(data)
+      },
+      getPlain: () => data
+    }
+  },
+
+  invalidUpdateProductServiceNameRequest: () => {
```
kakumara: its not really a product service name we are updating. Its more of a gateway account name

app/services/clients/products_client.js
```diff
@@ -96,9 +96,29 @@ function getProductsByGatewayAccountId (gatewayAccountId) {
   }).then(products => products.map(product => new Product(product)))
 }
 
+/**
+ * @param {String} gatewayAccountId: the external id of the product you wish to retrieve
```
SandorArpa: copy-paste?

app/services/clients/products_client.js
```diff
@@ -96,9 +96,29 @@ function getProductsByGatewayAccountId (gatewayAccountId) {
...
+ * @returns {Promise<Product>}
+ */
+function updateServiceNameOfProductsByGatewayAccountId (gatewayAccountId, serviceName) {
+  return baseClient.patch({
+    headers,
+    baseUrl,
+    url: `/gateway-account/${gatewayAccountId}`,
+    json: true,
+    body: {
+      op: 'update',
```
SandorArpa: op:replace, otherwise it will be rejected as `Operation update is not valid for path service_name`

test/fixtures/product_fixtures.js
```diff
@@ -14,6 +14,33 @@ module.exports = {
     pactProducts.pactify(opts)
   },
 
+  validUpdateServiceNameOfProductsByGatewayAccountIdRequest: (serviceName) => {
+    const data = {
+      op: 'update',
```
SandorArpa: same as above, it should be `replace`

app/services/clients/products_client.js
maxcbc: Ahh I see. Sorted

app/services/clients/products_client.js
maxcbc: Good spot

test/unit/clients/product_client/product/update_service_name_by_gateway_account_id_test.js
```diff
@@ -0,0 +1,137 @@
...
+  /**
+   * Remove the server and publish pacts to broker
+   */
+  after(done => {
+    mockServer.delete()
+      .then(() => pactProxy.removeAll())
+      .then(() => done())
+  })
+
+  describe('when a product is successfully disabled', () => {
```
kakumara: disabled ??? 😱

PR #620, opened 2018-01-15
PP-3048 - Added product service name update to edit service name in service_service.

maxcbc:
Updated service service to also update the service name in pay-products

### DIFF
--- a/app/services/service_service.js
+++ b/app/services/service_service.js
@@ -4,10 +4,11 @@ const q = require('q')
 const lodash = require('lodash')
 const getAdminUsersClient = require('./clients/adminusers_client')
 const {ConnectorClient} = require('../services/clients/connector_client')
+const productsClient = require('../services/clients/products_client')
 const GatewayAccount = require('../models/GatewayAccount.class')
 const Service = require('../models/Service.class')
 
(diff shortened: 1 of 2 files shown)


app/services/service_service.js
```diff
@@ -51,14 +52,14 @@ function updateServiceName (serviceExternalId, serviceName, correlationId) {
...
-        })).then(() => {
-          q.resolve(new Service(result))
-        })
-      } else {
+      if (gatewayAccountIds.length <= 0) {
         return q.resolve(new Service(result))
+      } else {
+        return q.all([
+          ...gatewayAccountIds.map(gatewayAccountId => connectorClient.patchServiceName(gatewayAccountId, serviceName, correlationId)),
+          ...gatewayAccountIds.map(gatewayAccountId => productsClient.product.updateServiceNameOfProductsByGatewayAccountId(gatewayAccountId, serviceName))
```
kakumara: no need to pass the `correlationId` here ?

maxcbc: nope. productsclient gets it by other means.

kakumara: 😳  👍

PR #621, opened 2018-01-15
Remove "central"

soniaturcotte:
We are starting a pilot with local authorities and so central government email is no longer required.

## WHAT
Remove 'central' from email hint copy


maxcbc:
Would it not be clearer to say 'must be from a central or local government organisation'?

jonheslop:
Replacing with #632 as I can’t rebase this one

PR #624, opened 2018-01-19
Revert "PP-3048 - Added product service name update to edit service name in service_service."

SandorArpa:
Reverts alphagov/pay-selfservice#620

End to end tests are failing, Max' going to fix it but needs to revert the code so it does not block the build line.


SandorArpa:
No longer needed.

PR #627, opened 2018-01-23
PP-3193 Fix transaction download bug

maxcbc:
## WHAT
- removed `state` param from transactions search when downloading csv and when refund_states or payment_states filters have been selected

## WHY
- because connector expects selfservice to make this call either with the old pattern ('state') or the new pattern ('refund_states', 'payment_states') and not both. This has been handled for the transactions list page but not the download page.

## HOW
- on selfservice, with refund reporting enabled, filter by refund states and try to download a csv.


maxcbc:
@alexbishop1 Did a refactor, including some tests

PR #633, opened 2018-01-25
pp-3167: use run-with-chamber.sh instead of npm

tlwr:
All the other microservices that require secrets use run-with-chamber.sh

This service used npm start-with-chamber

Now it uses run-with-chamber.sh

### DIFF
--- a/run-with-chamber.sh
+++ b/run-with-chamber.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+set -u
+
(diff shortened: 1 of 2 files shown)


run-with-chamber.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+AWS_REGION="${ECS_AWS_REGION}" chamber exec "${ECS_SERVICE}" -- npm start
```
rhowe-gds: curly braces are superfluous and you probably want `set -u`

PR #640, opened 2018-01-30
PP-3204 Fix bug (headers not set by base_client)

SandorArpa:
- removed payApiToken which is needed for publicauth. Placed it back

### DIFF
--- a/app/services/clients/products_client.js
+++ b/app/services/clients/products_client.js
@@ -48,6 +48,7 @@ function createProduct (options) {
     json: true,
     body: {
       gateway_account_id: options.gatewayAccountId,
+      pay_api_token: options.payApiToken,
       name: options.name,
       price: options.price,
       description: options.description,


app/services/clients/products_client.js
```diff
@@ -43,11 +45,13 @@ module.exports = {
  */
 function createProduct (options) {
   return baseClient.post({
+    headers,
     baseUrl,
     url: `/products`,
     json: true,
     body: {
       gateway_account_id: options.gatewayAccountId,
+      pay_api_token: options.payApiToken,
```
maxcbc: I think it was the lack of the `pay_api_token` that was making it break rather than the lack of headers.

kakumara: This is  a different token (publicapi bearer token for creating a payment). This is still needed. we removed the need for having `PRODUCTS_API_TOKEN` which was in header an no longer exist.

kakumara: oh i see what you mean @maxcbc . Why on earth was this removed??

PR #644, opened 2018-02-01
PP-3061 Create/revoke direct debit API keys

simad:
## WHAT
- We can now generate api keys for direct debit. As we store now gateway accounts in many places, using
  internal account ids is not sufficient anymore. We need to support using external ids in lieu of internal ones.
- The `get_gateway_account` middleware will retrieve a direct debit gateway account if the api token is prefixed
  with `DIRECT_DEBIT:`
- When creating an api token, we send to publicauth either the internal or the external id. Publicauth supports already this.
- Changed tests to use not numeric account ids.

### DIFF
--- a/app/controllers/api_keys_controller.js
+++ b/app/controllers/api_keys_controller.js
@@ -3,16 +3,14 @@ var csrf = require('csrf')
 var response = require('../utils/response.js').response
 var errorView = require('../utils/response.js').renderErrorView
 var auth = require('../services/auth_service.js')
-
 const publicAuthClient = require('../services/clients/public_auth_client')
-
+const {DIRECT_DEBIT_TOKEN_PREFIX} = require('../services/clients/direct_debit_connector_client.js')
 // TODO remove these and make them proper i.e. show update destroy etc
 const API_KEYS_INDEX = 'tokens'
 const API_KEY_GENERATE = 'token_generate'
 
 module.exports.index = function (req, res) {
-  var accountId = auth.getCurrentGatewayAccountId(req)
-
+  const accountId = auth.getCurrentGatewayAccountId(req)
   publicAuthClient.getActiveTokensForAccount({
     correlationId: req.correlationId,
     accountId: accountId
@@ -41,7 +39,7 @@ module.exports.index = function (req, res) {
 }
 
 module.exports.revoked = function (req, res) {
-  var accountId = auth.getCurrentGatewayAccountId(req)
+  const accountId = auth.getCurrentGatewayAccountId(req)
   publicAuthClient.getRevokedTokensForAccount({
     correlationId: req.correlationId,
     accountId: accountId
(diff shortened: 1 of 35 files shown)


test/unit/middleware/get_gateway_account_test.js
```diff
@@ -30,37 +30,63 @@ var setupGetGatewayAccount = function (currentGatewayAccountID, paymentProvider)
...
     getGatewayAccount(req, res, next).then(() => {
-      expect(req.account).to.deep.equal({id: 1, payment_provider: 'worldpay', supports3ds: true})
+      expect(req.account).to.deep.equal({id: '1', payment_provider: 'worldpay', supports3ds: true})
       done()
     })
   })
   it('should extend the account data with supports3ds set to false if the account type is not worldpay', done => {
     lodash.set(req, 'user.serviceRoles[0]', {gatewayAccountIds: ['1', '2', '3']})
-    let getGatewayAccount = setupGetGatewayAccount(1, 'epdq')
+    let getGatewayAccount = setupGetGatewayAccount('1', 'epdq')
```
maxcbc: should be const. same above

app/controllers/api_keys_controller.js
```diff
@@ -73,19 +71,20 @@ module.exports.show = function (req, res) {
 }
 
 module.exports.create = function (req, res) {
-  let accountId = auth.getCurrentGatewayAccountId(req)
+  // current account id is either external (DIRECT_DEBIT) or internal (CARD) for now
+  const currentAccountId = auth.getCurrentGatewayAccountId(req)
+  let tokenType = currentAccountId.startsWith(DIRECT_DEBIT_TOKEN_PREFIX) ? 'DIRECT_DEBIT' : 'CARD'
```
maxcbc: let should be const

app/controllers/dashboard/dashboard-activity-controller.js
```diff
@@ -9,13 +9,15 @@ const moment = require('moment-timezone')
 const response = require('../../utils/response').response
 const CORRELATION_HEADER = require('../../utils/correlation_header').CORRELATION_HEADER
 const ConnectorClient = require('../../services/clients/connector_client').ConnectorClient
+const directDebitConnectorClient = require('../../services/clients/direct_debit_connector_client.js')
```
maxcbc: Might be clearer to use destructuring assignment here as in:
`const {DIRECT_DEBIT_TOKEN_PREFIX} = require('../../services/clients/direct_debit_connector_client.js')`

app/services/clients/direct_debit_connector_client.js
```diff
@@ -0,0 +1,50 @@
+'use strict'
```
maxcbc: 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

test/integration/dev_tokens_ft_tests.js
```diff
@@ -22,6 +22,7 @@ var aCorrelationHeader = {
 }
 
 var connectorMock = nock(process.env.CONNECTOR_URL, aCorrelationHeader)
+var directDebitConnectorMock = nock(process.env.DIRECT_DEBIT_CONNECTOR_URL, aCorrelationHeader)
```
maxcbc: If you are using nock, you need to make sure you call `'nock.cleanAll'` after

test/integration/get_gateway_account_ft_test.js
```diff
@@ -65,4 +65,44 @@ describe('get account', function () {
       })
       .end(done)
   })
+
+  it('s.onlyhould get a direct debit account', function (done) {
```
maxcbc: .only lol

simad: ahahahahah

test/integration/dev_tokens_ft_tests.js
simad: Isn't it enough to do that in: ```  afterEach(function () {
      nock.cleanAll()
      app = null
    })
```?

maxcbc: sorry didnae see it

PR #645, opened 2018-02-05
PP-3120: Filter non-prototype links from links page

whpearson:
We don't want non-demo links showing up in the links page
for prototyping.

This PR filters it on the controller level and adds a test
to make sure this filtering is working.


kakumara:
Just needs rebasing

PR #647, opened 2018-02-09
Pp 3300 list dd gateway accounts

simad:
## WHAT
- If a user belongs to a service with associated dd accounts, we should show them in the services list, so that it's possible to switch to one of those and generate API keys.
 - If an account is not found (connector returns a 404), we now log the error and simply do not show the account in the list (like it was before)

- We render a gateway account in the services list by apending `(<payment_method>)` at the end of the gateway account name if present. This is currently populated just for direct debit. For example, a dd gateway account will be `Test account (sandbox) (direct debit)`
 - Serialising the internal id as external for card gateway accounts, and populating the hidden input with the external account when switching.


test/fixtures/gateway_account_fixtures.js
```diff
@@ -31,7 +31,25 @@ module.exports = {
       }
     }
   },
+  validDirectDebitGatewayAccountResponse: (opts = {}) => {
+    let data = {
```
maxcbc: This should be `const`

PR #648, opened 2018-02-09
PP-3305 Fix validation mismatch for products creation

SandorArpa:
- At the point of payment creation the validator MAX_PRICE is set to pounds while
       the incoming payload's `price` field is in pence.

with @kakumara

### DIFF
--- a/test/unit/browsered/field_validation_checks_test.js
+++ b/test/unit/browsered/field_validation_checks_test.js
@@ -8,8 +8,8 @@ const {isBelowMaxAmount, isPasswordLessThanTenChars} = require('../../../app/bro
 
 describe('field validation checks', () => {
   describe('isBelowMaxAmount', () => {
-    it('should return an error string if it is passed an currency string exceeding £10 million', () => {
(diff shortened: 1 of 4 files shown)


test/unit/browsered/field_validation_checks_test.js
```diff
@@ -8,8 +8,8 @@ const {isBelowMaxAmount, isPasswordLessThanTenChars} = require('../../../app/bro
 
 describe('field validation checks', () => {
   describe('isBelowMaxAmount', () => {
-    it('should return an error string if it is passed an currency string exceeding £10 million', () => {
-      expect(isBelowMaxAmount('10000000.01')).to.equal(`Choose an amount under £10,000,000`)
+    it('should return an error string if it is passed an currency string exceeding £100 thousands', () => {
+      expect(isBelowMaxAmount('10000000.01')).to.equal(`Choose an amount under £100,000`)
```
georgeracu: Shouldn't this be `100000.01`? Given the fact that here we are counting pence?

PR #652, opened 2018-02-13
PP-3295 Service name validation. Create and Edit service name

rauligs:
## WHAT
- Add common way to add service name validation
- Make Create and Edit validation consistent across microservices (max 50 chars)
- Create and Edit Service (add service name). 
- Refactor to use field_validations rather than Field classes
- Use same validation in registration process


app/browsered/field-validation-checks.js
```diff
@@ -65,4 +69,6 @@ exports.isAboveMaxAmount = value => {
   return false
 }
 
+exports.isFieldGreaterThanMaxLengthChars = (value, maxLength) => { return _.isNumber(maxLength) && value.length > maxLength ? validationErrors.isGreaterThanMaxLengthChars : false}
```
jonheslop: Need a space before the `{`

app/browsered/field-validation-checks.js
```diff
@@ -1,5 +1,8 @@
 'use strict'
 
+// npm dependencies
+const _ = require('lodash')
```
jonheslop: This is client side JS so including lodash loads the whole thing in the browser. Can you just include the function you need like so:

```js
const isNumber = require('lodash/isnumber')
```

PR #653, opened 2018-02-14
PP-3251 Add error logging to products controllers

SandorArpa:
- add error logging in catch clause to identify what went wrong in make_a_demo_payment and test_with_your_users controllers.

### DIFF
--- a/app/controllers/make_a_demo_payment/go_to_payment_controller.js
+++ b/app/controllers/make_a_demo_payment/go_to_payment_controller.js
@@ -3,6 +3,8 @@
 // NPM dependencies
 const lodash = require('lodash')
 
+const logger = require('winston')
+
 // Local dependencies
 const paths = require('../../paths')
 const productsClient = require('../../services/clients/products_client.js')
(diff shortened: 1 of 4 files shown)


app/controllers/make_a_demo_payment/go_to_payment_controller.js
```diff
@@ -38,7 +40,8 @@ module.exports = (req, res) => {
       lodash.unset(req, 'session.pageData.makeADemoPayment')
       res.redirect(product.links.pay.href)
     })
-    .catch(() => {
+    .catch((err) => {
+      logger.error(`[requestId=${req.correlationId}] Making a demo payment failed - ${err.message}`)
```
kakumara: Have you tested this (in integration). I guess we have a slight difference from `publicauth` and `products` the way we return error messages. products doesn't send `.message`
How about 
```
if(err.message) {
logger.error(`${err.message}`)
} else {
logger.error(`${JSON.stringify(err}`)
}
```

kakumara: or just `JSON.stringify` for both ?

SandorArpa: Tests shows that always an Error object is returned, which always have a message

PR #661, opened 2018-02-21
PP-3191 - adding virtual pageview so we can track signups

jonheslop:
We want to be able to track signups as a goal to see if people are dropping off through the journey, but to do this we need a definitive end, so over on [GaaP Analytics](https://github.com/alphagov/gaap-analytics/pull/4) I added the ability to do so. This PR just updates our version of the npm module and inserts the markup


kakumara:
seriously, no tests @jonheslop ?

jonheslop:
@kakumara Awww busted ☹️

PR #673, opened 2018-02-23
BAU add last know healthy shrinkwrap

maxcbc:
## WHAT
Add the shrinkwrap from the last known healthy version of selfservice to master.
This should pin the dependencies to the versions as they were before we experienced problems.


whpearson:
The version of nunjuks being pulled in

`+-- nunjucks@3.0.1 ` 
👍

PR #683, opened 2018-03-01
PP-3432 - Renaming Payment types to Card types

jonheslop:
With direct debit coming soon we will have two payment types.

So we need to rename the current payment types to be card types,
so things aren’t confusing.

I went further a replaced every single instance of payment with card, including file names.
It worked but would have required a major e2e rewrite too and it didnt seem worth it.

### DIFF
--- a/app/paths.js
+++ b/app/paths.js
@@ -46,9 +46,9 @@ module.exports = {
     delete: '/tokens'
   },
   paymentTypes: {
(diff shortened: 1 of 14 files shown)


jonheslop:
This will error until https://github.com/alphagov/pay-endtoend/pull/536 is merged

PR #685, opened 2018-03-01
Fix link to documentation about Worldpay 3D Secure setup

alexbishop1:
Change link from https://govukpay-docs.cloudapps.digital/switching-to-production#worldpay-setup-for-3d-secure to https://govukpay-docs.cloudapps.digital/#worldpay-setup-for-3d-secure


jonheslop:
But FYI, I just broke self service master, fixing acceptance tests now, but this is also needs to go in https://github.com/alphagov/pay-selfservice/pull/684

PR #691, opened 2018-03-05
PP-3421 set fixed amount

jonheslop:
Before this PR user could only have a variable amount payment link. Now there is a new screen in the flow where a user can opt for a fixed amount, this amount is validated and passed through.

Also wrote unit tests for it.

### DIFF
--- a/app/controllers/payment-links/get-amount-controller.js
+++ b/app/controllers/payment-links/get-amount-controller.js
@@ -10,9 +10,11 @@ const paths = require('../../paths')
 module.exports = (req, res) => {
   const pageData = lodash.get(req, 'session.pageData.createPaymentLink', {})
   const paymentLinkAmount = req.body['payment-amount'] || pageData.paymentLinkAmount || ''
+  const paymentAmountType = req.body['amount-type-group'] || pageData.paymentAmountType || ''
 
   return response(req, res, 'payment-links/amount', {
     paymentLinkAmount,
+    paymentAmountType,
     nextPage: paths.paymentLinks.amount,
     returnToStart: paths.paymentLinks.start,
     manage: paths.paymentLinks.manage
--- a/app/controllers/payment-links/get-amount-controller.js
+++ b/app/controllers/payment-links/get-amount-controller.js
@@ -9,15 +9,11 @@ const paths = require('../../paths')
 
 module.exports = (req, res) => {
   const pageData = lodash.get(req, 'session.pageData.createPaymentLink', {})
+  const paymentLinkAmount = req.body['payment-amount'] || pageData.paymentLinkAmount || ''
 
-  return response(req, res, 'payment-links/review', {
-    pageData,
-    paymentLinkTitle: pageData.paymentLinkTitle,
-    paymentLinkDescription: pageData.paymentLinkDescription,
-    paymentLinkAmount: pageData.paymentLinkAmount,
-    nextPage: paths.paymentLinks.review,
-    changeInformation: paths.paymentLinks.information,
-    changeAmount: paths.paymentLinks.amount,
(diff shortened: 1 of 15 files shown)


app/controllers/payment-links/post-amount-controller.js
```diff
@@ -0,0 +1,48 @@
...
+  const paymentAmountType = req.body['amount-type-group']
+  const paymentLinkAmount = req.body['payment-amount']
+
+  if (!paymentAmountType) {
+    req.flash('genericError', `<h2>There was a problem with the details you gave for:</h2><ul class="error-summary-list"><li><a href="#fixed-or-variable">Is the payment for a fixed amount?</a></li></ul>`)
+    req.flash('errorType', `paymentAmountType`)
+    return res.redirect(paths.paymentLinks.amount)
+  }
+
+  let formattedPaymentLinkAmount = paymentLinkAmount.replace(/[^0-9.-]+/g, '')
```
ckalista: What the minus sign in there for? Negative amount?

app/controllers/payment-links/get-amount-controller.js
```diff
@@ -0,0 +1,20 @@
...
+// NPM dependencies
+const lodash = require('lodash')
+
+// Local dependencies
+const {response} = require('../../utils/response.js')
+const paths = require('../../paths')
+
+module.exports = (req, res) => {
+  const pageData = lodash.get(req, 'session.pageData.createPaymentLink', {})
+  const paymentLinkAmount = req.body['payment-amount'] || pageData.paymentLinkAmount || ''
```
ckalista: Should we not derive paymentAmountType from paymentLinkAmount and pass it to down to the template instead of having that logic in the template?

app/controllers/payment-links/post-amount-controller.js
```diff
@@ -0,0 +1,48 @@
...
+  if (formattedPaymentLinkAmount && !currencyMatch[2]) {
+    formattedPaymentLinkAmount = paymentLinkAmount + '.00'
+  }
+
+  if (paymentAmountType === 'variable') {
+    formattedPaymentLinkAmount = ''
+  }
+
+  updatedPageData.paymentLinkAmount = formattedPaymentLinkAmount
+  lodash.set(req, 'session.pageData.createPaymentLink', updatedPageData)
```
ckalista: Question if we should persist the paymentAmountType as well?

app/views/payment-links/amount.njk
```diff
@@ -0,0 +1,64 @@
...
+      <input id="csrf" name="csrfToken" type="hidden" value="{{csrf}}"/>
+      <div class="form-group{% if flash.genericError %} error{% endif %}">
+        {% if flash.errorType == 'paymentAmountType' %}
+        <span class="error-message">
+          Please choose an option
+        </span>
+        {% endif %}
+        <fieldset data-module="show-hide-content" id="fixed-or-variable">
+          <div class="multiple-choice" data-target="fixed-amount">
+            <input id="amount-type-fixed" type="radio" name="amount-type-group" value="fixed" {% if paymentLinkAmount or flash.errorType == 'paymentAmountFormat' %}checked{% endif %}>
```
ckalista: I see the logic for paymentAmountType is here, but would be better to encapsulate in the controller?

app/views/payment-links/review.njk
```diff
@@ -64,11 +64,9 @@ Create a payment link - {{currentServiceName}} {{currentGatewayAccount.full_type
             {% endif %}
           </dd>
           <dd class="cya-change">
-<!--      
-            <a href="{{ change }}">
+            <a href="{{ changeAmount }}">
```
ckalista: I don't see any mention about fixed or variable amount... is that as expected?

PR #693, opened 2018-03-07
PP-3426 update content

jonheslop:
Mark and Sonia made some updates to the Create a payment link flow this is the updates

### DIFF
--- a/app/controllers/make_a_demo_payment/index_controller.js
+++ b/app/controllers/make_a_demo_payment/index_controller.js
@@ -12,7 +12,7 @@ const DEFAULTS = {
   paymentDescription: 'An example payment description',
   paymentAmount: '20.00'
 }
-const AMOUNT_FORMAT = /^([0-9]+)(?:\.([0-9]{2}))?$/
+const AMOUNT_FORMAT = /^([0-9]+)(?:\.([0-9]{1,2}))?$/
 
 module.exports = (req, res) => {
   const pageData = lodash.get(req, 'session.pageData.makeADemoPayment', {})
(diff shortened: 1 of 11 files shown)


app/controllers/make_a_demo_payment/index_controller.js
```diff
@@ -33,6 +33,8 @@ module.exports = (req, res) => {
   const currencyMatch = AMOUNT_FORMAT.exec(paymentAmount)
```
simad: Not a deal breaker, but maybe we could extract this bit in a function, with a name which makes it possible to understand what's it doing, at a glance? 
Just reading the code it's quite hard to understand what's going on - it's padding the amount, but what's the logic behind it? :)

app/controllers/payment-links/post-amount-controller.js
```diff
@@ -31,6 +31,8 @@ module.exports = (req, res) => {
 
   if (formattedPaymentLinkAmount && !currencyMatch[2]) {
```
simad: same as above

app/controllers/test_with_your_users/submit_controller.js
```diff
@@ -44,6 +44,8 @@ module.exports = (req, res) => {
   const currencyMatch = AMOUNT_FORMAT.exec(paymentAmount)
```
simad: same. The fact that we had to repeat this function 3 times, makes me think we could actually extract it

jonheslop: Ok I started doing this in this PR but actually I am going to it after I merge, as I have another PR that could benefit by being rebased on top of this

PR #698, opened 2018-03-08
PP-3482: Enclose filename in quotes

whpearson:
We had a bug on firefox in windows where it ignored
the string after GOVUK.

Just enclose it in quotes.

https://stackoverflow.com/questions/21442903/firefox-has-problems-when-downloading-with-a-space-in-filename


whpearson:
Fixed in this commit https://github.com/alphagov/pay-selfservice/commit/c33ff3e2fa20c68b29718b95efcdd39d4850ff6c

PR #707, opened 2018-03-12
PP-3449 Enable 3DS for EPDQ

georgeracu:
## WHAT 
- Will display functionality for enabling 3DS for EPDQ
- Will hide helper text in 3DS area that is specific for worldpay
- It uses an env var to toggle displaying the 3DS button

### DIFF
--- a/test/unit/middleware/get_gateway_account_test.js
+++ b/test/unit/middleware/get_gateway_account_test.js
@@ -1,7 +1,6 @@
 'use strict'
-// Core Dependencies
+
 const path = require('path')
-// NPM Dependencies
 const proxyquire = require('proxyquire')
 const lodash = require('lodash')
 const sinon = require('sinon')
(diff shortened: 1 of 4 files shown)


test/unit/middleware/get_gateway_account_test.js
```diff
@@ -71,11 +71,28 @@ describe('middleware: getGatewayAccount', () => {
...
+      })
+    } else {
+      lodash.set(req, 'user.serviceRoles[0]', {gatewayAccountIds: ['1', '2', '3']})
+      const getGatewayAccount = setupGetGatewayAccount('1', 'epdq')
+      getGatewayAccount(req, res, next).then(() => {
+        expect(req.account).to.deep.equal({id: '1', payment_provider: 'epdq', supports3ds: false})
+        done()
+      })
+    }
+  })
```
chrisgrimble: Is it possible to create two tests for this? One where we enable the environment variable and one where we do not. Rather than having a single test that depends on the state of EPDQ_3DS_ENABLED to decide what get run.

PR #708, opened 2018-03-12
PP-3491 - rename payment 'state' as 'payment status'

jonheslop:


georgeracu:
LOKTM even though when searching for a refund status, the text reads as `transactions with **Refund submitted** or **Refund error** payment status`

PR #715, opened 2018-03-13
WIP PP-3498 dropdown selects individual error states clean up

chrisgrimble:
Removing unneeded code. Previously we returned failed for the connector
state of a transaction and need to use the error code to map to the
state display value. Because we wanted to search by these states we
need to send different states for each one we displayed. This means we
no longer need to use the error code to map to display states so
removing the code.


chrisgrimble:
This cannot be merge yet as we are not changing the charge connector endpoint at the moment

PR #721, opened 2018-03-15
PP-3436 Run e2e, accept and zap separately in parallel

maxcbc:
# PP-3436 Run e2e, accept and zap separately in parallel

## WHAT
This is a jenkinsfile update which:
- switches to using separate e2e, accept and zap test jobs
- runs each of these jobs in parallel
- switches each of these jobs on its own param for testing on pull requests

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,7 +4,9 @@ pipeline {
   agent any
 
   parameters {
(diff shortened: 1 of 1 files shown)


jonheslop:
😱 This sound like an excellent idea

PR #723, opened 2018-03-19
PP-3510 - adding friendly url to the flow 

jonheslop:
Only merge after EndToEnd: https://github.com/alphagov/pay-endtoend/pull/586

## 688433f - adding slugify package to format urls
This is a tiny package with no other dependencies that makes it easy for
us to consistently format urls client and server side.

## c36bdc3 - Adding url playback confirmation 
- slugify is accessible as a nunjucks filter for ease of use
- extending input confirm to be able to use slugify
- also removing indefinite article (a/an) from the url to keep the url
short.
- remove the definite article (the) too

## f6c4540 - Passed friendly URL to link creation 
- Added friendly URL to the session
- extended products client to include friendly URL bits
- prefer display of friendly URL on manage page

### NOTE: 
This also contains a backward compatible fix to make the JS work in PhantomJS

### DIFF
--- a/app/browsered.js
+++ b/app/browsered.js
@@ -10,6 +10,7 @@ const fieldValidation = require('./browsered/field-validation')
 const dashboardActivity = require('./browsered/dashboard-activity')
 const targetToShow = require('./browsered/target-to-show')
 const analytics = require('gaap-analytics')
+const inputConfirm = require('./browsered/input-confirm')
 
 // GOV.UK Frontend Toolkit dependencies
 require('../govuk_modules/govuk_frontend_toolkit/javascripts/govuk/show-hide-content')
@@ -20,6 +21,7 @@ dashboardActivity.init()
 targetToShow.init()
 analytics.eventTracking.init()
 analytics.virtualPageview.init()
+inputConfirm()
 
 $(document).ready($ => {
   const showHideContent = new window.GOVUK.ShowHideContent()
--- a/app/browsered.js
+++ b/app/browsered.js
@@ -11,6 +11,7 @@ const dashboardActivity = require('./browsered/dashboard-activity')
 const targetToShow = require('./browsered/target-to-show')
 const analytics = require('gaap-analytics')
 const inputConfirm = require('./browsered/input-confirm')
+const niceURL = require('./browsered/nice-url')
 
 // GOV.UK Frontend Toolkit dependencies
 require('../govuk_modules/govuk_frontend_toolkit/javascripts/govuk/show-hide-content')
@@ -22,6 +23,7 @@ targetToShow.init()
 analytics.eventTracking.init()
 analytics.virtualPageview.init()
 inputConfirm()
+niceURL()
 
 $(document).ready($ => {
   const showHideContent = new window.GOVUK.ShowHideContent()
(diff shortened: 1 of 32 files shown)


app/controllers/payment-links/post-information-controller.js
```diff
@@ -12,8 +13,12 @@ module.exports = (req, res) => {
 
   updatedPageData.paymentLinkTitle = req.body['payment-link-title']
   updatedPageData.paymentLinkDescription = req.body['payment-link-description']
+  updatedPageData.serviceNamePath = req.body['service-name-path']
+  updatedPageData.productNamePath = slugify(req.body['payment-link-title'].replace(/\ba\s|\ban\s|\bthe\b/gi, ''))
   lodash.set(req, 'session.pageData.createPaymentLink', updatedPageData)
 
+  console.log(updatedPageData)
```
ckalista: To remove.

app/controllers/payment-links/post-information-controller.js
```diff
@@ -12,8 +13,12 @@ module.exports = (req, res) => {
 
   updatedPageData.paymentLinkTitle = req.body['payment-link-title']
   updatedPageData.paymentLinkDescription = req.body['payment-link-description']
+  updatedPageData.serviceNamePath = req.body['service-name-path']
```
ckalista: What have we decided for the service name. I reckon the same normalisation rules apply as for product name. May we should create a separate function that we can reuse as we type the title and then again when we validate the input fields after an update?

app/controllers/payment-links/post-information-controller.js
```diff
@@ -12,8 +13,12 @@ module.exports = (req, res) => {
...
+  updatedPageData.productNamePath = slugify(req.body['payment-link-title'].replace(/\ba\s|\ban\s|\bthe\b/gi, ''))
```
ckalista: We need to have test coverage for this.

app/controllers/payment-links/post-review-controller.js
```diff
@@ -54,7 +56,9 @@ module.exports = (req, res) => {
         paymentLinkTitle,
         paymentLinkDescription,
         paymentLinkAmount,
-        req.service.name
+        req.service.name,
+        serviceNamePath,
+        productNamePath
```
ckalista: We need test coverage for this.

jonheslop: Added tests for everything

app/controllers/payment-links/get-web-address-controller.js
```diff
@@ -0,0 +1,22 @@
...
+const lodash = require('lodash')
+
+// Local dependencies
+const {response} = require('../../utils/response.js')
+const paths = require('../../paths')
+
+module.exports = (req, res) => {
+  const friendlyURL = process.env.PRODUCTS_FRIENDLY_BASE_URI
+  const pageData = lodash.get(req, 'session.pageData.createPaymentLink', {})
+  const productNamePath = req.body['payment-amount'] || pageData.productNamePath || ''
```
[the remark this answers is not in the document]
ckalista: Good spot, I believe `req.body['payment-amount']` is always going to be undefined hence can safely be removed.

app/browsered.js
```diff
@@ -10,6 +10,8 @@ const fieldValidation = require('./browsered/field-validation')
 const dashboardActivity = require('./browsered/dashboard-activity')
 const targetToShow = require('./browsered/target-to-show')
 const analytics = require('gaap-analytics')
+const inputConfirm = require('./browsered/input-confirm')
+const niceURL = require('./browsered/nice-url')
```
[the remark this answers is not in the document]
ckalista: Technically this is not the friendly URL we are building here so I will leave it as it is.

PR #726, opened 2018-03-20
PP-2984 Add `hasDirectDebitGatewayAccount` dynamic property to `req.service`

DanailMinchev:
## WHAT

Modified `resolve_service` middleware to add `hasDirectDebitGatewayAccount` property which calculates if we the current service has Direct Debit gateway account or not

### DIFF
--- a/app/middleware/resolve_service.js
+++ b/app/middleware/resolve_service.js
@@ -17,5 +17,8 @@ module.exports = function (req, res, next) {
     return renderErrorView(req, res, 'You do not have the rights to access this service.')
   }
 
+  req.service.hasDirectDebitGatewayAccount =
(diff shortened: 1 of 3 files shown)


app/middleware/resolve_service.js
```diff
@@ -17,5 +17,15 @@ module.exports = function (req, res, next) {
     return renderErrorView(req, res, 'You do not have the rights to access this service.')
   }
 
+  req.service.hasDirectDebitGatewayAccount = false
```
simad: If you need to these kind of things (set a variable, iterate over an array, change that variable, break) usually languages nowadays have one-liners to do all of that. In this case: 
```  req.service.hasDirectDebitGatewayAccount = req.service.gateway_account_ids.some((gatewayAccountId) => gatewayAccountId.startsWith('DIRECT_DEBIT:'))```

DanailMinchev: Great optimisation!

PR #734, opened 2018-03-23
PP-3525 Enable 3DS for SmartPay with feature flag

georgeracu:
## WHAT
- Enable 3DS for SmartPay
- Based on a feature flag (env variable) we decide to enable/disable 3DS for SmartPay

### DIFF
--- a/app/middleware/get_gateway_account.js
+++ b/app/middleware/get_gateway_account.js
@@ -5,6 +5,7 @@ const directDebitConnectorClient = require('../services/clients/direct_debit_con
 const _ = require('lodash')
 const winston = require('winston')
 const EPDQ_3DS_ENABLED = process.env.EPDQ_3DS_ENABLED
+const SMARTPAY_3DS_ENABLED = process.env.SMARTPAY_3DS_ENABLED || 'false'
 
 module.exports = function (req, res, next) {
   const accountId = auth.getCurrentGatewayAccountId(req)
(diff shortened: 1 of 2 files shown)


app/middleware/get_gateway_account.js
```diff
@@ -25,7 +26,9 @@ module.exports = function (req, res, next) {
   }
   return connectorClient.getAccount(params)
     .then(data => {
-      const SUPPORTS_3DS = (EPDQ_3DS_ENABLED === 'true') ? ['worldpay', 'epdq'] : ['worldpay']
+      // env var values are treated as text so the comparisaon is done for text
+      let SUPPORTS_3DS = (EPDQ_3DS_ENABLED === 'true') ? ['worldpay', 'epdq'] : ['worldpay']
+      if (SMARTPAY_3DS_ENABLED === 'true') { SUPPORTS_3DS = _.concat(SUPPORTS_3DS, ['smartpay']) }
```
kakumara: This bit is okay, but at a glance bit cryptic to as the same thing handled slightly differently. How about something  like
```
const SUPPORTS_3DS = ['worldpay']
if(EPDQ_3DS_ENABLED === 'true') {
    _.concat(SUPPORTS_3DS, ['epdq'])
}
if(SMARTPAY_3DS_ENABLED === 'true') {
    _.concat(SUPPORTS_3DS, ['smartpay'])
}
```

georgeracu: I was thinking to use that style as well, I just pick one

PR #752, opened 2018-04-10
PP-3525 Show/hide 3DS helper text for Worldpay only

georgeracu:
## WHAT
- Moved all helper text for 3DS Worldpay into one div
- Show/hide 3DS helper text on 3DS confirmation page base don gateway
- Added tests for this cases

### DIFF
--- a/test/ui/toggle_3ds_ui_tests.js
+++ b/test/ui/toggle_3ds_ui_tests.js
@@ -107,7 +107,24 @@ describe('The toggle 3D Secure page when 3D Secure is off', function () {
 
     let body = renderTemplate('3d_secure/index', templateData)
 
-    body.should.containSelector('#threeds-helper-text')
+    body.should.containSelector('#3ds-helper-text-worldpay')
+  })
+
+  it('should not display helper test if not worldpay', function () {
+    let templateData = {
+      'supports3ds': true,
+      'requires3ds': false,
+      'justToggled': false,
+      'showHelper3ds': false,
+      permissions: {
+        toggle_3ds_read: true,
+        toggle_3ds_update: true
+      }
+    }
+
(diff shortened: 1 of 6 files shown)


test/ui/toggle_3ds_ui_tests.js
```diff
@@ -296,9 +313,38 @@ describe('The turn 3D Secure on confirmation page', function () {
...
+      }
+    }
+
+    let body = renderTemplate('3d_secure/on_confirm', templateData)
+
+    body.should.containNoSelector('#3ds-helper-text-worldpay')
+  })
+
+  it('should display helper test if worldpay', function () {
+    let templateData = {
```
DanailMinchev: We can use `const` by default for local and global variables. In case the variable will be changed (with another reference) we can use `let` and `var` is not used

test/ui/toggle_3ds_ui_tests.js
```diff
@@ -107,7 +107,24 @@ describe('The toggle 3D Secure page when 3D Secure is off', function () {
 
     let body = renderTemplate('3d_secure/index', templateData)
 
-    body.should.containSelector('#threeds-helper-text')
+    body.should.containSelector('#3ds-helper-text-worldpay')
+  })
+
+  it('should not display helper test if not worldpay', function () {
+    let templateData = {
```
DanailMinchev: I think the best practice currently not to use UI tests anymore, but to use `supertest` as for example in https://github.com/alphagov/pay-selfservice/tree/3a9370c80b190c090e3bf4248b7fe84e35fb8eb7/test/unit/controller/edit_merchant_details_controller
Because with UI tests we were skipping the controller logic, etc.
May be verify with someone else if the new way will be too complicated

PR #754, opened 2018-04-10
PP-3576 allow users to configure 2fa app

jonheslop:
## Team work makes the dream work
_A Bishlop production_

- Inital commits by @alexbishop1 to get rudimentary version working
- Refined by @jonheslop and then added the tests 🙄 

Currently users can only login using their phone for two factor authentication. This PR adds the ability to use an authenticator app instead. The change can be made via the users profile.

screen shot 2018-04-16 at 10 55 55


app/controllers/login/otp-login-controller.js
```diff
@@ -9,17 +9,20 @@ const errorView = require('../../utils/response').renderErrorView
 const CORRELATION_HEADER = require('../../utils/correlation_header').CORRELATION_HEADER
 
 module.exports = (req, res) => {
-  if (!req.session.sentCode) {
-    const correlationId = req.headers[CORRELATION_HEADER] || ''
+  let PAGE_PARAMS = {}
```
maxcbc: should be const

app/controllers/login/otp-login-controller.js
```diff
@@ -9,17 +9,20 @@ const errorView = require('../../utils/response').renderErrorView
...
+  const correlationId = req.headers[CORRELATION_HEADER] || ''
+  PAGE_PARAMS.authenticatorMethod = req.user.secondFactor
+
+  if (!req.session.sentCode && req.user.secondFactor === 'SMS') {
     userService.sendOTP(req.user, correlationId).then(function () {
       req.session.sentCode = true
-      res.render('login/otp-login')
+      res.render('login/otp-login', PAGE_PARAMS)
     }, function (err) {
```
maxcbc: use fat arrow function

app/controllers/login/send-again-post-controller.js
```diff
@@ -11,13 +11,17 @@ const CORRELATION_HEADER = require('../../utils/correlation_header').CORRELATION
...
-    (err) => {
-      errorView(req, res)
-      logger.error(err)
-    }
-  )
+  if (req.user.secondFactor === 'SMS') {
+    userService.sendOTP(req.user, correlationId).then(
+      () => {
+        res.redirect(paths.user.otpLogIn)
+      },
```
maxcbc: would be clearer with an explicit `.catch()`. This is the pattern we've used elsewhere.

app/controllers/two-factor-auth-controller/get-configure-controller.js
```diff
@@ -0,0 +1,35 @@
...
+const {response} = require('../../utils/response.js')
+const paths = require('../../paths')
+
+const PAGE_PARAMS = {
+  profile: paths.user.profile,
+  index: paths.user.twoFactorAuth.index,
+  configure: paths.user.twoFactorAuth.configure
+}
+
+const makeOtpUrl = (username, secret) => {
```
maxcbc: this is only used once so it doesn't need to be a function

app/controllers/two-factor-auth-controller/get-configure-controller.js
```diff
@@ -0,0 +1,35 @@
...
+
+// NPM dependencies
+const logger = require('winston')
+const qrcode = require('qrcode')
+
+// Local dependencies
+const {response} = require('../../utils/response.js')
+const paths = require('../../paths')
+
+const PAGE_PARAMS = {
```
maxcbc: this needs to be declared inside the exported function as you are assigning to it. Currently this object is shared between all requests to this controller.

app/controllers/two-factor-auth-controller/get-index-controller.js
```diff
@@ -0,0 +1,16 @@
...
+// Local dependencies
+const {response} = require('../../utils/response.js')
+const paths = require('../../paths')
+
+const PAGE_PARAMS = {
+  profile: paths.user.profile,
+  index: paths.user.twoFactorAuth.index
+}
+
+module.exports = (req, res) => {
```
maxcbc: again here, PAGE_PARAMS needs to be inside function.

app/controllers/two-factor-auth-controller/post-configure-controller.js
```diff
@@ -0,0 +1,30 @@
...
+
+module.exports = (req, res) => {
+  const code = req.body['code'] || ''
+  userService.configureNewOtpKey(req.user.externalId, code, 'APP', req.correlationId)
+    .then(user => {
+      req.flash('generic', `<h2>Your two factor authentication method has been&nbsp;updated</h2><p>Next time you sign in please use your configured authenticator app</p>`)
+      return res.redirect(paths.user.profile)
+    })
+    .catch((err) => {
+      let errorMessageLog, errorMessageUser
```
maxcbc: nice use of `let`

app/paths.js
```diff
@@ -22,9 +22,13 @@ module.exports = {
   user: {
     logIn: '/login',
     profile: '/my-profile',
+    twoFactorAuth: {
+      index: '/my-profile/two-factor-auth',
+      configure: '/my-profile/two-factor-auth/configure',
+      complete: '/my-profile/two-factor-auth/complete'
```
maxcbc: where is this used?

test/ui/login_ui_tests.js
```diff
@@ -3,66 +3,130 @@ let renderTemplate = require(path.join(__dirname, '/../test_helpers/html_asserti
...
         }
       }
-    }
 
-    let body = renderTemplate('login/login', templateDate)
-    body.should.containSelector('.error-summary-heading')
-    body.should.containSelector('label[for="username"] .error-message').withText(templateDate.flash.error.messages.username)
-    body.should.containSelector('label[for="password"] .error-message').withText(templateDate.flash.error.messages.password)
-  })
+      let body = renderTemplate('login/login', templateDate)
```
maxcbc: const-age genererally in this file

app/paths.js
jonheslop: Ha good spot, an earlier version had a complete page but settled on redirect to profile

app/views/twoFactorAuth/index.njk
```diff
@@ -0,0 +1,53 @@
...
+          Set up a different authenticator app
+        </button>
+      </div>
+    </form>
+  {% endif %}
+  <div>
+    <details>
+      <summary>What is an Authenticator app?</summary>
+      <h2 class="heading-small">Authenticator apps</h2>
+      <p>The GOV.UK Pay self-service website requires users to use a security code to sign-in, the code is sent to a users mobile number. It’s possible to get the code from an Authenticator app instead.</p>
```
alexbishop1: Despite it being in the URL, I don’t think we generally refer to self-service as “self-service” in prose (we call it the “GOV.UK Pay Admin site” or the “service admin site” in the docs if we have to but it’s usually possible to avoid this).

I think this text might work better in the second person. Something like: “Instead of text message codes, you can use an authenticator app to sign in to GOV.UK Pay. This is an app (usually installed on your phone) that generates a security code whenever you need to sign in. If you choose to use an authenticator app, you will only be able to sign in to GOV.UK Pay using that app  and you will no longer receive text message codes.”

app/views/twoFactorAuth/index.njk
```diff
@@ -0,0 +1,53 @@
...
+      <p>Users can use any Authenticator app they like but the most common are:</p>
```
alexbishop1: Bit informal. Perhaps “GOV.UK Pay works with most authenticator apps. If you do not have one installed already, there are several available. These include:”

app/views/twoFactorAuth/index.njk
```diff
@@ -0,0 +1,53 @@
...
+      <ul class="list list-bullet">
+        <li><a href="https://support.google.com/accounts/answer/1066447">Google Authenticator</a></li>
+        <li><a href="https://authy.com/">Authy</a></li>
+        <li><a href="https://duo.com/">Duo</a></li>
```
alexbishop1: It makes me sad that none of these links are particularly good. The Google Authenticator one assumes you’re using Google Authenticator just for your Google account (which might confuse people) and the Duo landing page tries to sell to companies rather than users. Even the Authy website (which is the best) basically assumes you already know what an authenticator app is and is built around their guides for high-profile sites.

Other apps we could possibly add to the list include Microsoft Authenticator (again, its site assumes you’re using it to secure a Microsoft Account or an account with an organisation that uses Microsoft’s Azure cloud services) and LastPass Authenticator (fairly good site but assumes you’re probably using LastPass).

PR #755, opened 2018-04-11
PP-3637 - Subsequent paginations didnt include the filters

jonheslop:
Since we made it so that you can filter by multiple states/brands the
logic that passed these filters through pagination broke as it expected
a string not an array. I've changed this and updated the test.

I also noticed we didnt pass email through to pagination, this probably
wasn't picked up before because no one has ever searched so unrefined as
to return that many emails that pagination is necessary. But I've added
it anyway.


jonheslop:
The issue for this is actually PP-3633, I misread the last 3 as a 7!?

PR #757, opened 2018-04-12
WIP - BROWSER TESTING

maxcbc:
## WHAT

## HOW 
_Steps to test or reproduce:_


package.json
```diff
@@ -33,6 +34,9 @@
     "lint": "./node_modules/.bin/standard",
     "lint-sass": "./node_modules/.bin/sass-lint -v",
     "test": "./node_modules/mocha/bin/mocha '!(node_modules)/**/*+(.|_)+(test|tests)'.js",
+    "pact-stub": "docker run -t -p 8000:8000 -v \"$(pwd)/pacts/:/app/pacts\" pactfoundation/pact-stub-server -p 8000 -d pacts",
```
brid: Suggest we include a compiled pact-stub binary which runs up a stub server in exactly the same manner, and removes the dependency on docker. The idea of needing docker as part of an 'npm test' script has never sat well with me, and would rather it encapsulates our processes rather that being integrated with them (if that makes sense). It helps us stick to being able to just clone, npm install and run our test commands which is better for our open sourcing efforts.

Any thoughts @mrlumbu and @rauligs  and @oswaldquek ? Happy to stick with this either way.

mrlumbu: yep that does make sense. ideally should be able to install the binary as a dependency, rather than shunting it around in git

mrlumbu: having said that, I think this is fine for the time being. I'll approve once rebased.

PR #759, opened 2018-04-12
PP-3574 Remove new states feature flag

georgeracu:
## What
Remove feature flag for new transaction states and enable it for everyone

## How
- Removed feature flag conditional code from transactions search,
details and list and CSV download
- Deleted files that contained code for the branch without feature
- Deleted old states and mapping between these states and the states
from `connector`
- Updated tests by removing the feature flag from the ones with the new
states and removed tests for the old states
- Updated tests for `json-to-csv` where status `succeeded` was mapped
successfully to `Succeeded` because the old states were just copying
the state from the JSON data from `connector`. This value doesn't exist
and should be `success` which is now fixed and the new states map
correctly.
- Updated tests for `json-to-csv` where status `canceled` didn't map to
any of the new states due to the spelling error. Now replaced with
`cancelled`
- Opportunistic code quality improvement by replacing `var` and adding
`use strict` where missing


app/controllers/dashboard/dashboard-activity-controller.js
```diff
@@ -60,7 +58,7 @@ module.exports = (req, res) => {
       name: req.user.username,
       serviceId: req.service.externalId,
       activity: activityResults,
-      successfulTransactionsState: newChargeStatusEnabled ? 'payment-success' : 'success',
+      successfulTransactionsState: 'payment-success',
```
jonheslop: Haha this didnt even last 24 hours 😂

jonheslop:
Ping me and will reapprove after a rebase

PR #760, opened 2018-04-12
PP-3603 my service update ready for direct debit gateway accounts

jonheslop:
- acc07b9 - Sort services by type
  > Layout update of My services in preparation for the incoming DD services
- 2764be3 - refactoring old style controllers
- c9868f1 - Moving Merchant details to work at service level
  > As merchant details are set at the service level it should be shown on the services page. But as it was it would show the active service rather than the service the user thinks they might be editing.
  > I’ve added the externalServiceId to the path so it’s always clear which service to edit.
- b3e4be7 - Updated tests to match updated services page

How it looks now
screen shot 2018-04-12 at 12 12 40


jonheslop:
@simad or @DanailMinchev I am not here tomorrow but either of you are probably best suited to reviewing this :)

app/controllers/my-services/get-index-controller.js
```diff
@@ -0,0 +1,54 @@
+const q = require('q')
```
simad: JS noob question: do we really need to use `q` here, when (I seem to remember) we decided to switch to native promises whenever possible?

PR #763, opened 2018-04-17
PP-3644 - hide settings from DD that arent relevant

jonheslop:
Some settings aren’t relevant to Direct debit this hides them.

### DIFF
--- a/app/utils/display_converter.js
+++ b/app/utils/display_converter.js
@@ -92,8 +92,8 @@ module.exports = function (req, data, template) {
   convertedData.isSandbox = _.get(convertedData, 'currentGatewayAccount.payment_provider') === 'sandbox'
   convertedData.currentServiceName = _.get(req, 'service.name')
   if (permissions) {
-    convertedData.serviceNavigationItems = serviceNavigationItems(originalUrl, permissions)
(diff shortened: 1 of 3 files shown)


app/utils/display_converter.js
```diff
@@ -92,8 +92,8 @@ module.exports = function (req, data, template) {
   convertedData.isSandbox = _.get(convertedData, 'currentGatewayAccount.payment_provider') === 'sandbox'
   convertedData.currentServiceName = _.get(req, 'service.name')
   if (permissions) {
-    convertedData.serviceNavigationItems = serviceNavigationItems(originalUrl, permissions)
-    convertedData.adminNavigationItems = adminNavigationItems(originalUrl, permissions)
+    convertedData.serviceNavigationItems = serviceNavigationItems(originalUrl, permissions, _.get(account, 'paymentMethod', 'card'))
```
simad: Looks good! Have you had a chance to test it locally? I remember there were some weird cases where `paymentMethod` was not correctly populated for direct debit accounts (which is something we might need to fix, in case)

jonheslop: Yeah it worked fine for me with local test/live DD accounts. 
`paymentMethod` is never populated for card gateways which is why i do the lodash get and set it to `'card'`

PR #772, opened 2018-04-19
Pp 3614 publish pact file on pr build

oswaldquek:
Publish pact file on PR build


test/unit/clients/adminusers_client/forgotten_password/create_forgotten_password_test.js
```diff
@@ -13,7 +13,7 @@ const FORGOTTEN_PASSWORD_PATH = '/v1/api/forgotten-passwords'
 
 describe('adminusers client - create forgotten password', function () {
   let provider = Pact({
-    consumer: 'selfservice',
+    consumer: 'selfservice-to-be',
```
mrlumbu: presumably this to-be thing is so as not to pollute contract history before we know we've got things right?

mrlumbu: 👍

PR #792, opened 2018-05-01
PP-3703 base self-service on new govukpay node 8-x image

brid:
## WHAT
Implements self-service as being based on our govukpay/nodejs:8.11.1 Docker image.


mrlumbu:
🎉 🎈

PR #793, opened 2018-05-02
PP-3711 Add email address to dd merchant details

georgeracu:
- Added frontend form field
- Added validation for email field (must be present and correct for DD)
- Handle the response from `adminusers` after update
- Update tests

with @alexbishop1


test/unit/controller/edit_merchant_details_controller/post_test.js
```diff
@@ -186,6 +190,153 @@ describe('edit merchant details controller - post', () => {
       })
     })
   })
+  describe('when the update merchant details call has invalid telephone number', () => {
```
DanailMinchev: Thanks for adding that 👍

test/unit/controller/edit_merchant_details_controller/post_test.js
```diff
@@ -186,6 +190,153 @@ describe('edit merchant details controller - post', () => {
...
+        csrfSecret: '123',
+        12345: {refunded_amount: 5},
+        passport: {
+          user: userInSession
+        },
+        secondFactor: 'totp',
+        last_url: 'last_url',
+        version: 0
+      }
+      let app = mockSession.createAppWithSession(getApp(), session)
```
DanailMinchev: Minor: we should use `const` by default, then if we need to reassign the variable we should use `let` and we should not use `var`. I know the whole file was already using `let` but it's good to update when we touch new files. Thanks

PR #796, opened 2018-05-10
PP-3764 Cypress testing - session handling

brid:
##WHAT

Includes Cypress.io plugin and scaffolding to 'inject' a logged in user session via the use of some injected cookies that bypass a login flow and take the user to authenticated pages, based on stubs provided by contract/pact.

New pacts/interations have been added on connector to return a gateway account with ID = 666, as well as transaction summary values which are requested by the the dashboard on load. To have these stubbed, they need to be created as pact interactions to be picked up by pact-stub.

Note, in order to run these new browser tests locally, the following steps need to take place.
- 'npm test' is first run, which runs unit tests and generates up to date contract/pact files
- 'npm run pact-stub' is run in a new terminal window, which mounts pact files into a container which then stubs all requests made in expected formats, with expected responses. 
- 'npm run server:cypress' is run in a new terminal window. This will now run the self service application, with environment variables pointing the API endpoints at the pact-stub server.
- 'npm run test:cypress' is run in a third terminal window. This will execute Cypress.io which then uses information in the /cypress directory to initialise and run the browser tests


mrlumbu:
I'm getting one error in the cypress tests when I run them 
```
 1) Login Page Form validation Valid submissions should progress to 2FA page if provided valid username and password:
     CypressError: Timed out retrying: expected 'Sign in to GOV.UK Pay' to equal 'Enter security code - GOV.UK Pay'
```
I guess this was already like this, but we should probably start running these tests as part of the pipeline soon to keep us honest

app/controllers/dashboard/dashboard-activity-controller.js
```diff
@@ -9,29 +9,49 @@ const moment = require('moment-timezone')
...
   if (period === 'yesterday') {
     daysAgo = 1
   } else if (period === 'previous-seven-days') {
     daysAgo = 8 // 7+1 because we count starting from yesterday
   } else if (period === 'previous-thirty-days') {
     daysAgo = 31 // 30+1 because we count starting from yesterday
+  } else if (period === 'custom') {
+    fromDateTime = _.get(req, 'query.fromDateTime', null)
+    toDateTime = _.get(req, 'query.toDateTime', null)
+    if (!fromDateTime || !toDateTime || !moment(fromDateTime).isValid() || !moment(toDateTime).isValid()) {
```
mrlumbu: when we discussed this I thought we said we keep it so you can only specify one end of the window, rather than the range, as this opens us up to DB spamming?

app/controllers/dashboard/dashboard-activity-controller.js
```diff
@@ -9,29 +9,49 @@ const moment = require('moment-timezone')
...
 
-  if (period !== 'today') {
-    toDateTime = moment().tz('Europe/London').startOf('day').format()
+  if (period !== 'custom') {
+    fromDateTime = moment().tz('Europe/London').startOf('day').subtract(daysAgo, 'days').format()
+
+    if (period !== 'today') {
+      toDateTime = moment().tz('Europe/London').startOf('day').format()
+    }
   }
```
mrlumbu: I think all this date logic needs to be extracted and unit tested. It's getting a bit convoluted

app/controllers/dashboard/dashboard-activity-controller.js
mrlumbu: Also I think it makes things more readable if you alias long chains of booleans with what they represent eg
```
validFromDate = fromDateTime &&  moment.isValid(fromDateTime)
```

PR #800, opened 2018-05-15
PP-3737 update api key flow

jonheslop:
The API ~tokens~ keys page was weird and used loads of AJAX to do things which was bad for people without JS. Also the controller and tests were massive and not easy to understand.

I’ve tried to make it all a lot simpler although it the methods still have bad names on the public auth API so like delete = revoke and token = key 

• 38af8ac — PP-3737 - Restructure and renamed API key stuff
  > moved from one looong controller to separate files. Also renamed everything so that it’s defined as api-key not ~token~

• 1368146 — PP-3737 - Updating API keys flow
> Simplified the controllers, stripped out all the AJAX stuff and made editing happen on separate page. Wrote JS so that API key can be copied to clipboard. Refactored target-to-show so it works ok with multiple targets per page.

• 5a2897f — PP-3737 - update nav tests to reflect new templates
> Nav tests expected /tokens, updated to /api-keys

• c826c8d — PP-3737 - old tests are so gross easier to delete and start again
> Removed all the old tests as we don’t like that style of tests anymore

• bc0a334 — PP-3737 - Rewriting tests starting with API key index
> Added tests for each of the rewritten controllers.


jonheslop:
Please can someone review 🙏

mrlumbu:
In that first commit the whole package-lock.json changes, but you don't say why. Can you update with some explanation

jonheslop:
@mrlumbu good point, I don’t know why. I think that I should probably not include the package-lock.json. I think when I did a rebase there was a conflict with the package-lock so I regenerated it. But as a package hasnt changed I probably shouldnt include it right?

mrlumbu:
yep agree - unless you are aware of an upgrade that needs to happen, you should not change package-lock.json IMHO

PR #803, opened 2018-05-17
PP-3761 allow users to switch back to use sms for auth

jonheslop:
In the firebreak we enabled users to be able to switch from SMS two factor auth to using an APP. 

But we didn’t let them switch back…

**UNTIL NOW**

## Things changed

* Views now have more logic to show the relevant switch the user is making, including a form with radio buttons to tell our controllers which switch they want to make.
* Controllers were passing routes/links to the view which I recently discovered is redunant as all routes are accessible to all views.
* Controllers understand which auth method users want to change to
* Tests were expanded upon (this is most of the line changes 🙄)


app/controllers/two-factor-auth-controller/get-index-controller.js
```diff
@@ -2,15 +2,9 @@
...
-  const PAGE_PARAMS = {
-    profile: paths.user.profile,
-    index: paths.user.twoFactorAuth.index
-  }
-
-  PAGE_PARAMS.authenticatorMethod = req.user.secondFactor
-
-  return response(req, res, 'twoFactorAuth/index', PAGE_PARAMS)
+  return response(req, res, 'twoFactorAuth/index', {
+    authenticatorMethod: req.user.secondFactor
```
jonheslop: This is the critical change, we used to assume it was always `APP` now, it can be `APP || SMS`

app/controllers/two-factor-auth-controller/post-configure-controller.js
```diff
@@ -2,14 +2,15 @@
...
+const lodash = require('lodash')
 
 // Local dependencies
 const paths = require('../../paths')
 const userService = require('../../services/user_service.js')
 
 module.exports = (req, res) => {
   const code = req.body['code'] || ''
-  userService.configureNewOtpKey(req.user.externalId, code, 'APP', req.correlationId)
+  userService.configureNewOtpKey(req.user.externalId, code, lodash.get(req, 'session.pageData.twoFactorAuthMethod', 'APP'), req.correlationId)
```
jonheslop: This is the critical change, we used to assume it was always `APP` now, it can be `APP || SMS`

PR #805, opened 2018-05-18
PP-3764 Moved date logic out of main dashboard controller logic. Added unit tests to support its operation

brid:
## WHAT
Moves period/custom date logic out of controller logic, tidies up a few things and adds unit tests to support it's operation.

The existing in-line date logic was starting to confuse the legibility the main controller function export, so this has been migrated to it's own private function, tided up to use switch statements and I've shaved things down a touch to hopefully make it a little neater and easier to interpret.

I did consider migrating this to it's own code file but felt the code was small and custom enough in nature to be kept in the controller. Happy to change this if there are conflicting opinions.

The date range logic could have been tested in isolation outside of the main controller but exposing it as a module exports, but this would have broken consistency with other controllers, and it would have been the wrong step (imo) to change code in one instance, or expose something publicly, merely to unit test that small(ish) function.

##TO TEST
In isolation, the new unit tests to support the controller can be modified with '.only' in the following test location:

/app/controllers/dashboard/dashboard-activity-controller.test.js

### DIFF
--- a/app/controllers/dashboard/dashboard-activity-controller.js
+++ b/app/controllers/dashboard/dashboard-activity-controller.js
@@ -14,41 +14,114 @@ const auth = require('../../services/auth_service.js')
 const connectorClient = () => new ConnectorClient(process.env.CONNECTOR_URL)
 const datetime = require('../../utils/nunjucks-filters/datetime')
 
+const getTimespanDays = (fromDateTime, toDateTime) => moment(toDateTime).diff(moment(fromDateTime), 'days')
+
 module.exports = (req, res) => {
-  const correlationId = _.get(req, 'headers.' + CORRELATION_HEADER, '')
   const gatewayAccountId = auth.getCurrentGatewayAccountId((req))
-  let fromDateTime
-  let toDateTime = moment().tz('Europe/London').format() // Today is the default
-  let daysAgo = 0
+
+  const correlationId = _.get(req, 'headers.' + CORRELATION_HEADER, '')
   const period = _.get(req, 'query.period', 'today')
 
-  if (period === 'yesterday') {
-    daysAgo = 1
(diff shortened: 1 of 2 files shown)


app/controllers/dashboard/dashboard-activity-controller.js
```diff
@@ -14,41 +14,114 @@ const auth = require('../../services/auth_service.js')
...
+        response(req, res, 'dashboard/index', {
+          name: req.user.username,
+          serviceId: req.service.externalId,
+          activityError: true,
+          period
+        })
+      })
+  } catch (err) {
+    logger.error(`[${correlationId}] ${err.message} -`, {
+      service: 'frontend',
```
jonheslop: is the service meant to be selfservice?

app/controllers/dashboard/dashboard-activity-controller.js
```diff
@@ -59,51 +132,5 @@ module.exports = (req, res) => {
...
-      })
-      res.status(status)
-      response(req, res, 'dashboard/index', {
-        name: req.user.username,
-        serviceId: req.service.externalId,
-        activityError: true,
-        period
-      })
-    })
+  return {fromDateTime: fromDateTime, toDateTime: toDateTime}
```
jonheslop: As key has the same name as the variable could use shorthand
```js
   return {fromDateTime, toDateTime} 
```

app/controllers/dashboard/dashboard-activity-controller.js
brid: In this instance, the service we call is connector, so that error message is flagged as relating to connector. There is a general catch block for the controller itself that sends back an error of 'self-service' for internal exceptions.

PR #806, opened 2018-05-21
PP 3812 replicate transactions search by dates.java in cypress

brid:
## WHAT

- moves cypress inside test/ directory, and moves config
- adds single .json configuration to create unit/contract test fixtures, and then as basis for browser tests
- replicates the spirit of the e2e test for date filtering of the transactions page, while adhering to our new principles for contract/browser tests with respect to not writing contracts/pacts/browser tests for complex combinations of search terms, interactivity and the like
- adds 3 pact interactions and unit tests for transactions, for situations with no filtering, and to/from dates specified
- adds a pact interaction and unit test for card types


test/fixtures/config/self_service_user.json
```diff
@@ -0,0 +1,22 @@
+{
+  "description": "defines agreed fixture config that can be relied upon by browser testing tools requiring defined pact/contract data",
+  "config": {
+    "users": [
+      { "isPrimary": "true",
```
mrlumbu: what does isPrimary mean?

brid: This is just means the primary user as far as default contract tests are based around. If we wanted to add other users to test other operations, we could use other filters to include/exclude this user. It doesn't relate to anything other than the fixture config (i.e. has no meaning/purpose outside of it).

Was a first crack at how to do a fixed config, so happy to change it.

test/cypress/integration/user/dashboard_spec.js
```diff
@@ -1,14 +1,22 @@
...
+
+    // Use a known configuration used to generate contracts/stubs.
+    // This is also used to generate the session/gateway_account cookies
+    const ssUser = ssDefaultUser.config.users.filter(fil => fil.isPrimary === 'true')[0]
+
     // Note : these from/to datetime strings exactly match those in the pact/contract, so are essential to match against stubs
     // Either change everything together, or map these do a single place like a .json document so the contracts/tests refer to one place
-    const from = encodeURIComponent('2018-05-14T00:00:00+01:00')
-    const to = encodeURIComponent('2018-05-15T00:00:00+01:00')
+    const from = encodeURIComponent(ssUser.sections.dashboard.transaction_summary.from_date)
```
mrlumbu: not a big fan of unnecessary contractions like ss for selfservice. would a bit of finding and replacing be too much to ask, unless there's a good reason for it

test/unit/clients/connector_client/connector_get_transactions_and_summary_test.js
```diff
@@ -0,0 +1,254 @@
...
+    const validGetFilteredTransactionsResponse = transactionSummaryFixtures.validTransactionsResponse(params)
+
+    // Stop the transactions data being flowed through into anything else
+    delete params.transactions
+
+    before((done) => {
+      const pactified = validGetFilteredTransactionsResponse.getPactified()
+      provider.addInteraction(
+        new PactInteractionBuilder(`${TRANSACTIONS_RESOURCE}/${params.gatewayAccountId}/charges`)
+          .withUponReceiving('a valid transactions request filtered by from_date only')
```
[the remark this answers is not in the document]
mrlumbu: having discussed, the thing that is missing here is `.withState` denoting the required state of the provider prior to the contract being tested. As discussed, we should add these in now, and do sme work to figure out best practice on how these should work, including talking to the Pact folk

PR #819, opened 2018-05-29
PP 3812 replicate transactions search by dates.java in cypress

brid:
## WHAT

- moves cypress inside test/ directory, and moves config
- adds single .json configuration to create unit/contract test fixtures, and then as basis for browser tests
- replicates the spirit of the e2e test for date filtering of the transactions page, while adhering to our new principles for contract/browser tests with respect to not writing contracts/pacts/browser tests for complex combinations of search terms, interactivity and the like
- adds 3 pact interactions and unit tests for transactions, for situations with no filtering, and to/from dates specified
- adds a pact interaction and unit test for card types


mrlumbu:
For context, discussion of this work is here https://github.com/alphagov/pay-selfservice/pull/806

PR #835, opened 2018-06-14
PP-3888 - Added overview page when details have been added

jonheslop:
Rather than showing the form filled in which is U G L Y, no offence, we
now show a summary page if the details have been entered already.

Also Merchant is jargon so renamed it to Organisation, but only for users, still called merchant in the background. I did start renaming it there too, but it got crazy! Especially seeing as to do it properly would have to rename in the java projects and on the DB too

### DIFF
--- a/test/unit/controller/edit_merchant_details_controller/get-index.test.js
+++ b/test/unit/controller/edit_merchant_details_controller/get-index.test.js
@@ -1,0 +1,243 @@
+const chai = require('chai')
+const cheerio = require('cheerio')
+const nock = require('nock')
+const mockSession = require('../../../test_helpers/mock_session.js')
+const getApp = require('../../../../server.js').getApp
+const supertest = require('supertest')
+const userFixtures = require('../../../fixtures/user_fixtures')
+const paths = require('../../../../app/paths.js')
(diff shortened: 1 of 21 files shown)


test/unit/controller/edit_merchant_details_controller/get-index.test.js
```diff
@@ -0,0 +1,243 @@
...
+        })
+    })
+    it(`should get a nice 302 status code`, () => {
+      expect(response.statusCode).to.equal(302)
+    })
+    it('should redirect to the edit page', () => {
+      expect(response.headers).to.have.property('location').to.equal(formattedPathFor(paths.merchantDetails.edit, EXTERNAL_SERVICE_ID))
+    })
+  })
+})
```
mrlumbu: cypress tests anyone?

jonheslop: 😬

jonheslop:
This PR is on ice, it adds Cypress tests but they don’t play nicely with Pact, so we need to think about it futher, so I created a copy of the PR #857 with the meaningful changes and we can come back to this PR when we resolve the Cypress/Pact issues

PR #842, opened 2018-06-25
PP-2684 add feedback form

jonheslop:
This makes it easier for our users to feedback on their experiences with GOV.UK Pay

Feedback is submitted straight to Zendesk
- it is marked as a question to the GOV.UK Pay group within the GDS organisation
- it is assigned to @soniaturcotte (she may regret this)
- The users current service name, ID and gateway account are prepended to the message so we can better help them with their query

### DIFF
--- a/app/controllers/feedback/post-index-controller.js
+++ b/app/controllers/feedback/post-index-controller.js
@@ -1,0 +1,9 @@
+'use strict'
+
+// Local dependencies
+const paths = require('../../paths')
+
+module.exports = (req, res) => {
+  req.flash('generic', `<h2>Thanks for your feedback</h2>`)
+  return res.redirect(paths.feedback)
+}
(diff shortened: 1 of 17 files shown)


app/controllers/feedback/post-index-controller.js
```diff
@@ -0,0 +1,56 @@
...
+      requester: {
+        email: req.body.email,
+        name: '(no name supplied)'
+      },
+      type: 'question',
+      subject: `Feedback from: ${req.body['service-name']}`,
+      comment: {
+        body: message
+      },
+      group_id: 20487877,
```
tlwr: perhaps put these somewhere

```
config/zendesk.js
module.exports = {
  GROUP_ID: 20487877,
  ORG_ID: 21891972,
  ASSIGNEE_ID: 360607861465,
}

```

curious as to your thoughts, but I won't block the PR for this as they are only used here.

jonheslop: Yeah that probably a good idea. Will change, got some weird test bug I need to sort anyway

PR #844, opened 2018-06-26
PP-3920 remove q from all node projects

brid:
## WHAT

Performs the not easy task of removing 'q' from self-service, which was an early implementation of ECMAScript Promises before they were natively supported by the version of Node.js we were running in Production.

Unfortunately these were not a simple swap out for regular native promises as the implementation was not like for like, meaning the code had to be re-factored to a simpler version.

Part of this involved swapping out the use of the 'old' base client for the 'new' one which is still currently only half adopted in Self-Service. For various reasons (mostly complexity) it doesn't make sense to completely implement it in this PR as it only directly applied to the Adminusers client. 

Moving over to a new base client had the knock on effect of affecting unit tests as the new base client uses 'requestretry', so some work around how that module works in test/production needed to be added.

The method of error handling between base clients varies subtly, and has knock on effects on how these are handled in controllers and views.

I'd advise a full migration to a single base client is picked up in another PR. 

Slightly controversial is the removal of 'npm run lint' from the build and test docker shell file. As our legacy 'q' Promise integration did not perform rejections with Error objects, these are now flagged up by the linter as failing that check. I'm slightly less inclined to go and include the multiple fixes to ensure these objects are handled by whatever they are raised to, than I am to disable linting on this PR and then do those fixes in another (shorter) PR.

### DIFF
--- a/app/middleware/otp_verify.js
+++ b/app/middleware/otp_verify.js
@@ -36,7 +36,9 @@ function verifyOtpForServiceInvite (req, res, next) {
   }
 
   return validateOtp(otpCode)
-    .then(() => registrationService.submitServiceInviteOtpCode(code, otpCode, correlationId))
+    .then(() => {
+      return registrationService.submitServiceInviteOtpCode(code, otpCode, correlationId)
+    })
     .then(next)
     .catch(err => {
       logger.warn(`[requestId=${req.correlationId}] Invalid invite code attempted ${req.code}, error = ${err.errorCode}`)
--- a/app/utils/reflect.js
+++ b/app/utils/reflect.js
@@ -1,0 +1,12 @@
+'use strict'
+
+module.exports = {
+  reflect: promise => {
+    return promise.then(function (v) {
+      return {v: v, status: 'resolved'}
+    },
+    function (e) {
+      return {e: e, status: 'rejected'}
+    })
+  }
+}
--- a/app/utils/reflect.js
+++ b/app/utils/reflect.js
@@ -2,11 +2,8 @@
 
 module.exports = {
   reflect: promise => {
-    return promise.then(function (v) {
-      return {v: v, status: 'resolved'}
-    },
-    function (e) {
-      return {e: e, status: 'rejected'}
-    })
+    return promise.then(result => ({v: result, status: 'resolved'}),
+      error => ({e: error, status: 'rejected'})
+    )
   }
 }
--- a/app/controllers/forgotten_password_controller.js
+++ b/app/controllers/forgotten_password_controller.js
@@ -1,79 +1,84 @@
+'use strict'
+
+// Local dependencies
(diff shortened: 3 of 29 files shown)


app/controllers/invite_validation_controller.js
```diff
@@ -85,6 +85,6 @@ module.exports = {
 
     return validateInviteService.getValidatedInvite(code, correlationId)
       .then(processAndRedirect)
-      .catch((err) => handleError(req, res, err))
+      .catch((err) => { handleError(req, res, err) })
```
jonheslop: Why the brackets?

brid: Good spot. Fixed this.

app/middleware/otp_verify.js
```diff
@@ -36,7 +36,9 @@ function verifyOtpForServiceInvite (req, res, next) {
   }
 
   return validateOtp(otpCode)
-    .then(() => registrationService.submitServiceInviteOtpCode(code, otpCode, correlationId))
+    .then(() => {
+      return registrationService.submitServiceInviteOtpCode(code, otpCode, correlationId)
+    })
```
jonheslop: Why is this better? More verbose?

brid: In order for it to work against the promise 'catch' statement in that promise chain, I had to investigate why the previous incarnation wouldn't work with native promises, so had to explicitly do it otherwise unit tests that rejected the underlying promise were never caught.

app/utils/reflect.js
```diff
@@ -0,0 +1,12 @@
...
+module.exports = {
+  reflect: promise => {
+    return promise.then(function (v) {
+      return {v: v, status: 'resolved'}
+    },
+    function (e) {
+      return {e: e, status: 'rejected'}
+    })
+  }
+}
```
jonheslop: Would this be too verbose?
```js
module.exports = {
  reflect: promise => {
    return promise.then(result => ({ result, status: 'resolved' }),
      error => ({ error, status: 'rejected' })
    )
  }
}
```

jonheslop: I think it probably is :) looking at it

jonheslop: What about 

```js
'use strict'

module.exports = {
  reflect: promise => {
    return promise.then(result => {
      return { result, status: 'resolved' }
    },
    error => {
      return { error, status: 'rejected' }
    })
  }
}
```

Feel free to totally ignore me though

brid: Agree that looks better, I've updated the code with that implementation.

tlwr:
absolute unit of a pr

i've had a cursory look but not a big review, however this is valuable for AWS x-ray so we have fewer things to shim

tlwr:
linting changes seem sensible to pick up in another pr

as does a singular base client (again useful for xray)

PR #848, opened 2018-06-27
PP-2684 zendesk integration proxy fix

jonheslop:
Zendesk request was stuck inside the box this sends it throught the
proxy, will still work locally where we don't need the proxy

@jonheslop with A LOT of help from @tlwr

### DIFF
--- a/app/views/layout.njk
+++ b/app/views/layout.njk
@@ -37,7 +37,7 @@
       </div>
     </main>
 
-    {% if hideFeedbackBanner %}
+    {% if not hideFeedbackBanner %}
       {% include "includes/feedback-banner.njk" %}
     {% endif %}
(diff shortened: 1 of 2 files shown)


tlwr:
Cheeky little undocumented opt: https://github.com/blakmatrix/node-zendesk/blob/3df49a8e3ffc473fc3c0c23e99fc42419d40d2fb/lib/client/client.js#L39

app/views/layout.njk
```diff
@@ -37,7 +37,7 @@
       </div>
     </main>
 
-    {% if hideFeedbackBanner %}
+    {% if not hideFeedbackBanner %}
```
mrlumbu: 😄

PR #849, opened 2018-06-28
BAU: Remove Sonia as assignee

tlwr:
@jonheslop _sigh_

solo @tlwr


tlwr:
apparently this was intended

PR #854, opened 2018-07-03
PP-3943 Do not show invite list if empty and tweak ui

simad:
## WHAT
- The invite list is shown only if there are invites
- Changed spacing between elements
- Also removed two `console.log`. Oooops.

### DIFF
--- a/app/views/services/team_members.njk
+++ b/app/views/services/team_members.njk
@@ -21,13 +21,13 @@ Team members - GOV.UK Pay
     {% endif %}
   </div>
 </div>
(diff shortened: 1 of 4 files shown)


app/views/services/team_members.njk
```diff
@@ -90,7 +90,8 @@ Team members - GOV.UK Pay
       </ul>
     </div>
   </div>
-  <div class="column-full admin-list-group">
+  {% if number_invited_members > 0 %}
```
jonheslop: 👏

PR #855, opened 2018-07-04
PP-3928 Reduce the number of outbound requests to get gateway account information

brid:
## WHAT
Previously the services index/switcher page made an initial call to Adminusers to get a list of services with associated gateway ids against them. These were composed into a large collection of Promises which made individual calls to Connector or Direct Debit Connector to get information about external gateways.

In the case where a development/test system has a few services, with perhaps one test gateway account under each one, this resulted in an inefficient but still fairly fast response time. As the number of services/accounts increases, the page starts to considerably slow down, as evident in our production systems.

The new implementation aggregates all gateway ids across all services and requests them in one go, before matching these up against individual services and rendering them on the page. To support this, new logic was added to connector and direct debit connector for a modified API call. As a result the maximum number of requests to connector/dd-connector is 1 per page load.

### DIFF
--- a/app/controllers/my-services/get-index-controller.js
+++ b/app/controllers/my-services/get-index-controller.js
@@ -5,7 +5,6 @@ const _ = require('lodash')
 
 // local dependencies
 const {response} = require('../../utils/response')
-const {reflect} = require('../../utils/reflect')
 const serviceService = require('../../services/service_service')
 const getHeldPermissions = require('../../utils/get_held_permissions')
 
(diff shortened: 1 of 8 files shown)


app/controllers/my-services/get-index-controller.js
```diff
@@ -29,32 +29,24 @@ module.exports = (req, res) => {
...
-              cardAccounts,
-              directdebitAccounts
-            },
-            permissions: getHeldPermissions(serviceRole.role.permissions.map(permission => permission.name))
-          })
-        })
-        .catch(() => reject(new Error('Error getting gateway account')))
+  const aggregatedGatewayAccountIds = servicesRoles
+    .map(servicesRole => servicesRole.service.gatewayAccountIds)
+    .reduce((acc, cv) => acc.concat(cv))
```
mrlumbu: is this known idiomatic js? I would prefer proper names `accumulator` `currentValue` - we don't have to optimise the size of our files these days

app/controllers/my-services/get-index-controller.js
```diff
@@ -29,32 +29,24 @@ module.exports = (req, res) => {
...
+
+  serviceService.getGatewayAccounts(aggregatedGatewayAccountIds, req.correlationId)
+    .then(aggregatedGatewayAccounts => {
+      return servicesRoles.map(serviceRole => {
+        const gatewayAccounts = aggregatedGatewayAccounts.filter(fil => serviceRole.service.gatewayAccountIds.includes(fil.id.toString()))
```
mrlumbu: I think `fil` should be changed to `gatewayAccount`, for that is what it is

app/controllers/my-services/get-index-controller.js
```diff
@@ -29,32 +29,24 @@ module.exports = (req, res) => {
...
+        return {
+          name: serviceRole.service.name === 'System Generated' ? 'Temporary Service Name' : serviceRole.service.name,
+          external_id: serviceRole.service.externalId,
+          gateway_accounts: {
+            cardAccounts: gatewayAccounts.filter(fil => fil.payment_method !== 'direct debit'),
```
mrlumbu: ditto

app/controllers/my-services/get-index-controller.js
```diff
@@ -29,32 +29,24 @@ module.exports = (req, res) => {
...
+            directdebitAccounts: gatewayAccounts.filter(fil => fil.payment_method === 'direct debit')
```
mrlumbu: and again

app/services/service_service.js
```diff
@@ -28,31 +29,37 @@ module.exports = {
...
 
-  return Promise.all([fetchCardGatewayAccounts, fetchDirectDebitGatewayAccounts]
-    .map(promise => promise))
-    .then(results => results
-      .reduce((a, b) => a.concat(b))
-      .filter(p => !(p instanceof Error)))
+  return Promise.all([fetchCardGatewayAccounts, fetchDirectDebitGatewayAccounts])
+    .then(results => {
+      return results
+        .reduce((acc, cv) => {
```
mrlumbu: as above , can these be full names

test/unit/controller/service_switch_controller_test.js
```diff
@@ -27,29 +27,27 @@ describe('service switch controller: list of accounts', function () {
     renderSpy = sinon.spy()
   })
 
-  it('should render a list of services when user has multiple services', function (done) {
+  it.only('should render a list of services when user has multiple services', function (done) {
```
mrlumbu: naughty

brid: Whoops. I should have had [WIP] on this PR the whole time as its being actively worked on.

tlwr:
directdebit connector accounts endpoint needs `externalAccountIds` instead of `accountIds`

PR #857, opened 2018-07-05
PP-3888 merchant details improvements 

jonheslop:
_This is a clone of https://github.com/alphagov/pay-selfservice/pull/835 but without the Pact/Cypress stuff which broke things, that we will come back to later_

Rather than showing the form filled in which is U G L Y, no offence, we
now show a summary page if the details have been entered already.

Also Merchant is jargon so renamed it to Organisation, but only for users, still called merchant in the background. I did start renaming it there too, but it got crazy! Especially seeing as to do it properly would have to rename in the java projects and on the DB too

### DIFF
--- a/app/views/merchant_details/edit_merchant_details.njk
+++ b/app/views/merchant_details/edit_merchant_details.njk
@@ -1,7 +1,7 @@
 {% extends "../layout.njk" %}
 
 {% block page_title %}
-  Merchant details - {{currentService.name}} {{currentGatewayAccount.full_type}} - GOV.UK Pay
+  Organisation details - {{currentService.name}} {{currentGatewayAccount.full_type}} - GOV.UK Pay
 {% endblock %}
 
 {% block mainContent %}
@@ -11,7 +11,7 @@
 <div class="column-two-thirds">
   {% if success %}
   <div class="notification">
-      Merchant details updated
+      Organisation details updated
   </div>
   {% endif %}
   {% if errors %}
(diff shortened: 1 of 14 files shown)


app/views/merchant_details/edit_merchant_details.njk
```diff
@@ -11,7 +11,7 @@
 <div class="column-two-thirds">
   {% if success %}
   <div class="notification">
-      Merchant details updated
+      Organisation details updated
```
simad: Yay, happy to see the naming change

app/views/merchant_details/merchant_details.njk
```diff
@@ -0,0 +1,80 @@
...
+  Organisation details - {{currentService.name}} {{currentGatewayAccount.full_type}} - GOV.UK Pay
+{% endblock %}
+
+{% block mainContent %}
+<div class="column-full">
+  <a href="/my-services" class="link-back">Back to services</a>
+</div>
+<div class="column-two-thirds">
+  <h1 class="page-title">Organisation details</h1>
+  {% if has_direct_debit_gateway_account %}
```
simad: This logic will create problems. Let's say that a service as both a card and dd gateway account, they will see all of three descriptions (the dd, card, and also the "has both accounts" text), which is not great. 
A solution would be to move the `{% if has_card_and_dd_gateway_account %}` first, and use else if for the subsequent two conditions

PR #860, opened 2018-07-06
Update 404 content

soniaturcotte:
@mpwoods

## WHAT
_A few small content tweaks and adding link to documentation_


tlwr:
#862 so it gets built by pr builder

PR #866, opened 2018-07-10
PP 3980 replace smartpay refund fail e2e test

brid:
## WHAT
Replaces the test in pay-endtoend which asserts that a refund attempt over the available refund amount, fails.

In order for this to work as a contract/browser test, the following needed do be done:

1. creating a pact interaction and unit test for:

a) viewing a transaction details page
b) posting an invalid refund request

2. creating Cypress browser tests to replicate the test from e2e

### DIFF
--- a/test/cypress/integration/user/dashboard_spec.js
+++ b/test/cypress/integration/user/dashboard_spec.js
@@ -1,5 +1,5 @@
 describe('Dashboard', () => {
-  const selfServceUsers = require('../../../fixtures/config/self_service_user.json')
+  const selfServiceUsers = require('../../../fixtures/config/self_service_user.json')
 
   beforeEach(() => {
     cy.setCookie('session', Cypress.env('encryptedSessionCookie'))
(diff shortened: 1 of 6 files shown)


test/cypress/integration/user/dashboard_spec.js
```diff
@@ -1,5 +1,5 @@
 describe('Dashboard', () => {
-  const selfServceUsers = require('../../../fixtures/config/self_service_user.json')
+  const selfServiceUsers = require('../../../fixtures/config/self_service_user.json')
```
jonheslop: took me ages to spot that 🙄

brid: There's a new sheriff in town.

test/fixtures/config/self_service_user.json
```diff
@@ -27,6 +29,15 @@
...
             "name": "transactions_card_type_read"
+          },
+          {
+            "name": "transactions-details:read"
+          },
+          {
+            "name": "refunds_create"
+          },
+          {
+            "name": "refunds:create"
```
jonheslop: do we need this twice?

PR #870, opened 2018-07-11
PP-3990 Prevents 'to-be' pacts from being published by changing their file extension before/after being published

brid:
##WHAT

The naming convention 'to-be' is used in our contract testing pact/contract interaction stage to define an interaction where we don't quite have the provider contract tests ready so the two can be tested in unison.

This PR ensures that 'to-be' pacts are not published to the broker, which ensures they aren't ran on the provider as expectations, which would normally fail.

This is done by simply renaming the pact files with 'to-be' in their filename to a different file extension, publishing the pacts, and then changing them back.

### DIFF
--- a/docker/build_and_test.sh
+++ b/docker/build_and_test.sh
@@ -6,4 +6,6 @@ npm run compile
 # disabled while promise rejections not returning error objects need to be addressed
 # npm run lint
 npm test -- --forbid-only --forbid-pending
+# prevent publish of any pacts with 'to-be' in their name
+for i in ./pacts/*-to-be-*.json; do mv "$i" "${i%.json}.ignore"; done
 npm run publish-pacts


docker/build_and_test.sh
```diff
@@ -6,4 +6,8 @@ npm run compile
 # disabled while promise rejections not returning error objects need to be addressed
 # npm run lint
 npm test -- --forbid-only --forbid-pending
+# prevent publish of any pacts with 'to-be' in their name
+for i in ./pacts/*-to-be-*.json; do mv "$i" "${i%.json}.ignore"; done
 npm run publish-pacts
+# restore 'to-be' pacts
+for i in ./pacts/*.ignore; do mv "$i" "${i%.ignore}.json"; done
```
oswaldquek: do we need to restore the pacts?
can you verify the publish pact job re-creates a new /pact folder on each build - if so we dont need to restore the 'to-be' pacts.

brid: My original rationale was that the pacts may be needed for stub purposes later on, but these happen outside of the docker container, so you're right. I'll remove the restore as it achieves nothing.

PR #880, opened 2018-08-09
 BAU: Remove Sonia as assignee 

soniaturcotte:
When someone submits feedback in the admin site, it automatically creates and ticket and assigns it to Sonia.

The tickets should no longer be assigned, but picked up in regular support rota.


tlwr:
This PR won't build until someone re-raises it

Thanks for volunteering @rhowe-gds

jonheslop:
@rhowe-gds too slow 🐢 

I am gonna do it

PR #882, opened 2018-08-09
PP-4043 Link GoCardless merchant to GOV.UK Pay Partner app

georgeracu:
## WHAT
Link GoCardless account to GOV.UK Pay

- Added middleware to handle user redirection to GoCardless Connect portal
- Added middleware to handle GET request from GoCardless after successful OAuth
- Added GoCardless Connect client to initiate OAuth
- Updated Direct Debit Connector client to send a request that will exchange
a GoCardless access code with an access token

with @SandorArpa


test/test.env
```diff
@@ -15,3 +15,9 @@ LOGIN_ATTEMPT_CAP=2
 DISABLE_REQUEST_LOGGING=true
 NODE_TEST_MODE=true
 NODE_ENV=test
+GOCARDLESS_TEST_CLIENT_ID=cjchuyewhdj23d8279dh29test
+GOCARDLESS_TEST_CLIENT_SECRET=nco2i3914yu1icbckjbeco3test
+GOCARDLESS_LIVE_CLIENT_ID=acsfhuyewbvd3d8279dh29live
+GOCARDLESS_LIVE_CLIENT_SECRET=saq2i3914yu1icbckjbeco3live
```
DanailMinchev: Where do we use these values? I suppose they are not valid secrets?

georgeracu: Nope, these are some values added by me, probably should have been easier for `test-client-id`. Sorry about the confusion

DanailMinchev: Cool, could you change that to something like:
```
GOCARDLESS_TEST_CLIENT_ID=fake-client-id-test
GOCARDLESS_TEST_CLIENT_SECRET=fake-client-secret-test
GOCARDLESS_LIVE_CLIENT_ID=fake-client-id-live
GOCARDLESS_LIVE_CLIENT_SECRET=fake-client-secret-live
```
Thanks

test/ui/navigation_ui_tests.js
```diff
@@ -234,6 +237,22 @@ describe('navigation menu', function () {
 
     const body = renderTemplate('api-keys/index', templateData)
 
-    body.should.containSelector('.settings-navigation li').withExactText('API keys')
+    body.should.containSelector('.settings-navigation li:nth-child(1)').withExactText('API keys')
+    body.should.containSelector('.settings-navigation li:nth-child(2)').withExactText('Link GoCardless Merchant Account')
+  })
+
+  it('should render Link GoCardless Merchant Account navigation link when user have tokens read permission', function () {
```
DanailMinchev: Great that you have separate test only for GC link 👍

app/routes.js
```diff
@@ -298,6 +305,9 @@ module.exports.bind = function (app) {
   app.get(paths.feedback, xraySegmentCls, hasServices, resolveService, getAccount, feedbackCtrl.getIndex)
   app.post(paths.feedback, xraySegmentCls, hasServices, resolveService, getAccount, feedbackCtrl.postIndex)
 
+  // Partner app link GoCardless account
+  app.get(paths.partnerApp.linkAccount, xraySegmentCls, getAccount, goCardlessRedirect.index)
```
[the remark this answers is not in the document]
georgeracu: This is something that if I don't add to the route, then I hit the problems where NodeJS crashes in tests. We need to investigate when @brid is back

app/paths.js
```diff
@@ -141,5 +141,9 @@ module.exports = {
     editAmount: '/create-payment-link/manage/edit/amount/:productExternalId'
   },
   feedback: '/feedback',
+  partnerApp: {
+    linkAccount: '/link-account',
```
DanailMinchev: We currently don't test navigating directly to URLs without valid session (as it is partly handled in other tests), but if you add tests for this case then it will be awesome!
Not sure how much time it will take as well, so something to discuss and clarify if we need it.

test/unit/middleware/partnerapp/handle_gocardless_connect_get_test.js
```diff
@@ -0,0 +1,83 @@
...
+      }
+    }
+
+    res = {
+      end: () => {
+        expect(true).to.equal(true)
+      }
+    }
+
+    stubbedBaseClientPost = baseClient.post = sinon.stub()
```
DanailMinchev: I like the initialisation of the two variables in one line :) 👍 
I guess we are creating `stubbedBaseClientPost` just for better readability

app/routes.js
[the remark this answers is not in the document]
georgeracu: I didn't have it initially and NodeJS was crashing during tests, so I added. Maybe we need to speak to @brid about it

test/test.env
georgeracu: Nope, no valid secrets. Replaced them with explicit values stating they are not secret, sorry about the confusion.

app/paths.js
georgeracu: I believe some of this functionality is tested when we set a valid request with account, session etc. but we can sit and see if we can find a way to test all the paths.

test/unit/middleware/partnerapp/handle_gocardless_connect_get_test.js
georgeracu: Yep, for readability, can be removed, depends on taste :)

app/paths.js
georgeracu: I changed the tests for middleware to include the routing bit as well and now there are clear tests asserting on user being redirected to login if not authenticated.

PR #885, opened 2018-08-16
PP-4133 Tidy up x-ray implementation

brid:
## WHAT
Tidies up some naming of namespace/key names, as well as ensuring 'continuation-local-storage' is explicitly defined as a dependency in package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "client-sessions": "^0.8.0",
     "common-password": "^0.1.2",
     "connect-flash": "^0.1.1",
+    "continuation-local-storage": "^3.2.1",
     "cookie-parser": "1.4.x",
     "correlation-id": "^2.1.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -55,6 +55,7 @@
     "client-sessions": "^0.8.0",
     "common-password": "^0.1.2",
     "connect-flash": "^0.1.1",
+    "continuation-local-storage": "^3.2.1",
```
DanailMinchev: Do we need to regenerate and commit `package-lock.json` as well?

jonheslop: Whoops @DanailMinchev is right

brid: Well spotted. I'll update that now.
Until we use 'npm ci' it's not actually used in the manner we want it to, but we should get into the habit of it none the less.

PR #891, opened 2018-08-31
PP-4096 - if checkbox not included then set welsh name to blank

jonheslop:
If checkbox not included then set welsh name to blank, then pass the blank string onto adminusers (also fix some linting errors).


app/services/clients/adminusers_client.js
```diff
@@ -567,23 +567,21 @@ module.exports = function (clientOptions = {}) {
   const updateServiceName = (serviceExternalId, serviceName, serviceNameCy) => {
     const body = []
     if (serviceName) {
```
simad: If this service is always going to be there (because it is populated before this call, we can remove this check.
The default is 'system generated' so we can put that in here, as `value: serviceName || 'System Generated'` so that it is immediately clear to whoever reads the code what's going on with the service name

PR #893, opened 2018-09-03
PP-4111 Refund, confirmation and email collection mode modifications

brid:
This PR adds the ability to, in line with changes merged into connector with the same branch jira ticket, extended functionality around the management of emails.

- Email collection mode can now be set to mandatory/optional/off
- Refund emails can be toggled off and on
- Confirmation emails can be toggled off and on
- Confirmation emails are now formatted correctly, where previously the would not show formatting in previews (but worked fine when sent for real).


app/controllers/email_notifications/email_notifications_controller.js
```diff
@@ -0,0 +1,170 @@
...
+
+// Local dependencies
+const response = require('../../utils/response.js').response
+const Email = require('../../models/email.js')
+const paths = require('../../paths.js')
+const CORRELATION_HEADER = require('../../utils/correlation_header.js').CORRELATION_HEADER
+const formattedPathFor = require('../../../app/utils/replace_params_in_path')
+
+// Constants
+const emailConfirmationToggleSubmitLink = formattedPathFor(paths.emailNotifications.confirmation)
```
jonheslop: All paths are [by default shared to the template](https://github.com/alphagov/pay-selfservice/blob/master/server.js#L61) so you don’t need to pass them in the controller manually.

You can just use {{ routes.emailNotifications.confirmation }} in the Nunjucks

(I only realised this recently and have been using is since — https://github.com/alphagov/pay-selfservice/blob/master/app/views/payment-links/review.njk#L11-L12

app/models/email.js
```diff
@@ -23,79 +24,121 @@ module.exports = function (correlationId) {
...
 
   return {
     get: get,
-    update: update,
-    setEnabled: setEnabled
+    updateConfirmationTemplate: updateConfirmationTemplate,
+    setEmailCollectionMode: setEmailCollectionMode,
+    setConfirmationEnabled: setConfirmationEnabled,
+    setRefundEmailEnabled: setRefundEmailEnabled
   }
```
jonheslop: Could use shorthand here:
```js
return {
    get,
    update,
    setEnabled
    updateConfirmationTemplate,
    setEmailCollectionMode,
    setConfirmationEnabled,
    setRefundEmailEnabled
  }
```

app/controllers/email_notifications/email_notifications_controller.js
jonheslop: Also meant to say you don’t need to use `formattedPathFor` unless the path actually needs formatting aka when it looks something like this `show: '/service/:externalServiceId/team-member/:externalUserId'`

package.json
```diff
@@ -35,7 +35,7 @@
     "watch-live-reload": "./node_modules/.bin/grunt watch",
     "lint": "./node_modules/.bin/standard",
     "lint-sass": "./node_modules/.bin/sass-lint -v",
-    "test": "npm run snyk-protect && rm -rf ./pacts && ./node_modules/mocha/bin/mocha '!(node_modules)/**/*+(.|_)+(test|tests)'.js",
```
jonheslop: I dont think you meant to check this change in 😬

PR #896, opened 2018-09-05
PP-4096 - update text to reccomendations from content designers

jonheslop:


karlchillmaid:
Spoke to @katiebates about the link to documentation last week and we decided to remove it because:

- it doesn't exist yet
- even when it does, it breaks the journey to link out to it mid-interaction

jonheslop:
@karlchillmaid ah ok, updated

app/views/services/edit_service_name.njk
```diff
@@ -33,21 +33,31 @@ Edit service name - GOV.UK Pay
         <input class="form-control" data-module="" id="service-name" name="service-name" type="text" value="{{current_name.en}}" data-validate="required isFieldGreaterThanMaxLengthChars" data-validate-max-length="50">
       </div>
 
+      {% if current_name.cy %}
+      <div class="form-group {% if errors.service_name %}error{% endif %}">
+        <label class="form-label" for="service-name">Welsh (Cymraeg) service name</label>
```
simad: I saw that you deleted (Cymraeg) in add service, is this supposed to be here?

PR #899, opened 2018-09-10
PP-3816 Create and delete a products token with payment links

simad:
## WHAT
 - when creating a payment link, we now create a special token of type `PRODUCTS`. This will not be displayed in the API tokens section of selfservice so it won't be possible to manually revoke it
 - The only way to revoke this token is deleting the payment link


jonheslop:
👀

PR #915, opened 2018-09-20
Pp 4109 add delayed capture to transaction details

georgeracu:
### What

- Added a label that displays 'Delayed capture: ON' when the charge has
delayed capture enabled. Otherwise it doesn't show anything different
- Added test data for Pact interactions
- Added cypress test and unit test

with @danworth

### DIFF
--- a/app/views/transaction_detail/_details.njk
+++ b/app/views/transaction_detail/_details.njk
@@ -53,7 +53,7 @@
   {% if delayed_capture %}
   <tr>
     <td>Delayed capture:</td>
-    <td id="delayed-capture-id">ON</td>
+    <td id="delayed-capture-id">On</td>
   </tr>
   {% endif %}
   </tbody>
--- a/app/views/transaction_detail/_details.njk
+++ b/app/views/transaction_detail/_details.njk
@@ -53,7 +53,7 @@
   {% if delayed_capture %}
   <tr>
     <td>Delayed capture:</td>
-    <td id="delayed-capture-id">On</td>
+    <td id="delayed-capture">On</td>
   </tr>
   {% endif %}
   </tbody>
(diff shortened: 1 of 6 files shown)


app/views/transaction_detail/_details.njk
```diff
@@ -49,5 +49,12 @@
     <td>GOV.UK payment ID:</td>
     <td id="payment-id">{{charge_id}}</td>
   </tr>
+
+  {% if delayed_capture %}
+  <tr>
+    <td>Delayed capture:</td>
+    <td id="delayed-capture-id">On</td>
```
alexbishop1: I think this should just be `delayed-capture` (the other things with `-id` actually are IDs — `pyament-id`, `transaction_id` — but this isn’t)

test/cypress/integration/user/transaction_details_spec.js
```diff
@@ -84,6 +84,62 @@ describe('Transactions details page', () => {
...
+      cy.get('.transaction-details tbody').find('tr').eq(2).find('td').eq(1).should('have.text',
+        capitalise(chargeWithDelayedCapture.state.status))
+      // Amount
+      cy.get('.transaction-details tbody').find('tr').eq(3).find('td').eq(1).should('have.text',
+        convertAmounts(chargeWithDelayedCapture.amount))
+      // Refunded amount
+      cy.get('.transaction-details tbody').find('tr').eq(4).find('td').eq(1).should('have.text',
+        convertAmounts(chargeDetails.refund_summary.amount_submitted))
+      // Date created
+      cy.get('.transaction-details tbody').find('tr').eq(5).find('td').eq(1).should('contain',
```
jonheslop: All these `td` have id’s so you should be able to do which is maybe less brittle (and also could be faster)

`cy.get('#description').should`…

georgeracu: I agree with you on that, but it looks like the test is going into a particular order and this might be important in not breaking the ordering in the future.

PR #923, opened 2018-09-25
[DO NOT MERGE] PP-4234 A Platform admin enters

brid:
Introduces the concept of a 'platform admin' into self-service, which have their own permissions and ability to view information across the platform as a whole.

Whether or not a user is determined to be a platform admin from the perspective of self-service, is derived from the user object sent back via Adminusers when making an initial login request. A platform admin will have extra information sent back to self-service, outlining their permissions as a platform admin, and the services they are able to access.

Navigation and pages that are for use only by a platform admin are treated by a separate permissions flow, so this doesn't confuse the regular permissions system. As such a non platform-admin will not be able to access any platform admin specific pages, and to avoid the use of dozens/hundreds/thousands of 'if else' statements, reside on their own controllers and templates (for now) to keep things simple.


SandorArpa:
Old concept, no longer relevant

PR #926, opened 2018-10-10
Replace apostrophe → ' ← with right single quotation mark → ’ ←

alexbishop1:
Change:

“It's quick and easy to create a payment link and you don’t need any technical knowledge.”

… to:

“It’s quick and easy to create a payment link and you don’t need any technical knowledge.”

### DIFF
--- a/app/views/payment-links/index.njk
+++ b/app/views/payment-links/index.njk
@@ -19,7 +19,7 @@ Create a payment link - {{currentService.name}} {{currentGatewayAccount.full_typ
 <section class="column-two-thirds">
   <h1 class="page-title">Create a payment link</h1>
   <p>You can create a payment link so that users can make online payments to your service. Even if you don’t have a digital service, GOV.UK Pay can still take payments for you.</p>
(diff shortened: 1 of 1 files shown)


jonheslop:
👏 the ' is called a prime and the latter is an apostrophe ’

jonheslop:
you can get a proper ’ with `shift+alt+[`

alexbishop1:
I was using the names that Unicode assigns to them:

0027 ' APOSTROPHE — https://www.unicode.org/charts/PDF/U2000.pdf
2019 ’ RIGHT SINGLE QUOTATION MARK — https://www.unicode.org/charts/PDF/U0000.pdf

It’s `Shift`+`Alt`+`]` to get a ’ on a British English Mac keyboard (`Shift`+`Alt`+`[` is ”).

jonheslop:
Oops yeah i meant the other bracket.

And that’s outrageous that unicode call the first thing an apostrophe

alexbishop1:
I suspect the name was inherited from ASCII or something.

PR #943, opened 2018-10-30
PP-4336 transactions index migration to GOV.UK Frontend

jonheslop:
- Migrated the Transaction filters over to GOV.UK Frontend
- Updated tests to work with new templates
  > The biggest change is supplying the states and card brands ready for the
nunjucks rather than doing another for loop in the template. This means
that tests and controllers had to be updated to match the new
object model.
- Finished migrating transactions list page to GOV.UK Frontend 
  > Converted the table, pagination and page display size components
  > Also put in a couple of checks to capitalise JCB so it doesn’t come out
as JCB or jcb

### DIFF
--- a/app/controllers/transactions/transaction_list_controller.js
+++ b/app/controllers/transactions/transaction_list_controller.js
@@ -35,17 +35,22 @@ module.exports = (req, res) => {
           model.search_path = router.paths.transactions.index
           model.filtersDescription = describeFilters(filters.result)
           model.eventStates = states.allDisplayStateSelectorObjects()
-          model.eventStates.forEach(state => {
-            state.value.selected = filters.result.selectedStates && filters.result.selectedStates.includes(state.name)
+          .map(state => {
+            return {
+              value: state.key,
+              text: state.name,
+              selected: filters.result.selectedStates && filters.result.selectedStates.includes(state.name)
+            }
           })
+          model.eventStates.unshift({value: '', text: 'Any', selected: false})
 
           model.stateFiltersFriendly = model.eventStates
-            .filter(state => state.value.selected)
-            .map(state => state.value.text)
+            .filter(state => state.selected)
+            .map(state => state.text)
             .join(', ')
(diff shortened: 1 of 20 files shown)


app/controllers/transactions/transaction_list_controller.js
```diff
@@ -35,17 +35,22 @@ module.exports = (req, res) => {
           model.search_path = router.paths.transactions.index
           model.filtersDescription = describeFilters(filters.result)
           model.eventStates = states.allDisplayStateSelectorObjects()
-          model.eventStates.forEach(state => {
-            state.value.selected = filters.result.selectedStates && filters.result.selectedStates.includes(state.name)
+          .map(state => {
```
brid: 👍 mad props for this

app/utils/filters.js
```diff
@@ -47,6 +47,9 @@ function describeFilters (filters) {
 
   const brandStates = Array.isArray(filters.brand) ? filters.brand.map(brand => brand.replace('-', ' ')) : []
   if (brandStates.length === 0 && filters.brand) {
+    if (filters.brand === 'jcb') {
+      filters.brand = 'JCB'
+    }
```
brid: Maybe we could fix this so it's coming from the downstream service in the right format, and then we can go back and fix this so we don't need to do that.

app/utils/transaction_view.js
```diff
@@ -33,15 +33,16 @@ module.exports = {
...
     connectorData.cardBrands = lodash.uniqBy(allCards.card_types, 'brand')
-      .map((card) => {
-        let value = {}
-        value.text = card.label
-        if (card.brand === filtersResult.brand) {
-          value.selected = true
+      .map(card => {
+        return {
+          value: card.brand,
+          text: card.label === 'Jcb' ? card.label.toUpperCase() : card.label,
```
brid: Another strange one with the treatment of jcb/Jcb/JCB. Are we just trying to work around it being stored in the database in the wrong format (or something).

PR #949, opened 2018-11-05
BAU - update old style docs links so they work

jonheslop:
The links were just linking to the docs homepage, updated the links so they link directly to where they meant to


bahmady:
💯

PR #958, opened 2018-11-08
PP-4363 remove legacy css

jonheslop:
Me and Mark

- bf8fc3b - Remove old templates
- c6bd4f7 - remove 'new' suffix from base templates
- fcce913 - Remove logic around shared includes with old/new classes
  > We had some shared templates that the new and old templates inheirited so they needed to know which template they were part of. We no longer have the old templates so can remove all this logic.
- 3945e48 - Remove legacy frontend modules
- 9a4c130 - Remove old sass and js
- 9fa576a - Remove legacy css compilation stuff from Grunt pipeline


brid:
My kind of diff.

jonheslop:
Thanks Silvia, you’re the best. @brid you snooze you lose

PR #974, opened 2018-11-22
PP-4159 Add billing address settings pages

stephencdaly:
## WHAT
- Added controller and views for billing address pages
- Added 'Billing address' to settings nav bar
- Reworded the read-only message on settings pages to mention the
  billing address setting
- Added UI and controller tests for new pages

with @SandorArpa


app/services/service_service.js
```diff
@@ -127,3 +128,24 @@ function createService (serviceName, serviceNameCy, correlationId) {
...
+ *
+ * @param serviceExternalId
+ * @param collectBillingAddress
+ * @param correlationId
+ * @returns {*|Promise|Promise}
+ */
+function toggleCollectBillingAddress (serviceExternalId, collectBillingAddress, correlationId) {
+  return new Promise(function (resolve, reject) {
+    if (!serviceExternalId) return reject(new Error(`argument: 'serviceExternalId' cannot be undefined`))
+    getAdminUsersClient({correlationId}).updateCollectBillingAddress(serviceExternalId, collectBillingAddress)
```
brid: instead of wrapping a promise, then getting back the results of the promise and resolving/rejecting, you could just return your call to getAdminUsersClient directly, and handle it higher up.

As you aren't doing anything specific in your then() or catch() (custom logging or something specific) then you might as well let whatever called this, handle your promise for you.

stephencdaly: We were already handling errors in the controller too, so we've just removed the wrapping of the promise here now.

PR #978, opened 2018-11-30
PP-4472 Fix user access bug

mrlumbu:
We discovered that in some cases it is possible for a logged in user to
view and modify data from other services. This PR fixes the bug. 
The basic cause of the bug is that for urls that contain the service id 
we were not correctly checking whether the user had access to that url.

fd93aa2d has a failing test describing the bug and the main fix.

881e48dd ensures that we won't repeat this mistake in future by ensuring
that all controllers use a service id that we have ensured is valid for 
current user.

This probably needs unit tests around the resolve service middleware,
but I think the high level test is enough to show the bug is fixed. I will
do further tests in a separate PR, which will better demonstrate what
resolve_service is supposed to do.

### DIFF
--- a/app/middleware/resolve_service.js
+++ b/app/middleware/resolve_service.js
@@ -30,6 +30,9 @@ module.exports = function (req, res, next) {
   if (!req.service) {
     return notAuthorised(req, res)
   }
+
+  delete req.params.externalServiceId
+
   req.service.hasDirectDebitGatewayAccount = gatewayAccountType(req) === 'has_direct_debit_gateway_account'
   req.service.hasCardGatewayAccount = gatewayAccountType(req) === 'has_card_gateway_account'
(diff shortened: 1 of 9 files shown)


app/middleware/resolve_service.js
```diff
@@ -19,16 +28,22 @@ module.exports = function (req, res, next) {
   }
 
   if (!req.service) {
-    return renderErrorView(req, res, 'You do not have the rights to access this service.')
+    return notAuthorised(req, res)
   }
 
+  delete req.params.externalServiceId
```
georgeracu: I think this is `req.service.externalId`

mrlumbu: no, this is correct - I am deleting the incoming parameter on the request, as we should have now set `req.service` (or thrown an error). Everything after this should now use `req.service.externalId`

jonheslop:
Nice one Dom!

PR #1002, opened 2018-12-05
PP-4488 show corporate card surcharge on transaction list

SandorArpa:
## WHAT

- show total amount and indicate that it includes a surcharge when relevant
- add ui test
- add Cypress test for transaction list and skip modifying the old `transaction_list_ft_tests.js` test as it is covered by Cypress and it is asserting on JSON response which is not good way to test integration and ui

    with @jonheslop and @DanailMinchev

### DIFF
--- a/app/assets/sass/application.scss
+++ b/app/assets/sass/application.scss
@@ -32,6 +32,6 @@
 @import "components/multi-select";
 @import "components/pagination";
 @import "components/transaction-refund";
-@import "components/vetical-align";
+@import "components/vertical-align";
 @import "components/check-your-answers";
 @import "components/borders";
--- a/app/assets/sass/components/vertical-align.scss
+++ b/app/assets/sass/components/vertical-align.scss
@@ -1,3 +1,3 @@
-.vetical-align-top {
+.vertical-align-top {
   vertical-align: top;
 }
--- a/test/ui/transaction_list_ui_tests.js
+++ b/test/ui/transaction_list_ui_tests.js
@@ -12,7 +12,7 @@ const renderTemplate = require(path.join(__dirname, '/../test_helpers/html_asser
 
 describe('The transaction list view', function () {
   it('should render all transactions', function () {
-    var templateData = {
+    const templateData = {
       'results': [
         {
           'charge_id': 100,
@@ -103,7 +103,7 @@ describe('The transaction list view', function () {
       csvMaxLimitFormatted: '10,000'
     }
 
-    var body = renderTemplate('transactions/index', templateData)
+    const body = renderTemplate('transactions/index', templateData)
 
     templateData.results.forEach(function (transactionData, ix) {
       body.should.containSelector('table#transactions-list')
@@ -119,7 +119,7 @@ describe('The transaction list view', function () {
   })
 
   it('should render all transactions without download link', function () {
-    var templateData = {
+    const templateData = {
       'results': [
         {
           'charge_id': 100,
@@ -210,7 +210,7 @@ describe('The transaction list view', function () {
       csvMaxLimitFormatted: '10,000'
     }
 
-    var body = renderTemplate('transactions/index', templateData)
+    const body = renderTemplate('transactions/index', templateData)
 
     templateData.results.forEach(function (transactionData, ix) {
       body.should.containSelector('table#transactions-list')
@@ -226,7 +226,7 @@ describe('The transaction list view', function () {
   })
 
   it('should not render amount if no permission', function () {
-    var templateData = {
+    const templateData = {
       'results': [
         {
           'charge_id': 100,
@@ -285,14 +285,14 @@ describe('The transaction list view', function () {
       }
     }
 
-    var body = renderTemplate('transactions/index', templateData)
+    const body = renderTemplate('transactions/index', templateData)
 
     body.should.not.containSelector('#transactions-list .amount')
     body.should.not.containSelector('#amount-header')
   })
 
   it('should not render email if no permission', function () {
-    var templateData = {
+    const templateData = {
       'results': [
         {
           'charge_id': 100,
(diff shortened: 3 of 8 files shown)


app/assets/sass/application.scss
```diff
@@ -32,6 +32,6 @@
 @import "components/multi-select";
 @import "components/pagination";
 @import "components/transaction-refund";
-@import "components/vetical-align";
+@import "components/vertical-align";
```
DanailMinchev: Good spot 👍 , we are not using this CSS I guess before

app/assets/sass/components/vertical-align.scss
```diff
@@ -1,3 +1,3 @@
-.vetical-align-top {
+.vertical-align-top {
```
DanailMinchev: Now I can see that we were using it with the wrong name before

test/integration/transaction_list_ft_tests.js
```diff
@@ -594,6 +594,98 @@ describe('The /transactions endpoint filtering by states)', () => {
       })
       .end(done)
   })
+
+  it('should return a list of transactions with corporate surcharge total amount when relevant', function (done) {
```
DanailMinchev: It's great that you added tests for the change here 👍 

Currently we are expecting as a result JSON, which is not ideal and it is not fully tested (in terms of integration test as it skips the view), if you can see how we set the `supertest` - https://github.com/alphagov/pay-selfservice/blob/afcd4dcf8602049f6c591bf9d517265e823ffd65/test/integration/transaction_list_ft_tests.js#L609

As discussed before, we are trying to avoid using this approach and move to `Cypress` or `supertest` but not to `assert` on JSON.

We have Cypress test for the page already here:

- https://github.com/alphagov/pay-selfservice/blob/afcd4dcf8602049f6c591bf9d517265e823ffd65/test/cypress/integration/user/transaction_search_spec.js#L14

On the other option `supertest` + `cheerio` you can have a look as an example here https://github.com/alphagov/pay-selfservice/blob/afcd4dcf8602049f6c591bf9d517265e823ffd65/test/unit/controller/edit_merchant_details_controller/get-index.test.js#L65

I think `Cypress` is better to have as a test as it tests real interaction.

In case you think `Cypress` is good option you can change https://github.com/alphagov/pay-selfservice/blob/afcd4dcf8602049f6c591bf9d517265e823ffd65/test/cypress/integration/user/transaction_search_spec.js file and add for example following:

```
  const selfServiceDefaultUser = selfServiceUsers.config.users.filter(fil => fil.isPrimary === 'true')[0]

  const aCorporateCardSurchargeCharge = selfServiceDefaultUser.sections.transactions.data.filter(fil => fil.corporate_card_surcharge !== undefined)[0]
```

to https://github.com/alphagov/pay-selfservice/blob/afcd4dcf8602049f6c591bf9d517265e823ffd65/test/cypress/integration/user/transaction_search_spec.js#L6

and later add a test, initial implementation can be like:

```JS
    it('should display card fee with corporate surcharge transaction', () => {
      // Ensure the transactions list has the right number of items
      const unfilteredTransactions = selfServiceDefaultUser.sections.transactions.data
      cy.get('#transactions-list tbody').find('tr').should('have.length', unfilteredTransactions.length)

      // Ensure the values are displayed correctly
      cy.get('#transactions-list tbody').find('tr').eq(4).find('td').eq(1).should('contain', convertAmounts(aCorporateCardSurchargeCharge.total_amount)).and('contain', '(with card fee)')
    })
```

test/ui/transaction_list_ui_tests.js
```diff
@@ -422,4 +422,88 @@ describe('The transaction list view', function () {
     body.should.not.containSelector('#transactions-list .brand')
     body.should.not.containSelector('#brand-header')
   })
+
+  it('should render all transactions with corporate surcharge', function () {
```
DanailMinchev: If you decide to add Cypress, I think we can keep this test as well, more like "unit" level test, nice that you added it 👍

test/integration/transaction_list_ft_tests.js
```diff
@@ -594,6 +594,98 @@ describe('The /transactions endpoint filtering by states)', () => {
...
+    connectorMock.get(CONNECTOR_ALL_CARD_TYPES_API_PATH)
+      .reply(200, ALL_CARD_TYPES)
+
+    let connectorData = {
```
DanailMinchev: can be `const`

test/integration/transaction_list_ft_tests.js
```diff
@@ -594,6 +594,98 @@ describe('The /transactions endpoint filtering by states)', () => {
...
+          'corporate_card_surcharge': 100,
+          'total_amount': 2100
+        }
+      ],
+      total: 2
+    }
+
+    connectorMockResponds(200, connectorData, connectorSearchParameters)
+
+    let expectedData = {
```
DanailMinchev: same here, can be `const`

test/ui/transaction_list_ui_tests.js
```diff
@@ -422,4 +422,88 @@ describe('The transaction list view', function () {
...
+    let templateData = {
```
DanailMinchev: can be `const`

test/ui/transaction_list_ui_tests.js
```diff
@@ -422,4 +422,88 @@ describe('The transaction list view', function () {
...
+        'transactions_download_read': true
+      },
+      hasResults: true,
+      total: 1,
+      showCsvDownload: true,
+      totalFormatted: '1',
+      csvMaxLimitFormatted: '10,000'
+    }
+
+    let body = renderTemplate('transactions/index', templateData)
```
DanailMinchev: can be `const`

test/integration/transaction_list_ft_tests.js
DanailMinchev: As discussed I have paired with @SandorArpa to add the Cypress test

PR #1007, opened 2018-12-06
PP-4472 - add more tests around resolving service

jonheslop:
- 6e30374 & 42c7028 - First did a bit of tidying up as the middleware/tests weren’t up to scratch with our Node.JS best practices
- c082763 - I found this function confusing so simplified it a bit, could have gone further but didnt (see commit message)
- 6b5dfae - The meat of the ticket this is what I was actually here to do

### DIFF
--- a/app/middleware/resolve_service.js
+++ b/app/middleware/resolve_service.js
@@ -16,25 +16,16 @@ const getServiceFromUserByExternalId = (req, externalServiceId) => {
 }
 
 const gatewayAccountType = req => {
-  let hasDDAccount = false
-  let hasCardAccount = false
-
   if (req.service.gatewayAccountIds) {
-    req.service.gatewayAccountIds.forEach(account => {
-      if (isADirectDebitAccount(account)) {
-        hasDDAccount = true
-      } else {
-        hasCardAccount = true
-      }
-    })
-  }
+    const DDorNot = req.service.gatewayAccountIds.map(account => isADirectDebitAccount(account))
 
-  if (hasDDAccount && hasCardAccount) {
-    return 'has_card_and_dd_gateway_account'
-  } else if (hasDDAccount) {
-    return 'has_direct_debit_gateway_account'
-  } else if (hasCardAccount) {
-    return 'has_card_gateway_account'
+    if (DDorNot.every(account => account === true)) {
+      return 'hasDirectDebitGatewayAccount'
+    } else if (DDorNot.some(account => account === true)) {
+      return 'hasCardAndDirectDebitGatewayAccount'
+    } else {
+      return 'hasCardGatewayAccount'
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


test/unit/middleware/resolve_service_test.js
```diff
@@ -8,24 +8,88 @@ const {expect} = require('chai')
...
+    expect(res.render.calledWith('error'))
+    expect(nextSpy.called).to.equal(false)
+  })
+
+  it('from gatewayAccountId in cookie', () => {
+    const req = {
+      user: user,
+      params: {},
+      gateway_account: {
+        currentGatewayAccountId: user.serviceRoles[0].service.externalServiceId
```
mrlumbu: don't think this set up is quite right? why does a gateway account id = a service id?

test/integration/service_users_ft_test.js
```diff
@@ -1,15 +1,24 @@
...
 const chaiAsPromised = require('chai-as-promised')
+
+const {expect} = chai
+
+// Local modules
+const session = require(path.join(__dirname, '/../test_helpers/mock_session.js'))
+const {getApp} = require(path.join(__dirname, '/../../server.js'))
+const serviceFixtures = require(path.join(__dirname, '/../fixtures/service_fixtures'))
+const userFixtures = require(path.join(__dirname, '/../fixtures/user_fixtures'))
+const paths = require(path.join(__dirname, '/../../app/paths.js'))
```
mrlumbu: is our standard to use the `path.join(__dirname...` stylee or not? Seems pointless, and is easier to navigate around without

app/middleware/resolve_service.js
```diff
@@ -1,17 +1,46 @@
...
+
+  if (req.service.gatewayAccountIds) {
+    req.service.gatewayAccountIds.forEach(account => {
+      if (isADirectDebitAccount(account)) {
+        hasDDAccount = true
+      } else {
+        hasCardAccount = true
+      }
+    })
+  }
```
mrlumbu: How about 
```
hasDDAccount = req.service.gatewayAccountIds.find(gatewayAccountId => isADirectDebitAccount(gatewayAccountId)) !== undefined
hasCardAccount = req.service.gatewayAccountIds.find(gatewayAccountId => !isADirectDebitAccount(gatewayAccountId)) !== undefined

jonheslop: Did you see a refactored this in the third commit — c082763?

test/integration/service_users_ft_test.js
jonheslop: Oh yeah nice spot dont need any of that

app/middleware/resolve_service.js
mrlumbu: oh sorry, yeah. 

`DDorNot, here I come, you can't hide`

jonheslop: I still think that could be improved later, `hasDD` is not actually right, because it’s more like `isOnlyDD`. I did start refactoring that but it would have meant changes to loads of controllers, views and tests and I thought better off doing that seperately.

PR #1012, opened 2018-12-07
PP-4489 Add corporate card surcharge and total amount columns to CSV

DanailMinchev:
## WHAT

**[REVIEW ONLY] [DO NOT MERGE]**

This PR adds two new columns into transaction's CSV:
- `Corporate Card Surcharge`
- `Total Amount`

`Total Amount` column should display `total_amount` property when we have corporate card surcharge or `amount` when we don't have corporate card surcharge.

This PR has some readability improvements and fixes the tests.

## HOW

- Modified the code not to override `refund_summary.user_external_id` with `username`.
- Added explicit column `refund_summary.user_username` to be used for displaying `username`.
  Ideally `refund_summary.user_username` should be returned from `pay-connector` - as discussed this can be done later depending on the available time.
- Fixed the `json_to_csv_tests` to test correctly for refunds.
- Modified `json_to_csv.js` utility to include `Corporate Card Surcharge` and `Total Amount` columns when generating CSV.
- In case when there is no corporate card surcharge `Total Amount` column should display the original `Amount` value.
- Modified tests.
- Applied code styling.

with @cobainc0


app/utils/json_to_csv.js
```diff
@@ -81,6 +81,20 @@ module.exports = function (data) {
...
+          value: row => {
+            const amountInPence = row.corporate_card_surcharge ? row.corporate_card_surcharge : 0
+            return (parseInt(amountInPence) / 100).toFixed(2)
+          }
+        },
+        {
+          label: 'Total Amount',
+          value: row => {
+            const amountInPence = row.total_amount ? row.total_amount : row.amount
+            return (row.transaction_type === 'refund') ? (parseInt(amountInPence) * -1 / 100).toFixed(2) : (parseInt(amountInPence) / 100).toFixed(2)
```
jonheslop: We have helper functions for this in https://github.com/alphagov/pay-selfservice/blob/master/app/utils/currency_formatter.js for example https://github.com/alphagov/pay-selfservice/blob/e869a5511523b04ad476e482973d136ae657430e/app/models/TransactionEvent.class.js#L26

jonheslop: Also just noticed, perhaps I wrong, but it says `amountInPence` but you dividing the pence amount by 100 so looks like it’s being transformed to pounds and pence, so the name is misleading

PR #1041, opened 2018-12-20
PP-4564: Address priority repo lint issues

sfount:
A number of lint warning were recently flagged against the `selfservice`
repository. This PR removes these issues, specifically high priority
concerns with checking equality (running boolean logic checks against
JavaScript strings always resulting in `true`)

Closes https://github.com/alphagov/pay-selfservice/pull/1040

### DIFF
--- a/app/assets/js/components/datetime-picker.js
+++ b/app/assets/js/components/datetime-picker.js
@@ -1,5 +1,5 @@
 (function () {
-  datetimePicker = function () {
+  var datetimePicker = function () {
     var init = function () {
         datePicker();
         timePicker();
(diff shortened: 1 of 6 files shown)


sjvs:
Ah, you fixed many more alerts than just my one in #1040! In #1040, I also flagged up that some of the affected code seems to exist in at least one other repository: https://lgtm.com/projects/g/alphagov/pay-products-ui.

And — just in case — if you're looking for more inspiration: LGTM.com has analysed over 500 repos in the `alphagov` GitHub org. Have fun! https://lgtm.com/search?q=alphagov

PR #1053, opened 2018-12-24
PP-4449 Request To Go Live static page

DanailMinchev:
## WHAT

This PR adds `Request to go live` static skeleton page which includes initial setup and design, but no logic.
Next PRs will implement the functionality.

## HOW 

see commits for more information

with @cobainc0


app/controllers/request-to-go-live/index/get.controller.js
```diff
@@ -0,0 +1,9 @@
+'use strict'
+
+// Local dependencies
+const response = require('../../../utils/response')
+
+module.exports = (req, res) => {
+  const externalServiceId = req.service.externalId
+  return response.response(req, res, 'request-to-go-live/index', {externalServiceId})
```
sfount: Just a quick note that it's generally considered bad practice to pass the `req` and `res` Express objects to other methods that aren't directly in the middleware stack. It makes it hard to reason about the request flow and can lead to accidentally terminating a response early or twice.

This is nothing to do with this PR or body of work I'm just noting to reference in future - I'll follow this up with a refactor ticket.

sfount: Also while we're at it `response.response` is hard to read.

app/controllers/request-to-go-live/index/post.controller.js
```diff
@@ -0,0 +1,9 @@
+'use strict'
+
+// Local dependencies
+const paths = require('../../../paths')
+
+module.exports = (req, res) => {
+  const externalServiceId = req.service.externalId
+  res.redirect(303, paths.requestToGoLive.index.replace(':externalServiceId', externalServiceId))
```
sfount: Is this route/ redirect just a placeholder for upcoming work? Not 100% on repo standards but a quick `todo` comment could make this clear and act as something to remove when the work is done.

DanailMinchev: Yes, good idea, I have added `TODO` comments to mark what is next, thank you

app/controllers/request-to-go-live/index/get.controller.js
DanailMinchev: Yes, as discussed with @sfount  we need to do this part of broader refactoring, for example not to pass `req` and `res` to "external" functionality and we should think of better naming as part of this refactoring.
This will be done after some research and reference implementation of `pay-toolbox`

PR #1072, opened 2019-01-04
PP-4449 "Request to go live" index page implementation and testing

DanailMinchev:
## WHAT

This PR implements "Request to go live" index page and adds Cypress testing.
Routes were added for future workflow pages.
The content/design of the page can change in near future, but currently this implementation is the latest one.

## HOW 

- Add all "request-to-go-live" url paths
- Add "go-live-stage-to-next-page-path" utility
 Display `COMPLETED` label for each step depending on `GoLiveStage`
  returned from adminusers
- Redirect to the next step depending on `GoLiveStage` returned from
  adminusers
- Add "Your service is live" panel which will be removed later when the
  new mockup is ready
- Modified the template to have test CSS ids
- Add Cypress tests, including the static fixtures and
  modified related Pact tests

with @stephencdaly

### DIFF
--- a/app/controllers/request-to-go-live/go-live-stage-to-next-page-path.js
+++ b/app/controllers/request-to-go-live/go-live-stage-to-next-page-path.js
@@ -2,20 +2,20 @@
 
 // Local dependencies
 const goLiveStage = require('../../models/go-live-stage')
-const paths = require('../../paths')
+const { requestToGoLive } = require('../../paths')
 
 const goLiveStageToNextPagePathMap = {}
-goLiveStageToNextPagePathMap[goLiveStage.NOT_STARTED] = paths.requestToGoLive.organisationName
-goLiveStageToNextPagePathMap[goLiveStage.ENTERED_ORGANISATION_NAME] = paths.requestToGoLive.chooseHowToProcessPayments
-goLiveStageToNextPagePathMap[goLiveStage.CHOSEN_PSP_STRIPE] = paths.requestToGoLive.agreement
-goLiveStageToNextPagePathMap[goLiveStage.CHOSEN_PSP_WORLDPAY] = paths.requestToGoLive.agreement
-goLiveStageToNextPagePathMap[goLiveStage.CHOSEN_PSP_SMARTPAY] = paths.requestToGoLive.agreement
-goLiveStageToNextPagePathMap[goLiveStage.CHOSEN_PSP_EPDQ] = paths.requestToGoLive.agreement
-goLiveStageToNextPagePathMap[goLiveStage.TERMS_AGREED_STRIPE] = paths.requestToGoLive.index
-goLiveStageToNextPagePathMap[goLiveStage.TERMS_AGREED_WORLDPAY] = paths.requestToGoLive.index
-goLiveStageToNextPagePathMap[goLiveStage.TERMS_AGREED_SMARTPAY] = paths.requestToGoLive.index
-goLiveStageToNextPagePathMap[goLiveStage.TERMS_AGREED_EPDQ] = paths.requestToGoLive.index
-goLiveStageToNextPagePathMap[goLiveStage.DENIED] = paths.requestToGoLive.index
-goLiveStageToNextPagePathMap[goLiveStage.LIVE] = paths.requestToGoLive.index
+goLiveStageToNextPagePathMap[goLiveStage.NOT_STARTED] = requestToGoLive.organisationName
+goLiveStageToNextPagePathMap[goLiveStage.ENTERED_ORGANISATION_NAME] = requestToGoLive.chooseHowToProcessPayments
+goLiveStageToNextPagePathMap[goLiveStage.CHOSEN_PSP_STRIPE] = requestToGoLive.agreement
+goLiveStageToNextPagePathMap[goLiveStage.CHOSEN_PSP_WORLDPAY] = requestToGoLive.agreement
+goLiveStageToNextPagePathMap[goLiveStage.CHOSEN_PSP_SMARTPAY] = requestToGoLive.agreement
+goLiveStageToNextPagePathMap[goLiveStage.CHOSEN_PSP_EPDQ] = requestToGoLive.agreement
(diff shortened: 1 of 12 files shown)


app/controllers/request-to-go-live/go-live-stage-to-next-page-path.js
```diff
@@ -0,0 +1,21 @@
+'use strict'
+
+// Local dependencies
+const goLiveStage = require('../../models/go-live-stage')
+const paths = require('../../paths')
```
danworth: could make this ```paths = require('../../paths').requestToGoLive``` and remove the `requestToGoLive` each time further down.

app/controllers/request-to-go-live/index/post.controller.js
```diff
@@ -1,10 +1,11 @@
...
-const paths = require('../../../paths')
+const goLiveStageToNextPagePath = require('../go-live-stage-to-next-page-path')
 
 module.exports = (req, res) => {
-  // TODO: finish the UI implementation and logic
-  const externalServiceId = req.service.externalId
-  res.redirect(303, paths.requestToGoLive.index.replace(':externalServiceId', externalServiceId))
+  res.redirect(
+    303,
+    goLiveStageToNextPagePath[req.service.currentGoLiveStage].replace(':externalServiceId', req.service.externalId)
```
danworth: This seems to follow the existing so this is merely a suggestions. To avoid the `replace` each time, `paths.requestToGoLive` could be changed to return a function which takes the `externalId` and returns the correct String using a template. eg.
```
requestToGoLive: {
    index: (externalId) => `/service/${externalId}/request-to-go-live`,
    organisationName: (externalId) => `/service/${externalId}/request-to-go-live/organisation-name`,
    chooseHowToProcessPayments: (externalId) => `/service/${externalId}/request-to-go-live/choose-how-to-process-payments`,
    agreement: (externalId) => `/service/${externalId}/request-to-go-live/agreement`
  }
```

and then in the POST controller...
```
goLiveStageToNextPagePath[req.service.currentGoLiveStage](req.service.externalId)
```

app/paths.js
```diff
@@ -156,6 +157,9 @@ module.exports = {
   },
   generateRoute: require(path.join(__dirname, '/utils/generate_route.js')),
   requestToGoLive: {
-    index: '/service/:externalServiceId/request-to-go-live'
+    index: '/service/:externalServiceId/request-to-go-live',
```
danworth: Suggestion above about making these return a function which takes the `externalId` which returns the correct String, rather than having to use `replace` later on.

app/controllers/request-to-go-live/index/get.controller.js
```diff
@@ -1,10 +1,16 @@
 'use strict'
```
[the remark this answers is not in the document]
DanailMinchev: I think in selfservice we try not to use `_` as file and directory naming and we replaced the naming with `-`.
But, also as you noticed we have `something.type.js` naming (only for classes currently) and I was thinking to start introducing https://github.com/alphagov/pay-direct-debit-frontend style where we have `something.type.js` for everything, this way we can implement the whole "Request to go live" feature with the new file naming.
We can discuss this as wider team, I can rollback to `-` naming, but I was thinking to use the latest patterns introduced in Direct Debit frontend which we discussed already when we were doing Direct Debit frontend
@danworth  let me know what do you think, thank you

app/controllers/request-to-go-live/index/post.controller.js
DanailMinchev: Good idea! I started changing it and noticed that we use https://github.com/alphagov/pay-selfservice/blob/d221e6a8fe9b229c048849c0f6b37946ae753588/app/routes.js#L187 which relies on `String`s
As discussed one way to fix that will be to call all of the functions in `requestToGoLive` and have them to return the placeholder again, as discussed again with @danworth it is actually doing things more complicated because of `authenticatedPaths` so we are going to skip changing this

app/paths.js
DanailMinchev: as discussed we are not changing this, please see the comment above, thanks

app/controllers/request-to-go-live/go-live-stage-to-next-page-path.js
DanailMinchev: Thanks, I changed that to use `Destructuring assignment`

test/cypress/plugins/index.js
```diff
@@ -18,30 +18,210 @@ module.exports = (on, config) => {
 
   // The same fixed user config is used to generate pacts/contracts, so generating cookies from this config
   // ensures our pact-stub server is ready to return canned responses.
-  const ssUserConfig = require('../../fixtures/config/self_service_user.json')
+  const selfServiceUserConfig = require('../../fixtures/config/self_service_user.json')
 
-  const ssUser = ssUserConfig.config.users.filter(fil => fil.isPrimary === 'true')[0]
+  // Default user
+  const selfServiceUser = selfServiceUserConfig.config.users.filter(fil => fil.isPrimary === 'true')[0]
```
danworth: I think you could you use `.find()` rather than `.filter()[0]`. `find()` will return the first matching record otherwise `undefined`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

DanailMinchev: Good point! I addressed it in https://github.com/alphagov/pay-selfservice/pull/1072/commits/424ccd9582f7785eb396e0b1fe02e39149011742

We are using `.filter()[0]` a lot in our code in other places, so probably we should refactor it at some point as well

test/fixtures/config/self_service_user.json
```diff
@@ -4,6 +4,7 @@
     "users": [
```
danworth: Sounds like we're all agreed that this is not the ideal way to set-up this test data given the amount of duplication and difficulty to review for accuracy. Since you and Stephen are looking for a more efficient and easier to understand way to do this, lets move forward with this to unblock progress and make sure it is revisited later on. It seems using something like a builder with defaults and dynamically generating the `.json` (if required) might be good.

PR #1075, opened 2019-01-08
PP-4590 fix self service filter csv download

sfount:
Card holder name filter was omitted from the CSV download transactions
link, this meant that clicking to download CSV would download all transactions. 

This ensures that it is templated into the download link to prevent this behaviour.

Note: this has been tested locally but the behaviour should be verified live before signing off on the ticket.

### DIFF
--- a/test/integration/transaction_list_ft_tests.js
+++ b/test/integration/transaction_list_ft_tests.js
@@ -483,7 +483,7 @@ describe('The /transactions endpoint', function () {
 })
 
 
-describe.only('The /transactions endpoint filtering', () => {
+describe('The /transactions endpoint filtering', () => {
   afterEach(function () {
     nock.cleanAll()
     app = null
--- a/test/integration/transaction_list_ft_tests.js
+++ b/test/integration/transaction_list_ft_tests.js
@@ -482,7 +482,8 @@ describe('The /transactions endpoint', function () {
   })
 })
 
-describe('The /transactions endpoint filtering by states)', () => {
+
+describe.only('The /transactions endpoint filtering', () => {
   afterEach(function () {
     nock.cleanAll()
     app = null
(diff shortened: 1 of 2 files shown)


test/integration/transaction_list_ft_tests.js
```diff
@@ -482,7 +482,8 @@ describe('The /transactions endpoint', function () {
   })
 })
 
-describe('The /transactions endpoint filtering by states)', () => {
+
+describe.only('The /transactions endpoint filtering', () => {
```
DanailMinchev: Probably you forgot to remove `.only()`

test/integration/transaction_list_ft_tests.js
```diff
@@ -598,6 +599,36 @@ describe('The /transactions endpoint filtering by states)', () => {
...
+    const connectorData = {
+      'results': []
+    }
+
+    connectorMockResponds(200, connectorData, expectedCSVFilters)
+
+    request(app)
+      .get(paths.transactions.index)
+      .query(expectedCSVFilters)
+      .set('Accept', 'application/json')
```
DanailMinchev: We are trying to avoid testing the "app/integration" tests and setting the `application/json` as it returns JSON and not actual page, but looks like all of the download CSV tests are implemented this way.
We have some Cypress tests for transaction list here https://github.com/alphagov/pay-selfservice/blob/5652dff08d64139f15e221051ee643b8ff844795/test/cypress/integration/user/transaction_search_spec.js
but it is not used for CSV download yet.
Given the urgency of the fix, I think we can refactor / add Cypress testing at some point later

test/integration/transaction_list_ft_tests.js
sfount: Good catch! 😮

test/integration/transaction_list_ft_tests.js
sfount: I agree with this - we should definitely add these tests following the potential upcoming updates to Cypress testing.

PR #1078, opened 2019-01-09
PP-4604 nunjucks filters from pay js commons

jonheslop:
I’ve migrated all the nunjucks filters from selfservice to https://github.com/alphagov/pay-js-commons. There are a few utils that the filters depend on, so I have migrated them too.

This will allow us to use the filters in other repos without duplication of code.


jonheslop:
Things I would question about what I’ve done, 

1) _One day_ this repo could get massive would it be best to start now and split it so like `pay-nunjucks` `pay-js-utils` `pay-js-clients` etc etc

2) Is -
https://github.com/alphagov/pay-selfservice/blob/480b524332f3e0fa0f2bff04e9a3ea48a242bba0/app/browsered/input-confirm.js#L3 

Rapidly going to become difficult to discover where to look? Is there a better way to do this, perhaps 1) is the answer to this

server.js
```diff
@@ -102,6 +102,7 @@ function initialiseTemplateEngine (app) {
   // Load custom Nunjucks filters
   for (let name in nunjucksFilters) {
     let filter = nunjucksFilters[name]
+    console.log(name, filter)
```
simad: `console.log('cough')`

jonheslop: 🙈

PR #1083, opened 2019-01-11
[DO NOT MERGE] Modify cypress tests to use Mountebank for stubbing

stephencdaly:
Note: this pull request isn't complete, and is still open to changing how we do it before and if we use it.

## AIMS
The idea is to split up cypress tests from the pacts so that hopefully:
  - It is easier to add new tests
  - Tests are less brittle
  - Fewer lines of code/json need to be be added for test configuration
  - It is easier to follow how the tests work
  - The pact tests are better as their purpose is not confused

## CURRENT STATE
How cypress tests currently work, is that running `npm test` runs the pact tests which generate pact json files. Then while the cypress tests are running, we run `pact-stub-server`, which loads in these pact files. When the cypress tests cause requests to be fired at the various microservices, they are instead caught by the pact stub server, and if they match a pact there is a request for, the response defined in the pact is returned.

### Problems with this
- Setting up new test configurations are difficult, as we have to add a pact test per configuration. In the current state, the vast majority of these pact tests don't have provider states setup so are not run.
- The pact tests we are adding for the configurations do not follow the guidelines of what a pact test should be testing, as they are too specific and are often testing the functionality of the endpoint rather than the contract. The rules in the pact tests are also too lax in some cases to don't provide any proper validation.
- We currently rely on a static json file self_service_user.json, which is confusing and difficult to add new test cases to without it getting massive

## What the new solution is
- Use Mountebank as a stub server. On this we set up rules to match incoming requests and specify the responses that will be returned.
- Do the stub setup on a per-test basis rather than globally. This allows for greater flexibility and ease for setting up different scenarios. It also makes it easy to follow through which stubs each test uses, and makes it possible to modify test data for one tests without affecting other tests.
- Rely on the fixture builders (e.g. user_fixtures.json) to be validated in order to provide the validation we were previously getting from pact. Make the fixture builders less flexible by not substituting in entire objects passed in the parameters. Then, the fixture builders that are used for building the stubs for cypress testing can be validated by using the same fixtures for creating pacts. I think this should be a sufficient level of validation, as proper validation should be done by adequately unit testing the clients using pact.


package.json
```diff
@@ -77,6 +77,7 @@
     "minimist": "1.2.x",
     "moment-timezone": "^0.5.23",
     "morgan": "^1.9.1",
+    "mountebank": "^1.16.0",
```
stephencdaly: This runs as a server and is started using the command `mb` in the selfservice directory

test/cypress/integration/user/dashboard_spec.js
```diff
@@ -1,25 +1,42 @@
...
   beforeEach(() => {
-    cy.setCookie('session', Cypress.env('encryptedSessionCookie'))
-    cy.setCookie('gateway_account', Cypress.env('encryptedGatewayAccountCookie'))
+    cy.task('getCookies', {
+      user_external_id: userExternalId,
+      gateway_account_id: gatewayAccountId
+    }).then(cookies => {
+      cy.setCookie('session', cookies.encryptedSessionCookie)
+      cy.setCookie('gateway_account', cookies.encryptedGatewayAccountCookie)
+    })
```
stephencdaly: Setup the cookies per test, before we did it globally by setting environment variables in plugins/index.js, but this assumed we were only ever using the same user/gateway account which wasn't the case for a recent test added so this got a bit messy.

test/cypress/integration/user/dashboard_spec.js
```diff
@@ -1,25 +1,42 @@
 describe('Dashboard', () => {
-  const selfServiceUsers = require('../../../fixtures/config/self_service_user.json')
+  const userExternalId = '7d19aff33f8948deb97ed16b2912dcd3'
+  const gatewayAccountId = '666'
+  const serviceName = 'Test Service'
 
   beforeEach(() => {
-    cy.setCookie('session', Cypress.env('encryptedSessionCookie'))
-    cy.setCookie('gateway_account', Cypress.env('encryptedGatewayAccountCookie'))
+    cy.task('getCookies', {
```
stephencdaly: Cypress tasks are a way of calling node code during a test, and are defined in plugins/index.js

test/cypress/integration/user/dashboard_spec.js
```diff
@@ -1,25 +1,42 @@
...
+        opts: {
+          gateway_account_ids: [gatewayAccountId.toString()],
+          service_name: serviceName
+        }
+      },
+      {
+        name: 'getGatewayAccountSuccess',
+        opts: { gateway_account_id: gatewayAccountId }
+      }
+    ])
```
stephencdaly: This sets up the stubs that are loaded into Mountebank and used throughout the test. The stubs are cleared after each test.

The stubs are defined in cypress/plugins/stubs.js

test/cypress/plugins/index.js
```diff
@@ -14,193 +22,41 @@ const cookieMonster = require('../integration/utils/cookie-monster')
...
+          stubs: stubsArray
+        }
+      })
+    },
+    clearStubs () {
+      return baseClient.delete({
+        url: 'http://localhost:2525/imposters'
+      })
+    }
+  })
```
stephencdaly: These tasks are invoked from the cypress tests themselves

test/cypress/plugins/index.js
```diff
@@ -14,193 +22,41 @@ const cookieMonster = require('../integration/utils/cookie-monster')
...
+
+      return baseClient.post({
+        url: 'http://localhost:2525/imposters',
+        json: true,
+        body: {
+          port: 8000,
+          protocol: 'http',
+          stubs: stubsArray
+        }
+      })
```
stephencdaly: we post to Mountebank to setup the stubs. I'll clean this up a bit to put the address in a config file at some point

test/cypress/plugins/stubs.js
```diff
@@ -0,0 +1,261 @@
+'use strict'
+
+// Local dependencies
+const userFixtures = require('../../fixtures/user_fixtures')
+const gatewayAccountFixtures = require('../../fixtures/gateway_account_fixtures')
+
+module.exports = {
+  getUserSuccess: (opts = {}) => {
+    const aValidUserResponse = userFixtures.validPasswordAuthenticateResponse(opts).getPlain()
```
stephencdaly: we use the fixtures to setup the stubs.
the aim will be that these fixtures are satisfactorily validated by also being used to generate stubs, which will have different data, but the same structure

jonheslop:
This sounds like a excellent plan 💫

PR #1087, opened 2019-01-14
PP-4449 New design changes

DanailMinchev:
## WHAT

- Modified the "Request to go live" page content and updated tests

with @cobainc0

### DIFF
--- a/app/views/request-to-go-live/index.njk
+++ b/app/views/request-to-go-live/index.njk
@@ -8,17 +8,10 @@
 
 {% block mainContent %}
   {% if currentService.currentGoLiveStage === goLiveStage.LIVE %}
-    <div class="govuk-grid-column-full">
+    <div class="govuk-grid-column-full govuk-!-margin-top-6">
       <div class="next-steps-panel">
-        <h3 class="govuk-heading-m govuk-!-margin-bottom-0">Your service is live</h3>
-        <p class="govuk-body-l govuk-!-margin-bottom-4">You can now start taking payments.</p>
-        <h3 class="govuk-heading-s govuk-!-margin-bottom-2">Next steps:</h3>
-        <ul class="govuk-list govuk-list--bullet">
-          {% if currentService.hasCardGatewayAccount %}
-            <li><a class="govuk-link govuk-link--no-visited-state" href="/card-types/summary">manage the card types</a> your service accepts and turn on 3D Secure</li>
-          {% endif %}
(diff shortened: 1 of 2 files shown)


app/views/request-to-go-live/index.njk
```diff
@@ -8,17 +8,10 @@
 
 {% block mainContent %}
   {% if currentService.currentGoLiveStage === goLiveStage.LIVE %}
-    <div class="govuk-grid-column-full">
+    <div class="govuk-grid-column-full govuk-!-margin-top-6">
       <div class="next-steps-panel">
```
jonheslop: What is this div for?

app/views/request-to-go-live/index.njk
```diff
@@ -91,7 +85,7 @@
               <span class="status">COMPLETED</span>
             {% endif %}
           </h3>
-          <p class="govuk-body-m">GOV.UK Pay will contact you within one working day to confirm whether or not your organisation has already signed an agreement. If not, we’ll send you the relevant documents.</p>
+          <p class="govuk-body-m">To start using GOV.UK Pay your organisation must accept GOV.UK Pay’s legal terms.</p>
```
jonheslop: As far I can see `govuk-body-m` isn’t a thing here — https://design-system.service.gov.uk/styles/typography/ and in the code in looks the same as `govuk-body` so maybe best use that instead.

app/views/request-to-go-live/index.njk
```diff
@@ -57,7 +51,7 @@
               <span class="status">COMPLETED</span>
             {% endif %}
           </h3>
-          <p class="govuk-body-m">This information will appear on your payment pages.</p>
+          <p class="govuk-body-m">We’ll use this to perform a security check.</p>
```
jonheslop: Same again https://github.com/alphagov/pay-selfservice/pull/1087/files#r247822150

app/views/request-to-go-live/index.njk
jonheslop: As in the one with `.next-steps-panel`

jonheslop: it’s not used in tests AFAIK and if it was in could be added to the parent div rather than adding a whole new one

DanailMinchev: Thanks, we tried the suggestion, looks like when we apply the changes the `next-steps-panel` is bigger than expected, also https://design-system.service.gov.uk/components/error-summary/ generates internal div as well, we can show you visually the changes and let's discuss when you are available

PR #1089, opened 2019-01-14
PP-4613 serve policy docs behind authentication

sfount:
**From the ticket:**
> Services need a place to access our contract, MOU and GOV.UK Pay card services contract terms. This should only be accessible to logged in users.

* Adds nunjucks template structure for policy document download intermediate page 
* Adds controller and library for generating signed URLs into an amazon S3 bucket 

**Steps required for integration:**
- [x] Test AWS credentials being populated by environment variables in `test` AWS environment
* Generate STS temporary credentials instead of directly inheriting instance credentials (for reliability/ transparency)
- [x] Replace `(PLACEHOLDER)` text with finalised document language
- [x] Ensure documents are available from the contract team in the S3 bucket
- [x] Dynamically render the footer based on authentication level
- [x] Production logs - ensure sumo logic can see user ID: private link
* Test glue between controllers and libraries

Integration discussion: 
* How can we validate live document access is available? Should this be included in smoke tests 
* The configuration and environment variables are only set for the production environment - this is also where the official documents are stored. In order to make this work with test + staging we will have to consider configuring these envs


app/controllers/policy/policyDownload.controller.js
```diff
@@ -0,0 +1,21 @@
+'use strict'
+
+const { response, renderErrorView } = require('../../utils/response')
+
+const supportedPolicyDocuments = require('./supportedPolicyDocuments')
+const policyBucket = require('./awsS3PolicyBucket')
+
+const downloadDocumentsPolicyPage = async function downloadDocumentsPolicyPage (req, res, next) {
```
jonheslop: Async :D

app/views/policy/policy.njk
```diff
@@ -0,0 +1,26 @@
...
+</div>
+
+<div class="govuk-grid-column-two-thirds">
+  <p class="govuk-body">
+    {% block policyDescription %} {% endblock %}
+  </p>
+  <ul class="govuk-list govuk-list--bullet">
+    <li>
+      <!-- @FIXME(sfount) check the name of the downloaded PDF document - is it according the PDF headers? If not set the name through the download attribute -->
+      <a href="{{ link }}" download>{% block policyDownloadLinkDescription %}{% endblock %}</a>
```
jonheslop: `download` attribute 😎

app/views/includes/footer-categories.njk
```diff
@@ -39,4 +39,19 @@
       </li>
     </ul>
   </div>
+  <div class="govuk-footer__section">
```
jonheslop: It’s a little crazy but we use this https://github.com/alphagov/pay-selfservice/blob/master/app/utils/display_converter.js is where permissions are injected into the template.

sfount: Perfect thanks, I'll check against that to render the third section selectively.

sfount: Addressed in https://github.com/alphagov/pay-selfservice/pull/1089/commits/1a4f01f06a7cf9c09ac839b19cccdac0b94fc204

app/views/policy/policy.njk
```diff
@@ -0,0 +1,29 @@
+{% extends "../layout.njk" %}
+
+{% block pageTitle %}
+   Policy Document Download - GOV.UK Pay
+{% endblock %}
+
+{% block mainContent %}
+<div class="govuk-grid-column-full">
+  <h1 class="govuk-heading-l page-title">
```
jonheslop: `page-title` is dotted around the place because lots of tests look for it, if you’re not using it for tests, you dont need it

PR #1091, opened 2019-01-16
PP-4451 Request to go live: "Organisation Name" static page

DanailMinchev:
## WHAT

This PR adds `Organisation Name` static page with Cypress testing.
Also, it fixes current service behaviour on the index page.

## HOW 

- Do not use `resolve_service` and `permissions` middleware at the same time when having service external ids in the path as it will wrongly resolve `currentService`, because it is not passing the `req.params` and the `gateway_account.currentGatewayAccountId` is used instead.
  For example multiple pages are open with service external ids in the path and then if we switch the current service/page then the `currentService` is wrongly evaluated in the second call of the `resolve_service` middleware.
- Added static template for the `Organisation Name` page and related tests

with @cobainc0


app/controllers/request-to-go-live/organisation-name/post.controller.js
```diff
@@ -0,0 +1,11 @@
+'use strict'
+
+// Local dependencies
+const goLiveStageToNextPagePath = require('../go-live-stage-to-next-page-path')
+
+module.exports = (req, res) => {
+  res.redirect(
+    303,
+    goLiveStageToNextPagePath[req.service.currentGoLiveStage].replace(':externalServiceId', req.service.externalId)
```
jonheslop: We have a util that can do this, but using a native JS func seems legit too

Util —> https://github.com/alphagov/pay-selfservice/blob/master/app/utils/replace_params_in_path.js

app/views/request-to-go-live/organisation-name.njk
```diff
@@ -0,0 +1,28 @@
...
+{% endblock %}
+
+{% block mainContent %}
+  <div class="govuk-grid-column-two-thirds govuk-!-margin-top-6">
+    <span id="request-to-go-live-current-step" class="govuk-caption-l">Step 1 of 3</span>
+    <h1 class="govuk-heading-l">What is the name of your organisation?</h1>
+
+    <form id="request-to-go-live-organisation-name-form" method="post" action="/service/{{currentService.externalId}}/request-to-go-live/organisation-name">
+      <input id="csrf" name="csrfToken" type="hidden" value="{{ csrf }}"/>
+      {{ govukInput({
```
jonheslop: Inputs need to have a label (for accessibility), your `h1` at line 13 is it I guess.

jonheslop: You could do it like
```html
label: {
    html: '
    <span id="request-to-go-live-current-step" class="govuk-caption-l">Step 1 of 3</span>
    <h1 class="govuk-heading-l">What is the name of your organisation?</h1>'
  },
```

But i dunno if this is dodgy, basically the page title is also the label. There is a pattern for this with guidance, will find it now.

jonheslop: Found it https://design-system.service.gov.uk/patterns/question-pages/#label-as-a-page-heading, so there is a boolean `isPageHeading: true`

app/views/request-to-go-live/organisation-name.njk
```diff
@@ -0,0 +1,28 @@
...
+{% from "components/error-summary/macro.njk" import govukErrorSummary %}
+
+{% block pageTitle %}
+  Request to go live - {{ currentService.name }} - GOV.UK Pay
+{% endblock %}
+
+{% block mainContent %}
+  <div class="govuk-grid-column-two-thirds govuk-!-margin-top-6">
+    <span id="request-to-go-live-current-step" class="govuk-caption-l">Step 1 of 3</span>
+    <h1 class="govuk-heading-l">What is the name of your organisation?</h1>
```
jonheslop: Basically, move this into the Nunjucks for the input but like

```suggestion
    label: {
      text: "What is the name of your organisation?",
      isPageHeading: true,
      classes: "govuk-heading-l"
    }
```

app/controllers/request-to-go-live/organisation-name/post.controller.js
DanailMinchev: Yes, thanks, we think if we have multiple params it's nice to use the custom util, but we think it is better not to depend on custom code in our case

app/views/request-to-go-live/organisation-name.njk
DanailMinchev: Yes, we are copying what is in the mockup, looks like mockup does not treat the label as h1, but this looks better

app/views/request-to-go-live/organisation-name.njk
DanailMinchev: We tried that, but the result looks different
screen shot 2019-01-16 at 17 25 27

jonheslop: ah sorry try `classes: "govuk-label--xl"` which is how it is in the example here — https://design-system.service.gov.uk/patterns/question-pages/#label-as-a-page-heading

DanailMinchev: It looks like we need to use two dashes for the class `govuk-label--l`, looks nice now, thanks!

PR #1099, opened 2019-01-18
PP-4451 Request to go live: "Organisation Name" page - display and validation logic

DanailMinchev:
## WHAT

This PR implements pre-population of the organisation name and validation.
See commits for more information.

with @cobainc0

### DIFF
--- a/app/views/request-to-go-live/organisation-name.njk
+++ b/app/views/request-to-go-live/organisation-name.njk
@@ -22,7 +22,8 @@
           text: "For example, Department for Transport"
         },
         id: "request-to-go-live-organisation-name-input",
-        name: "organisation-name"
+        name: "organisation-name",
+        value: organisationName
       }) }}
 
       {{ govukButton({ text: "Continue" }) }}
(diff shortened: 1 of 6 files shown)


app/views/request-to-go-live/organisation-name.njk
```diff
@@ -11,8 +11,29 @@
...
 
+    {% if errors %}
+      {% set errorList = [] %}
+      {% if errors['organisation-name'] %}
+        {% set errorList = (errorList.push({
+          text: 'What is the name of your organisation?',
+          href: '#request-to-go-live-organisation-name-input'
+        }), errorList) %}
+      {% endif %}
+      {{ govukErrorSummary({
```
jonheslop: The error summary should be at the top of the page - https://design-system.service.gov.uk/components/error-summary/ (as per the guidance) so step 1 of 3 should be below it.

Also don’t have to do it this way but every page has the flash component built in — https://github.com/alphagov/pay-selfservice/blob/master/app/views/layout.njk#L29 — so if you were to pass the error content into `req.flash` you wouldnt need any logic markup in the template. But 🤷‍♂️ either is fine I guess

DanailMinchev: Thanks for letting me know about Design System, as clarified we should not rely on mockups as html code (I was thinking the opposite before).

The `req.flash` is nice, but as discussed we will need to pass html directly from the controller which is not good way, so it needs to be implemented to handle list of errors with links etc. As discussed the current implementation with `govukErrorSummary` macro is better option in that case until `req.flash` has the other functionality.

PR #1107, opened 2019-01-23
PP-4451 Validate Organisation name maximum character length

cobainc0:
## WHAT
- prevents `organisation-name` characters from exceeding the upper limit of 255-characters, currently set in the db
 - removes organisation name default value from user fixture `self_service_user.json`to allow us to simulate realistic user behaviour
 - re-use validation provided in `field-validation-checks`
 - updates cypress tests to clear cookies prior to running tests to avoid persisting state
 - adds new cypress test for validating `organisation-name` character length
 - adds margin above error message because its visually more appealing

solo: @cobainc0

### DIFF
--- a/app/assets/sass/application.scss
+++ b/app/assets/sass/application.scss
@@ -36,3 +36,4 @@
 @import "components/check-your-answers";
 @import "components/borders";
 @import "components/request-to-go-live";
+@import "components/error-label";
--- a/app/controllers/request-to-go-live/organisation-name/post.controller.js
+++ b/app/controllers/request-to-go-live/organisation-name/post.controller.js
@@ -6,12 +6,15 @@ const lodash = require('lodash')
 // Local dependencies
 const goLiveStageToNextPagePath = require('../go-live-stage-to-next-page-path')
 const { requestToGoLive } = require('../../../paths')
+const { validateOrganisationName } = require('../../../utils/organisation_name_validation')
 
 // Constants
 const REQUEST_ORGANISATION_NAME_FIELD = 'organisation-name'
 
 module.exports = (req, res) => {
-  const errors = validateRequest(req)
+  const organisationName = lodash.get(req, 'body.organisation-name')
+  const errors = validateOrganisationName(organisationName, REQUEST_ORGANISATION_NAME_FIELD, true)
+
   if (lodash.isEmpty(errors)) {
     // TODO: handle submission
     res.redirect(
@@ -29,21 +32,4 @@ module.exports = (req, res) => {
       requestToGoLive.organisationName.replace(':externalServiceId', req.service.externalId)
     )
   }
-}
-
-function validateRequest (req) {
-  const mandatoryFields = [REQUEST_ORGANISATION_NAME_FIELD]
-
-  return validateNotEmpty(req, mandatoryFields)
-}
-
-function validateNotEmpty (req, fieldNames) {
-  const errors = {}
-  fieldNames.forEach(fieldName => {
-    let field = req.body[fieldName]
-    if (!field || typeof field !== 'string') {
-      errors[fieldName] = true
-    }
-  })
-  return errors
 }
--- a/app/utils/organisation_name_validation.js
+++ b/app/utils/organisation_name_validation.js
@@ -1,0 +1,20 @@
+'use strict'
+
+const trim = require('lodash/trim')
+const { isEmpty, isFieldGreaterThanMaxLengthChars } = require('../browsered/field-validation-checks')
+const ORGANISATION_NAME_MAX_LENGTH = 255
+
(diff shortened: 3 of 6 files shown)


app/utils/organisation_name_validation.js
```diff
@@ -0,0 +1,20 @@
+'use strict'
+
+const trim = require('lodash/trim')
+const { isEmpty, isFieldGreaterThanMaxLengthChars } = require('../browsered/field-validation-checks')
+const ORGANISATION_NAME_MAX_LENGTH = 50
```
DanailMinchev: I think we agreed on `255` when we discussed last time - https://github.com/alphagov/pay-adminusers/pull/373/files#diff-d10a0b9d3e0a9c446b2bb9cf543bd37eR24
Let me know if there are some updates on this, thanks

app/assets/sass/application.scss
```diff
@@ -36,3 +36,4 @@
 @import "components/check-your-answers";
 @import "components/borders";
 @import "components/request-to-go-live";
+@import "components/error-label";
```
DanailMinchev: Nice 👍

test/cypress/integration/request-to-go-live/organisation_name_spec.js
```diff
@@ -36,6 +38,7 @@ describe('Request to go live: organisation name', () => {
       cy.get('#request-to-go-live-organisation-name-form').should('exist')
 
       cy.get('input#request-to-go-live-organisation-name-input').should('exist')
+      cy.get('input#request-to-go-live-organisation-name-input').invoke('val', 'Government Digital Service')
```
stephencdaly: The behaviour of this test has now changed and isn't testing what it was before, i.e. that if the service already has an organisation name set, it is pre-filled on the page.
You'll have to add another service to the self_service_user configuration to allow for testing this behaviour 😿

app/controllers/request-to-go-live/organisation-name/post.controller.js
```diff
@@ -6,12 +6,15 @@ const lodash = require('lodash')
 // Local dependencies
 const goLiveStageToNextPagePath = require('../go-live-stage-to-next-page-path')
 const { requestToGoLive } = require('../../../paths')
+const { validateOrganisationName } = require('../../../utils/organisation_name_validation')
```
DanailMinchev: Great that you moved the logic to a separate file 👍 I think we use dashes for file delimiters when we were discussing last time

app/views/request-to-go-live/organisation-name.njk
```diff
@@ -49,7 +50,8 @@
         type: "text",
         errorMessage: organisationNameError,
         attributes: {
-          "data-validate": "required"
+          "data-validate": "required isFieldGreaterThanMaxLengthChars",
```
DanailMinchev: Awesome 👍

test/cypress/integration/request-to-go-live/organisation_name_spec.js
```diff
@@ -3,6 +3,7 @@ describe('Request to go live: organisation name', () => {
 
   describe('NO PERMISSIONS', () => {
     beforeEach(() => {
+      cy.clearCookies()
```
stephencdaly: I don't think clearing the cookies is necessary as they will automatically be cleared between each test

app/utils/organisation_name_validation.js
cobainc0: good spot, i missed this

test/fixtures/config/self_service_user.json
```diff
@@ -1133,7 +1133,7 @@
                 "666"
               ],
               "merchant_details": {
-                "name": "Government Digital Service",
```
DanailMinchev: I think we have existing test with this, so probably it's better do delete the predefined value and then submit in the test

app/controllers/request-to-go-live/organisation-name/post.controller.js
cobainc0: I'm not sure why but we seem to use snake-casing in the utils folder

app/views/request-to-go-live/organisation-name.njk
cobainc0: Thanks 😅

app/controllers/request-to-go-live/organisation-name/post.controller.js
DanailMinchev: Ah, ok, I didn't notice that, let's keep it consistent per directory then

test/cypress/integration/request-to-go-live/organisation_name_spec.js
cobainc0: Thanks, i'll update this 👍

PR #1119, opened 2019-01-29
use tel type for telephone fields

gidsg:
## WHAT
Use `tel` type for phone input fields. This is so the user agent can know how to handle this field. This is as per design system https://design-system.service.gov.uk/patterns/telephone-numbers/

## HOW 
go to: http://localhost:9400/create-service/register
input type on phone field should be `tel`
e.g. on iOS special input type should be displayed
img-1021

### DIFF
--- a/app/views/self_create_service/register.njk
+++ b/app/views/self_create_service/register.njk
@@ -34,6 +34,7 @@
       govukInput({
         id: "telephone-number",
         name: "telephone-number",
+        type: "tel",
         classes: "govuk-input--width-20",
         label: {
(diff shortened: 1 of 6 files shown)


gidsg:
> There’s also a telephone number input somewhere on the organisation (merchant) details page.

@alexbishop1 Yup, I found this one but it already uses `tel` https://github.com/alphagov/pay-selfservice/blob/c21230bb7b226068ef5e3d8534c6dc75aba300c0/app/views/merchant_details/edit_merchant_details.njk#L110

PR #1111, opened 2019-01-24
PP-4678 Modify Cypress tests to use Mountebank for stubbing

stephencdaly:
### How it works
- Use Mountebank as virtualisation server instead of pact-stub-server
- Stubs are set up on Mountebank by POSTing to the imposters endpoint on the running mountebank instance, managed in cypress/plugins/index.js
- Stubs are setup from within the Cypress specs by calling a cypress task as part of the initialisation. Stubs are cleared before tests are run as part of a common before step.
- Cypress tests no longer use the self_service_user.json static configuration file. Instead values in the response we want to assert on or share between requests are passed into the stub setup.

### Other changes
- Modifies some fixture builders to allow for greater flexibility and re-usability
- Modifies Cypress tests to get them working with the new stubbing method but aims to preserve the integrity of the tests
- Modifies cookie setup in Cypress tests to remove reliance on static configuration JSON file.

## HOW TO REVIEW
There are a few main parts to it which could probably be reviewed by different people, maybe in a few passes. These are:

1. General setup and use of stubs. This is mainly isolated to cypress/plugins/index.js, cypress/plugins/stubs.js and cypress/support/index.js. Someone reviewing this would probably want to look at a couple of the cypress tests focusing on the stub setup. Included in this is also the setup of the cookies for the Cypress tests.

2. Modifications to the Cypress tests. I've had to make modifications to all of the Cypress tests get them working. It would be good if someone could sanity check these.

3. Fixture builder modifications. I've modified some of the fixture builders for ease of creating customised responses from the cypress tests. I've attempted here to make a first pass at making what will be a single set of fixtures to build the various objects that make up the responses. The aim is in future to get unit/pact tests to use this single set of fixtures so they are well tested and we don't have several duplicated fixtures creating the same objects that have got out of date. These aren't perfect currently, but are good enough to get the cypress tests working.

PR #1124, opened 2019-01-31
BAU Invoke cleanup script from jenkins library

sfount:
Cleanup code exists in the Cypress groovy jenkins library to clean up
docker-compose, however it looks like this is never run, meaning that
Docker containers and networks are left hanging after job completion.

Relates to https://github.com/alphagov/pay-frontend/pull/620

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -39,6 +39,13 @@ pipeline {
       steps {
         cypress('selfservice')
       }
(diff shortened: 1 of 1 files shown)


sfount:
This change has been incorporated in https://github.com/alphagov/pay-selfservice/pull/1111

PR #1130, opened 2019-02-01
PP-4678 Move Cypress specs into folders

stephencdaly:
## WHAT
- Specs were mostly in `users` folder, even though most weren't to do
  with users. Moved into more appropriate folders
- Move cookie-monster.js into plugins folder, where it belongs


test/cypress/plugins/index.js
```diff
@@ -12,7 +12,7 @@ const lodash = require('lodash')
 const request = require('request-promise-native')
 
 // Local dependencies
-const cookieMonster = require('../integration/utils/cookie-monster')
+const cookieMonster = require('./cookie-monster')
```
alexbishop1: 🍪

PR #1131, opened 2019-02-01
PP-4678 Remove use of self_service_user.json from adminusers tests

stephencdaly:
## WHAT
- First steps towards removing self_service_user.json completely
- Remove pact tests to get service with each go live stage as these are unnecessary now that we don't use pacts to drive Cypress tests


test/unit/clients/adminusers_client/user/get_user_test.js
```diff
@@ -34,44 +33,37 @@ describe('adminusers client - get user', () => {
   before(() => provider.setup())
   after(done => provider.finalize().then(done()))
 
-  selfServiceUserConfig.config.users.forEach(currentUser => {
-    describe(`success "${currentUser.cypressTestingCategory}" user`, () => {
-      const existingExternalId = currentUser.external_id
+  describe('success', () => {
```
sfount: This describe could probably be more descriptive. I know it's a child of a larger block but something like 'find a valid user' might mean more to someone looking at this in isolation than 'success'

stephencdaly: good point 👍 , I've changed it

PR #1133, opened 2019-02-01
PP-4678 Remove self_service_user from email patch tests

stephencdaly:
## WHAT
- remove usages of static config file which will be removed in future
- modify fixtures for email setting patch requests to be stricter

### DIFF
--- a/test/fixtures/gateway_account_fixtures.js
+++ b/test/fixtures/gateway_account_fixtures.js
@@ -9,11 +9,11 @@ const pactBase = require(path.join(__dirname, '/pact_base'))
 const pactRegister = pactBase()
 
 module.exports = {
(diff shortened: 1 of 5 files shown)


test/fixtures/gateway_account_fixtures.js
```diff
@@ -9,11 +9,11 @@ const pactBase = require(path.join(__dirname, '/pact_base'))
 const pactRegister = pactBase()
 
 module.exports = {
-  validGatewayAccountEmailRefundToggleRequest: (opts = {}) => {
+  validGatewayAccountEmailRefundToggleRequest: (enabled = true) => {
```
sfount: Nice use of default params + a more specific API 👍

PR #1134, opened 2019-02-04
BAU: Reduce calls to adminusers/api/users

sfount:
All static assets (JS/ CSS) are currently served behind our
authorisation middlware, this requires making a request and
deserialising the user for these resources.

Bump up the middleware so that it is no longer private.

With @stephencdaly, @SandorArpa

Edit: do not merge - sanity check Auth corner cases

### DIFF
--- a/server.js
+++ b/server.js
@@ -159,8 +159,8 @@ function initialise () {
   initialiseTLS()
   initialisePublic(app)
   initialiseCookies(app)
-  initialiseAuth(app)
   initialiseGlobalMiddleware(app)
+  initialiseAuth(app)
   initialiseTemplateEngine(app)
   initialiseErrorHandling(app)
   initialiseRoutes(app) // This contains the 404 overrider and so should be last


heathd:
It's not very evident from the name that 'global middleware' is primarily the thing that serves static assets.

I can't see any security risk in serving static assets without auth, I'm sure this was an oversight.

It may not be evident to people reading this code that the order of middleware loading also controls the order that it's executed. A comment to that effect might be useful.

Also I think it might be worth splitting the 'initilizeGlobalMiddleware' function to separate into:
- middleware which serves static assets
- body parser middleware

that way it might be more self-evident why the static asset serving is prior to the auth?

sfount:
Agreed! This is a huge middleware that has lots of different behaviours. I'll look into splitting these up this week.

sfount:
With @heathd -- looks like a minor optimisation but as most assets are cached in the browser anyway it won't impact many users. 

Potential optimisation for future express apps.

PR #1144, opened 2019-02-06
PP-4451 Implements logic for updating Organisation name

cobainc0:
## WHAT
 - add PATCH request to allow selfservice to update merchant name in adminusers
 - updates cypress tests to incorporate new behaviour
 - minor refactoring

solo: @cobainc0

### DIFF
--- a/app/controllers/request-to-go-live/organisation-name/post.controller.js
+++ b/app/controllers/request-to-go-live/organisation-name/post.controller.js
@@ -5,27 +5,36 @@ const lodash = require('lodash')
 
 // Local dependencies
 const goLiveStageToNextPagePath = require('../go-live-stage-to-next-page-path')
+const goLiveStage = require('../../../models/go-live-stage')
 const { requestToGoLive } = require('../../../paths')
 const { validateOrganisationName } = require('../../../utils/organisation_name_validation')
-
+const { updateCurrentGoLiveStage, updateMerchantName } = require('../../../services/service_service')
+const { renderErrorView } = require('../../../utils/response')
 // Constants
-const REQUEST_ORGANISATION_NAME_FIELD = 'organisation-name'
+const ORGANISATION_NAME_FIELD = 'organisation-name'
 
 module.exports = (req, res) => {
-  const organisationName = lodash.get(req, 'body.organisation-name')
-  const errors = validateOrganisationName(organisationName, REQUEST_ORGANISATION_NAME_FIELD, true)
-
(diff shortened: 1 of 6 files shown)


app/controllers/request-to-go-live/organisation-name/post.controller.js
```diff
@@ -5,27 +5,36 @@ const lodash = require('lodash')
...
   if (lodash.isEmpty(errors)) {
-    // TODO: handle submission
-    res.redirect(
-      303,
-      goLiveStageToNextPagePath[req.service.currentGoLiveStage].replace(':externalServiceId', req.service.externalId)
-    )
+    return updateMerchantName(req.service.externalId, organisationName, req.correlationId)
+      .then(service => {
+        return updateCurrentGoLiveStage(service.externalId, goLiveStage.ENTERED_ORGANISATION_NAME, req.correlationId)
+          .then(updatedService => {
```
sfount: This promise structure is nested in the same way a callback chain might be, this code should actually work but it might become hard to follow in the future, especially for error handling. 

The goal should be to only ever have a depth of 1 in promise chains, that way the error is handled uniformly at the bottom and there aren't small issues with promises not being returned further down the chain. 

(we also don't need to return anything to express route handlers)

This would look something like: 
```js 
updateMerchantName()
  .then((service) => { 
    return updateCurrentGoLiveStage()
  })
  .then((updatedService) => { 
    // service successfully updated, redirect user to next stage
    res.redirect()
  })
  .catch((error) => { 
    renderErrorView()
  })
```

app/controllers/request-to-go-live/organisation-name/post.controller.js
```diff
@@ -5,27 +5,36 @@ const lodash = require('lodash')
...
+      })
+      .catch(err => {
+        renderErrorView(req, res, err.message)
+      })
   } else {
     lodash.set(req, 'session.pageData.requestToGoLive.organisationName', {
       success: false,
       errors: errors,
-      organisationName: req.body[REQUEST_ORGANISATION_NAME_FIELD]
+      organisationName: req.body[ORGANISATION_NAME_FIELD]
```
sfount: This constant makes the code very clear! 👍

cobainc0: Gracias

app/controllers/request-to-go-live/organisation-name/post.controller.js
cobainc0: Awesome thanks!  I've updated this now 👌🏿

PR #1145, opened 2019-02-06
PP-4453 Add patch request to go live stage pact test

SandorArpa:
## WHAT
- add pact test

### DIFF
--- a/test/fixtures/service_fixtures.js
+++ b/test/fixtures/service_fixtures.js
@@ -287,6 +287,23 @@ module.exports = {
     }
   },
 
+  validUpdateRequestToGoLiveRequest: (value = 'AGREED_TO_STRIPE') => {
+    const data = {
+      op: 'replace',
+      path: 'current_go_live_stage',
+      value: value
+    }
+
+    return {
+      getPactified: () => {
+        return pactServices.pactify(data)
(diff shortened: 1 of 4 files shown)


test/fixtures/service_fixtures.js
```diff
@@ -287,6 +287,25 @@ module.exports = {
...
   },
 
+  validUpdateRequestToGoLiveRequest: (opts) => {
+    opts = opts || {}
+
+    const data = {
+      op: opts.op || 'replace',
+      path: opts.path || 'current_go_live_stage',
+      value: opts.value || 'ENTERED_ORGANISATION_NAME'
+    }
```
stephencdaly: Could you make this fixture stricter by only passing in the `value` as a parameter?
We should then use this fixture directly in the cypress tests for stubbing the update go live stage requests.

This will give us confidence that making a patch request to the endpoint with the path `current_go_live_stage` works, and if this changes the Cypress tests won't get out of date.

stephencdaly:
Could you change the `patchGoLiveStageSuccess` stub in `test/cypress/plugins/stubs.js` to use the new fixture you've added so that we maintain the link between the pacts and Cypress tests?

PR #1152, opened 2019-02-07
PP-4451 organisation name page

cobainc0:
## WHAT
Add contract test for merchant details PATCH request
 - new pact test `update_merchant_name_test` verifies that a
 request to the PATCH endpoint in adminUsers updates the merchant name
 successfully

Fixes flaky browser tests
 - adds missing assertions to the end of tests where previous flaky behaviour was exhibited. See SHA: 143a9b2
 - refactors `organisation_name_spec` to incorporate new utils in `request_to_go_live_utils`.

### DIFF
--- a/test/cypress/integration/request-to-go-live/organisation_name_spec.js
+++ b/test/cypress/integration/request-to-go-live/organisation_name_spec.js
@@ -1,38 +1,17 @@
+'use strict'
+
 const lodash = require('lodash')
+const utils = require('../../utils/request_to_go_live_utils')
+const { userExternalId, gatewayAccountId, serviceExternalId } = utils.variables
 
 describe('Request to go live: organisation name page', () => {
-  const userExternalId = 'cd0fa54cf3b7408a80ae2f1b93e7c16e'
-  const gatewayAccountId = 42
-  const serviceExternalId = 'afe452323dd04d1898672bfaba25e3a6'
   const organisationName = 'Government Digital Service'
 
-  const buildServiceRoleForGoLiveStage = (goLiveStage) => {
-    return {
-      service: {
-        external_id: serviceExternalId,
-        current_go_live_stage: goLiveStage,
-        gateway_account_ids: [gatewayAccountId]
-      }
-    }
-  }
+  beforeEach(() => {
+    cy.setEncryptedCookies(userExternalId, gatewayAccountId)
+  })
 
-  const simpleStub = (serviceRole) => {
-    return [
-      {
-        name: 'getUserSuccess',
-        opts: {
-          external_id: userExternalId,
-          service_roles: [serviceRole]
-        }
-      },
-      {
-        name: 'getGatewayAccountSuccess',
-        opts: { gateway_account_id: gatewayAccountId }
-      }
-    ]
-  }
-
-  const stubsWithGoLiveStageAndOrganisationName = (currentGoLiveStage = 'NOT_STARTED', organisationName = undefined) => {
+  const stubPatchRequests = (currentGoLiveStage, organisationName) => {
     return [{
       name: 'patchUpdateServiceSuccess',
       opts: {
@@ -44,7 +23,7 @@ describe('Request to go live: organisation name page', () => {
       }
     },
     {
-      name: 'patchUpdateServiceSuccess',
+      name: 'patchSingleMerchantDetailsSuccess',
       opts: {
         external_id: serviceExternalId,
         gateway_account_ids: [gatewayAccountId],
@@ -55,21 +34,11 @@ describe('Request to go live: organisation name page', () => {
     }]
   }
 
-  const setupStubs = (serviceRole) => {
-    cy.task('setupStubs', simpleStub(serviceRole))
-  }
-
-  beforeEach(() => {
-    cy.setEncryptedCookies(userExternalId, gatewayAccountId)
-  })
-
   describe('User does not have the correct permissions', () => {
+    const serviceRole = utils.buildServiceRoleForGoLiveStage('NOT_STARTED')
+    serviceRole.role = { permissions: [ ] }
     beforeEach(() => {
-      const serviceRole = buildServiceRoleForGoLiveStage('NOT_STARTED')
-      serviceRole.role = {
-        permissions: []
-      }
-      setupStubs(serviceRole)
+      utils.setupStubs(serviceRole)
     })
 
     it('should show an error when the user does not have enough permissions', () => {
@@ -80,11 +49,31 @@ describe('Request to go live: organisation name page', () => {
     })
   })
 
-  describe('Service has correct go live stage and organisation name is not pre-filled', () => {
+  describe('Service has invalid go live stage', () => {
+    const serviceRole = utils.buildServiceRoleForGoLiveStage('INVALID_GO_LIVE_STAGE')
     beforeEach(() => {
-      const serviceRole = buildServiceRoleForGoLiveStage('NOT_STARTED')
-      const stubPayload = lodash.concat(simpleStub(serviceRole), stubsWithGoLiveStageAndOrganisationName(
-        'ENTERED_ORGANISATION_NAME'))
+      utils.setupStubs(serviceRole)
+    })
+    it('should redirect to "Request to go live: index" page when in wrong stage', () => {
+      const requestToGoLivePageOrganisationNameUrl = `/service/${serviceExternalId}/request-to-go-live/organisation-name`
+      cy.visit(requestToGoLivePageOrganisationNameUrl)
+
+      cy.get('h1').should('contain', 'Request to go live')
+
+      cy.location().should((location) => {
+        expect(location.pathname).to.eq(`/service/${serviceExternalId}/request-to-go-live`)
+      })
+    })
+  })
+
+  describe('Service has NOT_STARTED go live stage and organisation name is not pre-filled', () => {
+    const serviceRoleBefore = utils.buildServiceRoleForGoLiveStage('NOT_STARTED')
+    const serviceRoleAfter = utils.buildServiceRoleForGoLiveStage('ENTERED_ORGANISATION_NAME')
+    const stubPayload = lodash.concat(
+      utils.stubUserSuccessResponse(serviceRoleBefore, serviceRoleAfter),
+      stubPatchRequests('ENTERED_ORGANISATION_NAME', organisationName))
+
+    beforeEach(() => {
       cy.task('setupStubs', stubPayload)
     })
 
@@ -99,37 +88,35 @@ describe('Request to go live: organisation name page', () => {
       cy.get('#request-to-go-live-organisation-name-form').should('exist')
       cy.get('input#request-to-go-live-organisation-name-input').should('exist')
 
+      cy.get('input#request-to-go-live-organisation-name-input').should('be.empty')
+
       cy.get('input#request-to-go-live-organisation-name-input').type(organisationName)
       cy.get('input#request-to-go-live-organisation-name-input').should('have.value', organisationName)
 
       cy.get('#request-to-go-live-organisation-name-form > button').should('exist')
       cy.get('#request-to-go-live-organisation-name-form > button').should('contain', 'Continue')
-    })
+      cy.get('#request-to-go-live-organisation-name-form > button').click()
 
-    it('should show empty input box if organisation name is not pre-filled', () => {
-      const requestToGoLivePageOrganisationNameUrl = `/service/${serviceExternalId}/request-to-go-live/organisation-name`
-      cy.visit(requestToGoLivePageOrganisationNameUrl)
-
-      cy.get('h1').should('contain', 'What is the name of your organisation?')
-
-      cy.get('#request-to-go-live-current-step').should('exist')
-
-      cy.get('#request-to-go-live-organisation-name-form').should('exist')
-      cy.get('input#request-to-go-live-organisation-name-input').should('exist')
-      cy.get('input#request-to-go-live-organisation-name-input').should('be.empty')
+      cy.location().should((location) => {
+        expect(location.pathname).to.eq(`/service/${serviceExternalId}/request-to-go-live/choose-how-to-process-payments`)
+      })
     })
   })
 
-  describe('Service has correct go live stage and organisation name is pre-filled', () => {
+  describe('Service has NOT_STARTED go live stage and organisation name is pre-filled', () => {
+    const serviceRoleBefore = utils.buildServiceRoleForMerchantDetailsField({ name: organisationName }, 'NOT_STARTED')
+    const serviceRoleAfter = utils.buildServiceRoleForGoLiveStage('ENTERED_ORGANISATION_NAME')
+    const changeOrganisationName = 'GDS'
+
+    const stubPayload = lodash.concat(
+      utils.stubUserSuccessResponse(serviceRoleBefore, serviceRoleAfter),
+      stubPatchRequests('ENTERED_ORGANISATION_NAME', changeOrganisationName))
+
     beforeEach(() => {
-      const serviceRole = buildServiceRoleForGoLiveStage('NOT_STARTED')
-      serviceRole.service.merchant_details = { name: organisationName }
-      const stubPrefilledPayload = lodash.concat(simpleStub(serviceRole), stubsWithGoLiveStageAndOrganisationName(
-        'ENTERED_ORGANISATION_NAME', organisationName))
-      cy.task('setupStubs', stubPrefilledPayload)
+      cy.task('setupStubs', stubPayload)
     })
 
-    it('should show page correctly with pre-filled organisation name', () => {
+    it('should show page correctly and allow users to change pre-filled organisation name successfully', () => {
       const requestToGoLivePageOrganisationNameUrl = `/service/${serviceExternalId}/request-to-go-live/organisation-name`
       cy.visit(requestToGoLivePageOrganisationNameUrl)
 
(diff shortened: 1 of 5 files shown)


test/cypress/integration/request-to-go-live/organisation_name_spec.js
```diff
@@ -1,40 +1,22 @@
+'use strict'
+
 const lodash = require('lodash')
+const utils = require('../../utils/request_to_go_live_utils')
+const variables = utils.variables
```
stephencdaly: You could use object destructuring to save having to initialise the id variables later.

So:
```const { userExternalId, gatewayAccountId, serviceExternalId } = utils.variables```

File: test/cypress/integration/request-to-go-live/organisation_name_spec.js [hunk 2]
```diff
@@ -55,21 +37,33 @@ describe('Request to go live: organisation name page', () => {
     }]
   }
 
-  const setupStubs = (serviceRole) => {
-    cy.task('setupStubs', simpleStub(serviceRole))
+  const serviceRoleBeforePatchRequests = utils.buildServiceRoleForMerchantDetailsField({ name: organisationName })
```
stephencdaly: Is this setup re-used enough to warrant going into the utils?
Also do we also want to explicitly set the go-live stage? I think the default for the fixture is 'NOT_STARTED' but we maybe shouldn't rely on that, and explicitly passing it in helps for visibility.

test/cypress/integration/request-to-go-live/organisation_name_spec.js
```diff
@@ -55,21 +37,33 @@ describe('Request to go live: organisation name page', () => {
...
+  const serviceRoleAfterPatchRequests = utils.buildServiceRoleForGoLiveStage('ENTERED_ORGANISATION_NAME')
+  function getUserSuccess (serviceRolesBefore = [serviceRoleBeforePatchRequests], serviceRolesAfter = serviceRoleAfterPatchRequests) {
```
stephencdaly: I think using the default parameters here makes it slightly confusing, I think it would be better to create the service roles per test and always pass them in.

stephencdaly: Is this function also a candidate to go into the utils, as we use it in other _specs?

test/cypress/integration/request-to-go-live/organisation_name_spec.js
stephencdaly: Before we had this which is maybe clearer
```
const serviceRole = buildServiceRoleForGoLiveStage('NOT_STARTED')
serviceRole.service.merchant_details = { name: organisationName }
````

File: test/cypress/integration/request-to-go-live/organisation_name_spec.js [hunk 4]
```diff
@@ -80,11 +74,29 @@ describe('Request to go live: organisation name page', () => {
...
+
+      cy.location().should((location) => {
+        expect(location.pathname).to.eq(`/service/${serviceExternalId}/request-to-go-live`)
+      })
+    })
+  })
+
+  describe('Service has NOT_STARTED go live stage and organisation name is not pre-filled', () => {
+    const serviceRoleBeforePatchRequests = utils.buildServiceRoleForGoLiveStage('NOT_STARTED')
+    const serviceRoleAfterPatchRequests = utils.buildServiceRoleForGoLiveStage('ENTERED_ORGANISATION_NAME')
```
stephencdaly: I like the naming of these variables, I think it makes the confusing thing we are doing with the different stub responses a lot clearer 👍

test/cypress/integration/request-to-go-live/organisation_name_spec.js
```diff
@@ -99,34 +111,26 @@ describe('Request to go live: organisation name page', () => {
...
-      cy.get('input#request-to-go-live-organisation-name-input').should('be.empty')
+      cy.location().should((location) => {
+        expect(location.pathname).to.eq(`/service/${serviceExternalId}/request-to-go-live/choose-how-to-process-payments`)
+      })
     })
   })
 
-  describe('Service has correct go live stage and organisation name is pre-filled', () => {
+  describe('Service has NOT_STARTED go live stage and organisation name is pre-filled', () => {
+    const stubUserSuccessResponse = getUserSuccess()
```
stephencdaly: As I said above, I think setting up the service roles with the go-live stage explicitly for this test would be clearer. Some code duplication in tests is often not such a bad thing if it improves readability.

test/fixtures/service_fixtures.js
```diff
@@ -304,6 +304,48 @@ module.exports = {
...
+
+    return {
+      getPactified: () => {
+        return pactServices.pactify(data)
+      },
+      getPlain: () => {
+        return _.clone(data)
+      }
+    }
+  },
```
stephencdaly: This fixture isn't needed, we should just use `validServiceResponse` as the request returns a service.

test/cypress/plugins/stubs.js
```diff
@@ -475,6 +475,31 @@ module.exports = {
       }
     ]
   },
+  patchSuccess: (opts = {}) => {
```
stephencdaly: This should be a stub that is only used to update the merchant name, and should use the new `validUpdateMerchantNameRequest` fixture you've added.

This will improve our confidence in the Cypress tests, as we know that we're building a request using a strict fixture that has been validated by the Pact tests. In this case, the Pact tests validate that the `path` in the request, `merchant_details/name` is a valid path as this is hardcoded into the fixture.

test/unit/clients/adminusers_client/service/update_merchant_name_test.js
```diff
@@ -0,0 +1,72 @@
...
+    spec: 2,
+    pactfileWriteMode: 'merge'
+  })
+
+  before(() => provider.setup())
+  after(done => provider.finalize().then(done()))
+
+  describe('a valid update merchant name patch request', () => {
+    const serviceExternalId = 'cp5wa'
+    const serviceName = 'serviceName'
```
stephencdaly: We don't care about the service name in this test so no need to put it in the requests or validate on it.

test/unit/clients/adminusers_client/service/update_merchant_name_test.js
```diff
@@ -0,0 +1,72 @@
...
+    const validUpdateMerchantNameRequest = serviceFixtures.validUpdateMerchantNameRequest('updated-merchant-details-name')
+    const validUpdateMerchantNameResponse = serviceFixtures.validUpdateMerchantNameResponse({
+      external_id: serviceExternalId,
+      name: serviceName
```
stephencdaly: the response should use the `validServiceResponse` fixture and include:
```
merchant_details: {
    name: 'updated-merchant-details-name'
}
````

It looks like currently it works due to the default values in the fixture you've added, but any field we are asserting on should be explicitly passed into the fixture as we shouldn't rely on the fixture defaults.

File: test/fixtures/service_fixtures.js [hunk 10]
```diff
@@ -304,6 +304,48 @@ module.exports = {
...
```
stephencdaly: Good fixture 👍

test/cypress/integration/request-to-go-live/organisation_name_spec.js
cobainc0: Thanks. I was gonna do this but opted to follow the other integration tests for consistency. I guess we could start refactoring once we're done, so we're not all working on the same common modules at once.

test/cypress/integration/request-to-go-live/organisation_name_spec.js
cobainc0: It might be useful when we want to update other merchant details, which I think we still plan to do.

cobainc0: I agree, it does look a lot clearer. I've updated this! 👍

test/cypress/integration/request-to-go-live/organisation_name_spec.js
cobainc0: Good thinking! I've added this to the utils folder so it can be used for future refactoring. I guess we could start refactoring once we're done, so we're not all working on the same common modules at once.

test/unit/clients/adminusers_client/service/update_merchant_name_test.js
cobainc0: Thanks, I have updated this to:
`merchant_details: {
        name: opts.value || ''
      }`

I have amended a test, now on line 121, to additionally assert on when a pre-filled organisation name is updated prior to submitting the form.

test/fixtures/service_fixtures.js
cobainc0: nuked

test/unit/clients/adminusers_client/service/update_merchant_name_test.js
cobainc0: also nuked

test/fixtures/service_fixtures.js
```diff
@@ -281,7 +298,10 @@ module.exports = {
       },
       redirect_to_service_immediately_on_terminal_state: false,
       collect_billing_address: false,
-      current_go_live_stage: 'NOT_STARTED'
+      current_go_live_stage: 'NOT_STARTED',
+      merchant_details: {
+        name: opts.value || ''
+      }
```
stephencdaly: I don't think we want to do this, as we can have a service response with no merchant details if there are none defined.
If you pass in any merchant details in the opts currently, it will build with the defaults defined in buildMerchantDetailsWithDefaults.

test/cypress/utils/request_to_go_live_utils.js
```diff
@@ -60,11 +70,30 @@ const setupStubs = (serviceRole) => {
   cy.task('setupStubs', simpleStub(serviceRole))
 }
 
+const stubUserSuccessResponse = (serviceRolesBefore, serviceRolesAfter) =>
+  [{
+    name: 'getUserSuccessRepeatFirstResponseNTimes',
+    opts: [{
+      external_id: variables.userExternalId,
+      service_roles: [serviceRolesBefore],
```
stephencdaly: This is just a single role you're passing in as a parameter, so the variable names should probably be `serviceRoleBefore` and `serviceRoleAfter`

test/cypress/integration/request-to-go-live/organisation_name_spec.js
```diff
@@ -80,11 +49,33 @@ describe('Request to go live: organisation name page', () => {
...
+        expect(location.pathname).to.eq(`/service/${serviceExternalId}/request-to-go-live`)
+      })
+    })
+  })
+
+  describe('Service has NOT_STARTED go live stage and organisation name is not pre-filled', () => {
+    const serviceRolesBefore = utils.buildServiceRoleForGoLiveStage('NOT_STARTED')
+    const serviceRolesAfter = lodash.concat(
+      utils.buildServiceRoleForGoLiveStage('ENTERED_ORGANISATION_NAME'),
+      utils.buildServiceRoleForMerchantDetailsField({ name: undefined }))
```
stephencdaly: I'm not sure this works, as the method only expects a single service role.

test/cypress/integration/request-to-go-live/organisation_name_spec.js
```diff
@@ -99,37 +90,35 @@ describe('Request to go live: organisation name page', () => {
...
-      cy.get('input#request-to-go-live-organisation-name-input').should('be.empty')
+      cy.location().should((location) => {
+        expect(location.pathname).to.eq(`/service/${serviceExternalId}/request-to-go-live/choose-how-to-process-payments`)
+      })
     })
   })
 
-  describe('Service has correct go live stage and organisation name is pre-filled', () => {
+  describe('Service has NOT_STARTED go live stage and organisation name is pre-filled', () => {
+    const serviceRolesBefore = utils.buildServiceRoleForMerchantDetailsField({ name: organisationName })
```
stephencdaly: Would be good to explicitly set the `current_go_live_stage` in the stub here, as currently we are relying on the fixture default of `NOT_STARTED` which we can't rely on to not change, maybe make this a parameter of the method.

test/fixtures/service_fixtures.js
stephencdaly: Just remove this change and I think everything will be fine

test/cypress/utils/request_to_go_live_utils.js
```diff
@@ -60,11 +70,30 @@ const setupStubs = (serviceRole) => {
   cy.task('setupStubs', simpleStub(serviceRole))
 }
 
+const stubUserSuccessResponse = (serviceRolesBefore, serviceRolesAfter) =>
```
stephencdaly:
```suggestion
const stubUserSuccessResponse = (serviceRoleBefore, serviceRoleAfter) =>
```

test/cypress/utils/request_to_go_live_utils.js
stephencdaly:
```suggestion
      service_roles: [serviceRoleBefore],
```

test/cypress/utils/request_to_go_live_utils.js
```diff
@@ -60,11 +70,30 @@ const setupStubs = (serviceRole) => {
...
+  [{
+    name: 'getUserSuccessRepeatFirstResponseNTimes',
+    opts: [{
+      external_id: variables.userExternalId,
+      service_roles: [serviceRolesBefore],
+      repeat: 2
+    }, {
+      external_id: variables.userExternalId,
+      service_roles: [serviceRolesAfter],
```
stephencdaly:
```suggestion
      service_roles: [serviceRoleAfter],
```

test/cypress/utils/request_to_go_live_utils.js
```diff
@@ -16,6 +16,16 @@ const buildServiceRoleForGoLiveStage = (goLiveStage) => {
   }
 }
 
+const buildServiceRoleForMerchantDetailsField = (merchantDetail) => {
```
stephencdaly:
```suggestion
const buildServiceRoleForMerchantDetailsField = (merchantDetails, currentGoLiveStage) => {
```

test/cypress/utils/request_to_go_live_utils.js
```diff
@@ -16,6 +16,16 @@ const buildServiceRoleForGoLiveStage = (goLiveStage) => {
...
+  return {
+    service: {
+      external_id: variables.serviceExternalId,
+      gateway_account_ids: [variables.gatewayAccountId],
+      merchant_details: merchantDetail
```
stephencdaly:
```suggestion
      merchant_details: merchantDetail,
      current_go_live_stage: currentGoLiveStage
```

test/cypress/integration/request-to-go-live/organisation_name_spec.js
stephencdaly:
```suggestion
    const serviceRolesBefore = utils.buildServiceRoleForMerchantDetailsField({ name: organisationName }, 'NOT_STARTED')
```

test/cypress/integration/request-to-go-live/organisation_name_spec.js
```diff
@@ -80,11 +49,33 @@ describe('Request to go live: organisation name page', () => {
...
+      cy.get('h1').should('contain', 'Request to go live')
+
+      cy.location().should((location) => {
+        expect(location.pathname).to.eq(`/service/${serviceExternalId}/request-to-go-live`)
+      })
+    })
+  })
+
+  describe('Service has NOT_STARTED go live stage and organisation name is not pre-filled', () => {
+    const serviceRolesBefore = utils.buildServiceRoleForGoLiveStage('NOT_STARTED')
```
stephencdaly:
```suggestion
    const serviceRoleBefore = utils.buildServiceRoleForGoLiveStage('NOT_STARTED')
```

test/cypress/integration/request-to-go-live/organisation_name_spec.js
```diff
@@ -99,37 +90,35 @@ describe('Request to go live: organisation name page', () => {
...
+      cy.location().should((location) => {
+        expect(location.pathname).to.eq(`/service/${serviceExternalId}/request-to-go-live/choose-how-to-process-payments`)
+      })
     })
   })
 
-  describe('Service has correct go live stage and organisation name is pre-filled', () => {
+  describe('Service has NOT_STARTED go live stage and organisation name is pre-filled', () => {
+    const serviceRolesBefore = utils.buildServiceRoleForMerchantDetailsField({ name: organisationName })
+    const serviceRolesAfter = utils.buildServiceRoleForGoLiveStage('ENTERED_ORGANISATION_NAME')
```
stephencdaly:
```suggestion
    const serviceRoleAfter = utils.buildServiceRoleForGoLiveStage('ENTERED_ORGANISATION_NAME')
```

test/cypress/plugins/stubs.js
stephencdaly:
```suggestion
  patchSingleMerchantDetailsSuccess: (opts = {}) => {
```

test/cypress/integration/request-to-go-live/organisation_name_spec.js
cobainc0: This was a slightly hacky attempt to make it clear that merchant details are not pre-filled. Also, thanks for clarifying the `validServiceResponse` default values in `service_fixtures` earlier,  i didn't notice it at all. When I see a code-block end `}` I still assume the method is done.

46bit:
The build may fail because I had to reboot CI. Re-run it if it goes red.

PR #1157, opened 2019-02-08
PP-4453 Implement post govuk pay agreement

SandorArpa:
## WHAT
- implement posting the agreeing user's external id to adminusers
- update nunjuck test

### DIFF
--- a/test/cypress/plugins/stubs.js
+++ b/test/cypress/plugins/stubs.js
@@ -9,6 +9,7 @@ const gatewayAccountFixtures = require('../../fixtures/gateway_account_fixtures'
 const transactionDetailsFixtures = require('../../fixtures/transaction_fixtures')
 const cardFixtures = require('../../fixtures/card_fixtures')
 const serviceFixtures = require('../../fixtures/service_fixtures')
+const goLiveRequestFixtures = require('../../fixtures/go_live_requests_fixture')
 
 /**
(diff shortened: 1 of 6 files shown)


test/cypress/plugins/stubs.js
```diff
@@ -498,5 +498,27 @@ module.exports = {
...
+    return [
+      {
+        predicates: [{
+          equals: {
+            method: 'POST',
+            path: `v1/api/services/${opts.external_id}/govuk-pay-agreement`,
+            headers: {
+              'Accept': 'application/json'
+            },
+            body: { user_external_id: opts.user_external_id }
```
stephencdaly: Even though its really simple, this should use a fixture which is also used by the pact test.

test/fixtures/go_live_requests_fixture.js
```diff
@@ -0,0 +1,30 @@
...
+
+// Global setup
+const pactServices = pactBase({ array: ['service_ids'] })
+
+module.exports = {
+  validPostGovUkPayAgreementRequest: opts => {
+    opts = opts || {}
+
+    const data = {
+      user_external_id: opts.user_external_id || utils.variables.userExternalId
```
stephencdaly: not sure we should be importing the cypress utils here, can just use any value as the default

PR #1170, opened 2019-02-13
PP-4453 Implement Zendesk ticket creation for request to go live

SandorArpa:
## WHAT
- implement ticket creation
- stub Zendesk client
- update tests


app/controllers/request-to-go-live/agreement/post.controller.js
```diff
@@ -21,6 +22,8 @@ const stages = {
   CHOSEN_PSP_EPDQ: goLiveStage.TERMS_AGREED_EPDQ
 }
 
+let userIpAddress
```
sfount: This relies on what is essentially a global variable - the controller is a singleton so this variable assignment will be the same for every request on the Node process. If there are multiple requests, the asynchronous behaviour here could cause it to be set my multiple different requests.

It should not be set at the controller level.

Instead you could keep it within local contexts, potentially passing it down the promise chain. 

Option 1: 
Move the variable into the context of the (req, res) => { } method, this will at least limit the context to one request. 

Option 2: 
Pass the IP address back into the Promise chain

i.e 
```js
postUserIpAddress()
  .then((ipAddress) => { 
    // do something else 
    return { otherDetails, ipAddress }     
  })
  .then() // use ipAddress
```

PR #1175, opened 2019-02-14
PP-4819 web payments activation as digital wallets

jonheslop:
# What
functionality to enable and disable digital wallets (apple pay & google pay)

# How
Note, that this functionality is only available if you have: 
* A username and password set
* Your payment provider is 'worldpay'
In order to test this locally you can change your account in connector to meet this requirements e.g. 
```
update gateway_accounts set payment_provider = 'worldpay' where id = 10;
update gateway_accounts set credentials = '{"username":"foo","password":"bar"}' WHERE id = 10;
```
Then the new functionality should be visible at `/card-types/summary`


app/controllers/digital-wallet/post-disable-apple-pay-controller.js
```diff
@@ -0,0 +1,21 @@
...
+const auth = require('../../services/auth_service')
+const { CORRELATION_HEADER } = require('../../utils/correlation_header.js')
+const connector = new ConnectorClient(process.env.CONNECTOR_URL)
+
+module.exports = (req, res) => {
+  const gatewayAccountId = auth.getCurrentGatewayAccountId(req)
+  const correlationId = req.headers[CORRELATION_HEADER] || ''
+  const disableApplePayBoolean = connector.toggleApplePay(gatewayAccountId, false, correlationId)
+  disableApplePayBoolean.then(() => {
+    req.flash('generic', '<h2>Apple Pay successfully disabled.</h2>')
```
sfount: `flash`ing markup seems like a bad precedent to set -- in general the controller should not be concerned with how anything is rendered. 

Could this be picked up and formatted by the template?

app/controllers/digital-wallet/post-disable-google-pay-controller.js
```diff
@@ -0,0 +1,22 @@
...
+const paths = require('../../paths')
+const { ConnectorClient } = require('../../services/clients/connector_client')
+const auth = require('../../services/auth_service')
+const { CORRELATION_HEADER } = require('../../utils/correlation_header.js')
+const connector = new ConnectorClient(process.env.CONNECTOR_URL)
+
+module.exports = (req, res) => {
+  const gatewayAccountId = auth.getCurrentGatewayAccountId(req)
+  const correlationId = req.headers[CORRELATION_HEADER] || ''
+  const enableGooglePayBoolean = connector.toggleGooglePay(gatewayAccountId, false, correlationId)
```
sfount: This is confusing to read, if we are not planning to do anything with the value that is eventually propagated there should be no reason to assign this

```suggestion
connector.toggleGooglePay(gatewayAccountId, false, correlationId)
  .then(() => { 
```

app/controllers/digital-wallet/post-disable-google-pay-controller.js
```diff
@@ -0,0 +1,22 @@
...
+
+// Local dependencies
+const paths = require('../../paths')
+const { ConnectorClient } = require('../../services/clients/connector_client')
+const auth = require('../../services/auth_service')
+const { CORRELATION_HEADER } = require('../../utils/correlation_header.js')
+const connector = new ConnectorClient(process.env.CONNECTOR_URL)
+
+module.exports = (req, res) => {
+  const gatewayAccountId = auth.getCurrentGatewayAccountId(req)
```
sfount: Is this value not assigned to the `req` object by virtue of the `get_gateway_account` middleware? I haven't looked into this in depth just wondering why there are two potential places this could be parsed.

app/controllers/digital-wallet/post-disable-apple-pay-controller.js
gidsg: I agree this is a bit irregular but we do it already _all over the place_ e.g.  
https://github.com/alphagov/pay-selfservice/blob/b78d02a609896bb1208d256428eefd4fa2f9d203/app/controllers/credentials_controller.js#L100
can you think of a better way to do this @jonheslop

app/controllers/digital-wallet/post-disable-google-pay-controller.js
gidsg: @sfount I believe this is derived from the `req` object, this function just wraps the logic to fetch it: https://github.com/alphagov/pay-selfservice/blob/ceb9e51484483776a0918757e294f639f29eff74/app/services/auth_service.js#L134-L157

app/controllers/digital-wallet/post-disable-apple-pay-controller.js
jonheslop: Yeah it’s bad, but maybe I can come up with something better in a separate PR after this is merged

something like 
`req.flash('genericError', {heading:"Please enter a valid username",body:""})`

jonheslop: Would also mean we wouldn’t have to have `| safe` in the flash.njk template

gidsg: @jonheslop 👍 I would rather we raise a tech debt card for this and fix in a separate PR as this pattern is not restricted to this story.

gidsg: ...OK I made a tech debt card now: https://trello.com/c/ro30NARU/50-selfservice-uses-in-line-html-in-controllers-to-render-error-messages

sfount: 👍

app/controllers/digital-wallet/post-disable-google-pay-controller.js
sfount: Yeah I think you're correct. My comment was just that this controller sits behind middleware that (I think?) ensures there's an `account` object passed through req, so potentially you could: 

* remove the dependancy / required `auth` import 
* use the account that is passed through the middleware

```suggestion
  const gatewayAccountId = req.account.id
```

sfount:
(➕ I think the git history for this PR could be rebased/ cleaned up to reflect the significant changes)

PR #1176, opened 2019-02-14
PP-4448 Modify logic for displaying links on dashboard

stephencdaly:
Move logic for deciding which links will be displayed from the Nunjucks template to the controller. This allows us to know how many links will be displayed, and so format them correctly into either 2 or 3 columns with the correct borders on the links. This is required as the logic gets a bit more complicated when we add in the "request to go live" link

Move the decision for whether to display a bottom border on the links from the template, where we were conditionally adding a css class. We now have a css rule using the nth-child selector to decide whether a link box should have a border.

Add Cypress tests for displaying the links.

### DIFF
--- a/app/controllers/dashboard/dashboard-activity-controller.js
+++ b/app/controllers/dashboard/dashboard-activity-controller.js
@@ -13,12 +13,41 @@ const CORRELATION_HEADER = require('../../utils/correlation_header').CORRELATION
 const ConnectorClient = require('../../services/clients/connector_client').ConnectorClient
 const { isADirectDebitAccount } = require('../../services/clients/direct_debit_connector_client.js')
 const auth = require('../../services/auth_service.js')
-const connectorClient = () => new ConnectorClient(process.env.CONNECTOR_URL)
 const { datetime } = require('@govuk-pay/pay-js-commons').nunjucksFilters
+
+const connectorClient = () => new ConnectorClient(process.env.CONNECTOR_URL)
 const getTimespanDays = (fromDateTime, toDateTime) => moment(toDateTime).diff(moment(fromDateTime), 'days')
 
 // Constants
 const clsXrayConfig = require('../../../config/xray-cls')
+
+const links = {
+  manageService: 0,
+  demoPayment: 1,
+  testPaymentLink: 2,
+  directDebitPaymentFlow: 3,
+  paymentLinks: 4,
+  goLive: 5
+}
+
+const getLinksToDisplay = function getLinksToDisplay (service, account) {
+  const linksToDisplay = [links.manageService]
+
+  if (account.payment_provider === 'sandbox') {
+    linksToDisplay.push(links.demoPayment)
+    linksToDisplay.push(links.testPaymentLink)
+  } else if (account.paymentMethod === 'direct debit') {
+    linksToDisplay.push(links.directDebitPaymentFlow)
+  } else {
+    linksToDisplay.push(links.paymentLinks)
+  }
+
+  if (account.type === 'test') {
+    linksToDisplay.push(links.goLive)
+  }
+
+  return linksToDisplay
+}
 
 module.exports = (req, res) => {
   const gatewayAccountId = auth.getCurrentGatewayAccountId((req))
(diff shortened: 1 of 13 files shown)


app/controllers/dashboard/dashboard-activity-controller.js
```diff
@@ -13,13 +13,59 @@ const CORRELATION_HEADER = require('../../utils/correlation_header').CORRELATION
...
 const { datetime } = require('@govuk-pay/pay-js-commons').nunjucksFilters
+const goLiveStage = require('../../models/go-live-stage')
+
+const connectorClient = () => new ConnectorClient(process.env.CONNECTOR_URL)
 const getTimespanDays = (fromDateTime, toDateTime) => moment(toDateTime).diff(moment(fromDateTime), 'days')
 
 // Constants
 const clsXrayConfig = require('../../../config/xray-cls')
 
+const getLinksToDisplay = (service, account) => {
```
sfount:
```suggestion
const getLinksToDisplay = function getLinksToDisplay (service, account) => {
```

It would be useful to name these instead of using anonymous functions, if an exception is thrown the name will then be used in the stack trace in favour of `anonymous`.

app/controllers/dashboard/dashboard-activity-controller.js
```diff
@@ -13,13 +13,59 @@ const CORRELATION_HEADER = require('../../utils/correlation_header').CORRELATION
...
+  return links
+}
+
+/**
+ * Gets an array of the css classes to be applied to the link boxes on the dashboard in the order they are displayed in.
+ * @param numberOfLinks the number of links that will be displayed
+ * @returns {Array}
+ */
+const getLinkBoxClasses = (numberOfLinks) => {
+  const columnClass = numberOfLinks % 3 === 0 ? 'flex-grid--column-third' : 'flex-grid--column-half'
```
sfount: In general it would be good if this logic could be moved over to the template, [nunjucks](https://mozilla.github.io/nunjucks/) lets you do a fair amount of processing. 

I don't have a suggestion here and I don't think it should block the PR -- having a clean separation between the controller figuring out what to display and the templates deciding how to display it is useful for reasoning about how the service works.

test/cypress/integration/dashboard/dashboard_links_spec.js
```diff
@@ -0,0 +1,199 @@
...
+  const userExternalId = 'cd0fa54cf3b7408a80ae2f1b93e7c16e'
+
+  describe('card gateway account', () => {
+    const gatewayAccountId = 42
+
+    beforeEach(() => {
+      cy.setEncryptedCookies(userExternalId, gatewayAccountId)
+    })
+
+    it('should display 3 links for a live sandbox account', () => {
```
sfount: Great E2E tests confirming the functionality throughout 👍

app/controllers/dashboard/dashboard-activity-controller.js
stephencdaly: Yes I think that's sensible. Doing this here doesn't feel quite right, as the controller probably shouldn't be specifying the CSS classes, right?
Do you know if I would be able to do it in a similar way in Nunjucks? I think the most difficult part is deciding whether or not to display the border, deciding which column class to use would be straightforward enough in Nunjucks.

app/controllers/dashboard/dashboard-activity-controller.js
stephencdaly: Thanks, I've made this change

app/controllers/dashboard/dashboard-activity-controller.js
stephencdaly: I have got rid of using the bottom-border css class, and we're now using an nth-child selector in the css to add in the grey line separators to the links. Thanks for all the suggestions and help.

app/views/dashboard/_links.njk
```diff
@@ -1,21 +1,32 @@
...
+{% else %}
+  {% set columnClass = "flex-grid--column-half" %}
+{% endif %}
+
+<div class="govuk-grid-column-full flex-grid links-container" data-click-events data-click-category="Dashboard" data-click-action="First steps link clicked">
+  
   <div class="flex-grid--row">
-    {% if isSandbox %}
-    <article class="flex-grid--column-half links__box border-bottom">
+    {% if links.demoPayment in linksToDisplay %}
```
sfount: Much simpler to read 👍 Great work!

PR #1181, opened 2019-02-15
PP-4759 international phone number support backward compatibility - part 2

DanailMinchev:
## WHAT

Currently pay-adminusers saves and returns international telephone number format starting with `+44` (UK example).
The current validation in selfservice still expects numbers only telephone numbers.
This commit modifies the value of the input to be reformatted in numbers only international telephone number.
Once the validation of the selfservice can handle any format we need to remove this formatting.

### DIFF
--- a/app/views/user_registration/re_verify_phone.njk
+++ b/app/views/user_registration/re_verify_phone.njk
@@ -24,7 +24,7 @@
               text: "Mobile number",
               classes: "govuk-label--s"
           },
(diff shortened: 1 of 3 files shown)


DanailMinchev:
Looks like this is not needed as we save raw telephone numbers for new users

DanailMinchev:
Actually it is needed as we use session to store temporary the user input, but we still save the `+44` format in the database

PR #1188, opened 2019-02-18
PP-4846 Go live task list: show next steps box at right point

alexbishop1:
Show the next steps box saying we are creating the live service when the current go live stage is `AGREED_TERMS_STRIPE`, `AGREED_TERMS_WORLDPAY`, `AGREED_TERMS_SMARTPAY` or `AGREED_TERMS_EPDQ` and not when it’s `LIVE`.

### DIFF
--- a/app/controllers/request-to-go-live/index/get.controller.js
+++ b/app/controllers/request-to-go-live/index/get.controller.js
@@ -1,16 +1,82 @@
 'use strict'
 
+// NPM dependencies
+const lodash = require('lodash')
+
 // Local dependencies
 const goLiveStage = require('../../../models/go-live-stage')
 const response = require('../../../utils/response')
 
+const {
+  NOT_STARTED,
+  ENTERED_ORGANISATION_NAME,
+  CHOSEN_PSP_STRIPE,
+  CHOSEN_PSP_WORLDPAY,
+  CHOSEN_PSP_SMARTPAY,
+  CHOSEN_PSP_EPDQ,
+  TERMS_AGREED_STRIPE,
+  TERMS_AGREED_WORLDPAY,
+  TERMS_AGREED_SMARTPAY,
+  TERMS_AGREED_EPDQ,
+  LIVE,
+  DENIED
+} = goLiveStage
+
+const live = [
+  LIVE
+]
+
+const agreedToTerms = [
+  TERMS_AGREED_STRIPE,
+  TERMS_AGREED_WORLDPAY,
+  TERMS_AGREED_SMARTPAY,
+  TERMS_AGREED_EPDQ
+].concat(live)
+
+const chosenHowToProcessPayments = [
+  CHOSEN_PSP_STRIPE,
+  CHOSEN_PSP_WORLDPAY,
+  CHOSEN_PSP_SMARTPAY,
+  CHOSEN_PSP_EPDQ
+].concat(agreedToTerms)
+
+const enteredOrganisationName = [
+  ENTERED_ORGANISATION_NAME
+].concat(chosenHowToProcessPayments)
+
+const startedButStillStepsToComplete = [
+  ENTERED_ORGANISATION_NAME,
+  CHOSEN_PSP_STRIPE,
+  CHOSEN_PSP_WORLDPAY,
+  CHOSEN_PSP_SMARTPAY,
+  CHOSEN_PSP_EPDQ
+]
+
+const showNextSteps = [
+  TERMS_AGREED_STRIPE,
+  TERMS_AGREED_WORLDPAY,
+  TERMS_AGREED_SMARTPAY,
+  TERMS_AGREED_EPDQ
+]
+
 module.exports = (req, res) => {
-  return response.response(
-    req,
-    res,
-    'request-to-go-live/index',
-    {
-      goLiveStage
-    }
-  )
+  const currentGoLiveStage = lodash.get(req, 'service.currentGoLiveStage', '')
+
+  let pageData = lodash.get(req, 'session.pageData.requestToGoLive.index')
+  if (pageData) {
+    delete req.session.pageData.requestToGoLive.index
+  }
+
+  pageData = {
+    notStarted: currentGoLiveStage === NOT_STARTED,
+    enteredOrganisationName: enteredOrganisationName.includes(currentGoLiveStage),
+    chosenHowToProcessPayments: chosenHowToProcessPayments.includes(currentGoLiveStage),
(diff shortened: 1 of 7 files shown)


app/controllers/request-to-go-live/index/get.controller.js
```diff
@@ -1,16 +1,82 @@
...
+]
+
+const agreedToTerms = [
+  TERMS_AGREED_STRIPE,
+  TERMS_AGREED_WORLDPAY,
+  TERMS_AGREED_SMARTPAY,
+  TERMS_AGREED_EPDQ
+].concat(live)
+
+const chosenHowToProcessPayments = [
```
sfount: (This certainly shouldn't be blocking - it depends on how often this pattern will continue in the future. tl;dr define constants and groups once)

It seems like we're pulling and categorising these constant out of this object map in a number of different controllers. Would it make sense to move this logic over to the helper object/ model? 

Potential pros: 
* controllers don't have to duplicate code 
* controllers are simple and easy to read (although this code is very clear!) 

Something like: 
```js
const goLiveStage = require('goLiveStage')

const goLiveIndexHandler = function goLiveIndexHandler (req, res, next) { 
  const { currentGoLiveStage } = req.service
  const goLiveProcessStatus = goLiveStage.status(currentGoLiveStage)

  // ensure cookies set in previous session aren't propagated
  delete req.session.pageData.requestToGoLive.index

  return response(req, res, 'request-to-go-live/index', pageData)
}

module.exports = goLiveIndexHandler
```

Or  a middle ground would be to do the categorising only once in the helper/ model and then use those for logic in the controllers: 

```js
const { agreedToTermsStates, sessionIsAgreeingToTerms } = require('goLiveStage') 

// check if my status is in the agreed terms
agreedToTermsStates.includes(currentGoLiveStage) 

// or 
sessionIsAgreeingToTerms(currentGoLiveStage)
```

Files that also do this: 
https://github.com/alphagov/pay-selfservice/blob/master/app/controllers/request-to-go-live/choose-how-to-process-payments/post.controller.js#L16
https://github.com/alphagov/pay-selfservice/blob/master/app/controllers/request-to-go-live/agreement/get.controller.js#L12

app/views/request-to-go-live/index.njk
```diff
@@ -7,15 +7,15 @@
 {% endblock %}
 
 {% block mainContent %}
-  {% if currentService.currentGoLiveStage === goLiveStage.LIVE %}
```
DanailMinchev: Moving the logic to the controller looks nice 👍

app/controllers/request-to-go-live/index/get.controller.js
```diff
@@ -1,16 +1,82 @@
...
+const live = [
+  LIVE
+]
+
+const agreedToTerms = [
+  TERMS_AGREED_STRIPE,
+  TERMS_AGREED_WORLDPAY,
+  TERMS_AGREED_SMARTPAY,
+  TERMS_AGREED_EPDQ
+].concat(live)
```
DanailMinchev: You can use `Destructuring assignment`, for example:
```
const agreedToTerms = [
  TERMS_AGREED_STRIPE,
  TERMS_AGREED_WORLDPAY,
  TERMS_AGREED_SMARTPAY,
  TERMS_AGREED_EPDQ,
  ...live
]
```

We use it in https://github.com/alphagov/pay-selfservice/blob/884eaf4eaf9b1b1796b65911f288110cef1f0959/app/routes.js#L174 and https://github.com/alphagov/pay-selfservice/blob/884eaf4eaf9b1b1796b65911f288110cef1f0959/app/utils/json_to_csv.js#L43 and I think it's the latest syntax which we discussed

app/controllers/request-to-go-live/index/get.controller.js
```diff
@@ -1,16 +1,82 @@
...
+
+const chosenHowToProcessPayments = [
+  CHOSEN_PSP_STRIPE,
+  CHOSEN_PSP_WORLDPAY,
+  CHOSEN_PSP_SMARTPAY,
+  CHOSEN_PSP_EPDQ
+].concat(agreedToTerms)
```
DanailMinchev: same for other `concat`s

app/views/policy/document-downloads/contract_for_non_crown_bodies.njk
```diff
@@ -5,7 +5,7 @@
 {% endblock %}
 
 {% block policyDescription %}
-  If you're a non-crown body, the contract terms apply when you use GOV.UK Pay.
+  If you’re a non-crown body, the contract terms apply when you use GOV.UK Pay.
```
DanailMinchev: Sorry, a bit picky: this can be in a different commit, thanks for fixing that according to latest content rules 👍

app/controllers/request-to-go-live/index/get.controller.js
alexbishop1: Ooh, I like that!

app/views/policy/document-downloads/contract_for_non_crown_bodies.njk
alexbishop1: Yeah, it’s unrelated but I thought I’d sneak it in because it’s tiny

test/cypress/integration/request-to-go-live/index_spec.js
```diff
@@ -60,14 +60,15 @@ describe('Request to go live: index', () => {
       cy.visit(requestToGoLivePageUrl)
 
       cy.get('h1').should('contain', 'Request to go live')
+      cy.get('h1 + p').should('contain', 'Complete these steps to start taking real payments')
 
-      cy.get('#request-to-go-live-step-1 > h3').should('exist')
```
DanailMinchev: In the original code it is not obvious here why we assert on existence and not on content, the idea was not to depend on text/content of each step in case there are minor content changes / or major which should not break the test. That is why I prefer to rely on ids not on content.
But, after discussion with @alexbishop1 I understand that it is not clear why we have this assertions and which step is what.
We discussed to change the ids to be more descriptive or to leave the content as it is highly coupled with implementation (as steps).

Thanks

alexbishop1: Just pushed a commit that asserts less on text and makes the IDs more descriptive

app/controllers/request-to-go-live/index/get.controller.js
alexbishop1: Yes, something to consider for the future

PR #1196, opened 2019-02-20
PP-4462 Add uk-modulus-checking package

DanailMinchev:
## WHAT

Add `uk-modulus-checking` package which will be used to validate UK bank account details: sort code and account number.
It is using VocaLink modulus check - https://www.vocalink.com/customer-support/modulus-checking/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3038,773 +3038,6 @@
         "upath": "^1.0.5"
       }
     },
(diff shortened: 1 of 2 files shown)


DanailMinchev:
We discussed that probably we should do only "basic" checks and not rely on 3rd party package

PR #1198, opened 2019-02-21
PP-4462 Add client methods to get and update Stripe setup flags

stephencdaly:
Add method to connector_client to call the GET stripe account setup endpoint. The response of this request is mapped into a new StripeAccountSetup model.
Add method to call the PATCH stripe account setup endpoint to set a flag for an account setup task being completed.

Add pact tests for these interactions.

### DIFF
--- a/app/services/clients/connector_client.js
+++ b/app/services/clients/connector_client.js
@@ -8,10 +8,12 @@ const logger = require('winston')
 const querystring = require('querystring')
 
 // Local dependencies
-const baseClient = require('./old_base_client')
+const oldBaseClient = require('./old_base_client')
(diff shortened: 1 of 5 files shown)


app/services/clients/connector_client.js
```diff
@@ -169,7 +174,7 @@ ConnectorClient.prototype = {
       url: url
     })
 
-    baseClient.get(url, {correlationId: params.correlationId}, function (error, response, body) {
+    oldBaseClient.get(url, {correlationId: params.correlationId}, function (error, response, body) {
```
DanailMinchev: Thanks for renaming 👍

PR #1202, opened 2019-02-22
PP-4462 Add Stripe client

stephencdaly:
Add a stripe client to make calls to the stripe API through the stripe node package.

Add a model for the payload sent to stripe to add bank account details to a stripe account.

with @DanailMinchev


app/services/clients/stripe/stripeBankAccount.model.js
```diff
@@ -0,0 +1,42 @@
...
+  bank_account_number: Joi.string().required()
+}
+
+class StripeBankAccount {
+  constructor (body) {
+    const params = Object.assign({}, body)
+    const { error, value: model } = Joi.validate(params, schema, { allowUnknown: true, stripUnknown: true })
+
+    if (error) {
+      throw new ValidationError(`StripeBankAccount ${error.details[0].message}`)
```
sfount: Good idea to use validation and explicitly throwing errors 👍 I'm not sure if we need to create an extended `ValidationError` in favour of just throwing an `Error` as there is nowhere right now that will be looking for error types. I suppose it shouldn't hurt anything. 

Questions following this will be: 
* How will these errors be caught and handled by controllers using this client

app/services/clients/stripe/stripe_client.js
```diff
@@ -0,0 +1,41 @@
...
+const StripeBankAccount = require('./stripeBankAccount.model')
+
+// Constants
+const STRIPE_LIVE_API_KEY = process.env.STRIPE_LIVE_API_KEY
+const STRIPE_TEST_API_KEY = process.env.STRIPE_TEST_API_KEY
+const STRIPE_HOST = process.env.STRIPE_HOST
+const STRIPE_PORT = process.env.STRIPE_PORT
+
+const getStripe = function getStripe (gatewayAccount) {
+  const apiKey = (gatewayAccount.type === 'test') ? STRIPE_TEST_API_KEY : STRIPE_LIVE_API_KEY
```
sfount: I don't think it makes sense for this to have access to both the `TEST` and `LIVE` API keys at once -- it seems this process moves a gateway account from being `test` to `live` which isn't quite understood by this conditional (i.e there won't exist Stripe test accounts for a test Gateway Accounts)

Potentially we shouldn't include this dual logic until we have a clear use case (instead favouring just `STRIPE_API_KEY` being set by the environment with test or production keys as required).

app/services/clients/stripe/stripe_client.js
```diff
@@ -0,0 +1,41 @@
...
+    stripe.setPort(STRIPE_PORT)
+  }
+
+  return stripe
+}
+
+module.exports = {
+  updateBankAccount: function (gatewayAccount, stripeAccountId, body) {
+    const bankAccount = new StripeBankAccount(body)
+    const stripe = getStripe(gatewayAccount)
```
sfount: At the moment this initialises a new Stripe object per request to update a bank account, self service doesn't have the same scale requirements as other using facing services however this means that for 10 requests we make 10 (relatively large) objects, for 100 we make 100. 

My understanding is that the Stripe SDK returns a fairly large object that is able to make requests based on its configuration -- that shouldn't change after the server has started. 

If we have to connect to both Live and Test accounts we would need two calls to `Stripe`, otherwise we may only need one. 

For example: 
```js
const stripe = require('stripe')(STRIPE_API_KEY)

stripe.setHttpAgent() //...
stripe.setHost() //...
stripe.setPort() //...

const updateBankAccount = function updateBankAccount () { 
  return stripe.accounts.update() //...
}

module.exports = { updateBankAccount }
```

stephencdaly: Makes sense, thanks 👍

app/services/clients/stripe/stripe_client.js
stephencdaly: Agreed, we'll change this

app/services/clients/stripe/stripeBankAccount.model.js
DanailMinchev: As discussed we will use `Error`, thanks

PR #1204, opened 2019-02-25
PP-4463 Adds Template for Responsible person

cobainc0:
## WHAT
Nunjucks template for the main nominate a responsible person form. Not hooked up to anything yet.
with: @alexbishop1

### DIFF
--- a/app/views/stripe-setup/responsible-person/index.njk
+++ b/app/views/stripe-setup/responsible-person/index.njk
@@ -1,0 +1,156 @@
+{% extends "../../layout.njk" %}
+{% from "components/input/macro.njk" import govukInput %}
+{% from "components/button/macro.njk" import govukButton %}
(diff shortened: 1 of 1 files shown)


app/views/stripe-setup/responsible-person/index.njk
```diff
@@ -0,0 +1,156 @@
+{% extends "../../layout.njk" %}
+{% from "components/input/macro.njk" import govukInput %}
+{% from "components/button/macro.njk" import govukButton %}
+{% from "components/date-input/macro.njk" import govukDateInput %}
+{% from "components/error-summary/macro.njk" import govukErrorSummary %}
```
sfount: Doesn't look like you need `govukErrorSummary` here

alexbishop1: Not yet but we will soon 😀

PR #1211, opened 2019-02-26
PP-4462 Add bank details validation

stephencdaly:
Add validation checks to be used by both the client-side javascript validation and the server side validation to validate the account number and sort code.

Add bank-details-validation.js which will be used by the controller to perform the server-side validation.

with @DanailMinchev

### DIFF
--- a/app/browsered/field-validation-checks.js
+++ b/app/browsered/field-validation-checks.js
@@ -85,7 +85,7 @@ exports.isNaxsiSafe = function (value) {
 }
 
 exports.isNotAccountNumber = value => {
-  if (NUMBERS_ONLY.test(value) && (value.length >= 6 && value.length <= 8)) {
+  if (/^[0-9]{6,8}$/.test(value)) {
     return false
   } else {
     return validationErrors.invalidBankAccountNumber
@@ -93,7 +93,7 @@ exports.isNotAccountNumber = value => {
 }
 
 exports.isNotSortCode = value => {
-  if (/^[ -]*(?:[0-9][ -]*){6}$/.test(value)) {
+  if (/^(?:[ -]*[0-9][ -]*){6}$/.test(value)) {
     return false
   } else {
     return validationErrors.invalidSortCode
--- a/app/browsered/field-validation-checks.js
+++ b/app/browsered/field-validation-checks.js
@@ -16,7 +16,9 @@ const validationErrors = {
   isAboveMaxAmount: `Choose an amount under £${MAX_AMOUNT.toLocaleString()}`,
   isPasswordLessThanTenChars: `Choose a Password of 10 characters or longer`,
   isGreaterThanMaxLengthChars: `The text is too long`,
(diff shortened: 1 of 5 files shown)


test/unit/browsered/field_validation_checks_test.js
```diff
@@ -41,4 +47,85 @@ describe('field validation checks', () => {
       expect(isFieldGreaterThanMaxLengthChars('123456ABC', 'que')).to.equal(false)
     })
   })
+  describe('isNotValidAccountNumber', () => {
+    it('should validate successfully for 8 digits', () => {
+      const accountNumber = '00012345'
```
alexbishop1: As somebody with a bank account number that begins with a zero and who has been bitten by shoddy websites before, I thank you! 🙇🏼‍♂️

test/unit/browsered/field_validation_checks_test.js
```diff
@@ -41,4 +47,85 @@ describe('field validation checks', () => {
...
+      const sortCode = '10 88 00'
+
+      expect(isNotSortCode(sortCode)).to.be.false // eslint-disable-line
+    })
+
+    it('should validate successfully for 6 digits with mix of dashes and spaces', () => {
+      const sortCode = '10-88 00'
+
+      expect(isNotSortCode(sortCode)).to.be.false // eslint-disable-line
+    })
```
alexbishop1: No tests for doing something like `1 0 88 00` even though our regex would appear to allow it — do we want that?

app/browsered/field-validation-checks.js
```diff
@@ -81,3 +83,19 @@ exports.isNaxsiSafe = function (value) {
     return false
   }
 }
+
+exports.isNotAccountNumber = value => {
```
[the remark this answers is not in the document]
DanailMinchev: Yes, we initially named it `isAccountNumber`, but then we discussed that we should fix all the validation, for example `isEmpty` returns `false` if the string is empty:

https://github.com/alphagov/pay-selfservice/blob/8f562914984c8da646e884b7af720bb85337d7b4/app/browsered/field-validation-checks.js#L22

same for:
https://github.com/alphagov/pay-selfservice/blob/8f562914984c8da646e884b7af720bb85337d7b4/app/browsered/field-validation-checks.js#L30
https://github.com/alphagov/pay-selfservice/blob/8f562914984c8da646e884b7af720bb85337d7b4/app/browsered/field-validation-checks.js#L38
etc

And we decided to be consistent with current ones (in terms of returning false and string), but at least name the function correctly.

We should refactor the validation as in Direct Debit frontend if the time permits

test/unit/browsered/field_validation_checks_test.js
DanailMinchev: We allow `10-88 00` case because we are going to do some normalisation, so we decided that `1 0 88 00` is valid as well and added the test, thanks

PR #1222, opened 2019-02-28
PP-4462 Stripe setup middlewares

DanailMinchev:
## WHAT

This PR adds two middelwares which can be used for Stripe setup page validations:

- Restrict to live Stripe account middleware
- Check bank details not submitted middleware

with @stephencdaly

### DIFF
--- a/app/middleware/stripe-setup/check-bank-details-not-submitted.js
+++ b/app/middleware/stripe-setup/check-bank-details-not-submitted.js
@@ -1,0 +1,25 @@
+'use strict'
+
+// Local dependencies
+const ConnectorClient = require('../../services/clients/connector_client').ConnectorClient
+const connector = new ConnectorClient(process.env.CONNECTOR_URL)
+const { renderErrorView } = require('../../utils/response')
+const paths = require('../../paths')
+
+module.exports = (req, res, next) => {
+  if (!req.account) {
+    renderErrorView(req, res, 'Internal server error')
+    return
+  }
+
+  connector.getStripeAccountSetup(req.account.gateway_account_id, req.correlationId).then(stripeSetupResponse => {
+    if (stripeSetupResponse.bankAccount) {
+      req.flash('genericError', 'Bank details flag already set')
+      res.redirect(303, paths.dashboard.index)
+    } else {
+      next()
+    }
+  }).catch(() => {
+    renderErrorView(req, res, 'Please try again or contact support team')
+  })
+}
--- a/app/middleware/stripe-setup/check-bank-details-not-submitted.js
+++ b/app/middleware/stripe-setup/check-bank-details-not-submitted.js
@@ -1,12 +1,12 @@
 'use strict'
 
 // Local dependencies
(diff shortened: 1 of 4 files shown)


app/middleware/stripe-setup/check-bank-details-not-submitted.js
```diff
@@ -0,0 +1,25 @@
...
+
+// Local dependencies
+const ConnectorClient = require('../../services/clients/connector_client').ConnectorClient
+const connector = new ConnectorClient(process.env.CONNECTOR_URL)
+const { renderErrorView } = require('../../utils/response')
+const paths = require('../../paths')
+
+module.exports = (req, res, next) => {
+  if (!req.account) {
+    renderErrorView(req, res, 'Internal server error')
```
sfount: As this is just a generic error can it not be handled by middleware further down the stack, something like 

```js
next(new Error('Invalid Stripe account provided')) 
```

Throwing generic errors within middleware might be hard to track down in the future.

app/middleware/stripe-setup/check-bank-details-not-submitted.js
```diff
@@ -0,0 +1,25 @@
+'use strict'
+
+// Local dependencies
+const ConnectorClient = require('../../services/clients/connector_client').ConnectorClient
```
sfount:
```suggestion
const { ConnectorClient } = require('../../services/clients/connector_client')
```

app/middleware/stripe-setup/restrict-to-live-stripe-account.js
```diff
@@ -0,0 +1,15 @@
+'use strict'
+
+module.exports = (req, res, next) => {
```
sfount: This middleware could do with potentially a few names/ comments to make this 
* easier to read 
* provide more complete information during a stack trace

Suggestions: 
```js
const restrictRequestsToLiveAccounts = function restrictRequestsToLiveAccounts (req, res, next) { 
  const requestHasValidLiveAccount = req.account && req.account.type // ... 

  if (requestHasValidLiveAccount) { 
    next()
  } else { 
    // request is missing information - a 404 Not Found should be displayed? 
  }
}
module.exports = restrictRequestsToLiveAccounts

app/middleware/stripe-setup/check-bank-details-not-submitted.js
DanailMinchev: Yes, I agree, as discussed we do not have `Error` handling middleware and we use https://github.com/alphagov/pay-selfservice/blob/a90957ad2045b3d124cab59e6572608927f4df89/app/utils/response.js#L49 in a lot of places, so we decided to use the current way and refactor generally at some point

app/middleware/stripe-setup/restrict-to-live-stripe-account.js
DanailMinchev: Thanks, I have added `function` to both of the middlewares
and comment explaining why the error should be `404`

sfount:
Reviewed under :fire:

PR #1226, opened 2019-03-01
PP-4564: Address negation always true lint flag

sfount:
Relates to https://github.com/alphagov/pay-selfservice/pull/1041


jankowiakmaria:
Discussed with @sfount.
The change is not correct (but can be confusing therefore it should be split into more readable methods) - closing.

PR #1228, opened 2019-03-01
PP-4463 Server-side validation for responsible person page

alexbishop1:
Except for date of birth validation, which is coming later.

Update max lengths to match what we think Stripe accepts.

with @cobainc0

### DIFF
--- a/app/controllers/stripe-setup/responsible-person/post.controller.js
+++ b/app/controllers/stripe-setup/responsible-person/post.controller.js
@@ -2,27 +2,98 @@
 
 // NPM dependencies
 const lodash = require('lodash')
+
 // Local dependencies
 const paths = require('../../../paths')
 const response = require('../../../utils/response')
+const {
+  validateMandatoryField, validateOptionalField, validatePostcode
+} = require('./responsible-person-validations')
+
+const FIRST_NAME_FIELD = 'first-name'
+const LAST_NAME_FIELD = 'last-name'
+const HOME_ADDRESS_LINE1_FIELD = 'home-address-line-1'
+const HOME_ADDRESS_LINE2_FIELD = 'home-address-line-2'
+const HOME_ADDRESS_CITY_FIELD = 'home-address-city'
+const HOME_ADDRESS_POSTCODE_FIELD = 'home-address-postcode'
+const DOB_DAY_FIELD = 'dob-day'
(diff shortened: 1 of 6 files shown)


app/controllers/stripe-setup/responsible-person/responsible-person-validations.test.js
```diff
@@ -0,0 +1,71 @@
+'use strict'
```
sfount: Do we name any other tests `.test.js`? I haven't seen this before - I think normally `spec` is used.

app/controllers/stripe-setup/responsible-person/post.controller.js
```diff
@@ -2,27 +2,98 @@
...
+    maxLength: 100
+  },
+  {
+    field: HOME_ADDRESS_POSTCODE_FIELD,
+    validator: validatePostcode
+  }
+]
 
 module.exports = (req, res) => {
+  const errors = validationRules.reduce((errors, validationRule) => {
```
sfount: 👍

app/controllers/stripe-setup/responsible-person/post.controller.js
```diff
@@ -2,27 +2,98 @@
...
+  } else if (lodash.get(req.body, 'answers-need-changing') === 'true') {
     return response.response(req, res, 'stripe-setup/responsible-person/index', pageData)
+  } else {
+    return response.response(req, res, 'stripe-setup/responsible-person/check-your-answers', pageData)
   }
-  return response.response(req, res, 'stripe-setup/responsible-person/check-your-answers', pageData)
+}
+
+const validate = (req, fieldName, fieldValidator, maxLength) => {
+  const field = getFieldValue(req, fieldName)
```
sfount: Is there a need for this method? It looks like it directly maps to `lodash.get`

app/controllers/stripe-setup/responsible-person/responsible-person-validations.test.js
alexbishop1: We have … a few different conventions, including `.test.js`

app/controllers/stripe-setup/responsible-person/post.controller.js
alexbishop1: We’ll remove this in a future PR

PR #1237, opened 2019-03-05
PP-4910: Add Wallet Type to single transaction page

gidsg:
## WHAT
Add Wallet Type to single transaction page

## HOW 
Charge with Wallet Type should appear on single transaction page

### DIFF
--- a/test/cypress/integration/transactions/transaction_details_spec.js
+++ b/test/cypress/integration/transactions/transaction_details_spec.js
@@ -317,6 +317,13 @@ describe('Transactions details page', () => {
     chargeDetails.charge.wallet_type = 'APPLE_PAY'
     cy.task('setupStubs', getStubs(chargeDetails))
     cy.visit(`${transactionsUrl}/${chargeDetails.charge.charge_id}`)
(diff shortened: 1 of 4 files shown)


test/cypress/integration/transactions/transaction_details_spec.js
```diff
@@ -311,4 +311,12 @@ describe('Transactions details page', () => {
...
     })
   })
+
+  it('should display Wallet Type where available', () => {
+    const chargeDetails = defaultChargeDetails()
+    chargeDetails.charge.wallet_type = 'APPLE_PAY'
+    cy.task('setupStubs', getStubs(chargeDetails))
+    cy.visit(`${transactionsUrl}/${chargeDetails.charge.charge_id}`)
+    cy.get('th').contains('Wallet type').siblings().first().contains('Apple Pay')
+  })
```
oswaldquek: can we have a test where wallet type is not displayed when it shouldn't be?

gidsg: @oswaldquek OK I added a test for this now.

PR #1243, opened 2019-03-06
PP-4463 Server-side responsible person date of birth validation

alexbishop1:
Add server-side validation for the date of the birth on the Stripe responsible person page. Requires a date of birth in past and uses the error messages described in the GOV.UK Design System at https://design-system.service.gov.uk/components/date-input/.

### DIFF
--- a/app/controllers/stripe-setup/responsible-person/post.controller.js
+++ b/app/controllers/stripe-setup/responsible-person/post.controller.js
@@ -7,7 +7,7 @@ const lodash = require('lodash')
 const paths = require('../../../paths')
 const response = require('../../../utils/response')
 const {
-  validateMandatoryField, validateOptionalField, validatePostcode
+  validateMandatoryField, validateOptionalField, validatePostcode, validateDateOfBirth
 } = require('./responsible-person-validations')
 
 const FIRST_NAME_FIELD = 'first-name'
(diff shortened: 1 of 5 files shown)


app/controllers/stripe-setup/responsible-person/post.controller.js
```diff
@@ -86,14 +91,21 @@ module.exports = (req, res) => {
...
   const isFieldValid = fieldValidator(field, maxLength)
   if (!isFieldValid.valid) {
     return isFieldValid.message
   }
   return null
 }
 
-const getFieldValue = (req, fieldNames) => {
-  return lodash.get(req.body, fieldNames)
+const validateDoB = (req) => {
```
sfount: Small thing but do you need the whole `req` object? it would be best to not pass this around in general as it has a lot of information on it and can impact the Express stack.

sfount: To be clear - recommend just passing exactly what you need, this will reduce the risk of developers not being sure how to use this object further down the line. It also means it can have a nicer name.

+ also naming methods

```suggestion
const validateDoB = function validDoB (formValues) => {
```

alexbishop1:
Thanks, will implement your changes in the next PR.

PR #1244, opened 2019-03-06
PP-4463 Tidy up responsible person form

alexbishop1:
- On the Stripe responsible person page, make it so that if someone enters a date of birth that is otherwise valid but does not use a four-digit year, a specific “Date of birth must have a four-digit year” error message is shown.

- Name some anonymous functions.

- Trim whitespace from the beginning and the end of the responsible person form submission. This has the desired side-effect that all-whitespace fields will fail validation in the same way empty ones do.

- If a form submission to the Stripe responsible person page does not contain a field at all, default it to the empty string. If the field is mandatory, the existing validation will fail, which is what we want.

- On the check answers page, don’t show a blank line if there is no address line 2 and format the date of birth nicely (e.g. “10 June 2000”).

with @alan-gds

### DIFF
--- a/app/controllers/stripe-setup/responsible-person/post.controller.js
+++ b/app/controllers/stripe-setup/responsible-person/post.controller.js
@@ -53,29 +53,34 @@ const validationRules = [
 ]
 
 module.exports = (req, res) => {
+  const trimmedFormFields = {}
+  Object.entries(req.body).forEach(
+    ([name, value]) => trimmedFormFields[name] = value.trim()
+  );
+
   const errors = validationRules.reduce((errors, validationRule) => {
-    const errorMessage = validate(req.body, validationRule.field, validationRule.validator, validationRule.maxLength)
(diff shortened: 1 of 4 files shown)


app/controllers/stripe-setup/responsible-person/post.controller.js
```diff
@@ -86,26 +102,39 @@ module.exports = (req, res) => {
...
+    pageData['friendlyDateOfBirth'] = friendlyDob
     return response.response(req, res, 'stripe-setup/responsible-person/check-your-answers', pageData)
   }
 }
 
-const validate = (req, fieldName, fieldValidator, maxLength) => {
-  const field = lodash.get(req.body, fieldName)
+const validate = (formFields, fieldName, fieldValidator, maxLength) => {
+  console.log('field name ' + fieldName)
+  console.log('field value ' +  formFields[fieldName])
```
stephencdaly: some logging left in

alexbishop1: Whoops

PR #1251, opened 2019-03-07
PP-4462 Bank details submission

DanailMinchev:
## WHAT

This PR contains Stripe Connect bank details submission implementation and minor tweaks (disable autocomplete and error messages).

Please see commits for more information.

with @alan-gds

### DIFF
--- a/app/controllers/stripe-setup/bank-details/post.controller.js
+++ b/app/controllers/stripe-setup/bank-details/post.controller.js
@@ -4,9 +4,13 @@
 const lodash = require('lodash')
 
 // Local dependencies
-const response = require('../../../utils/response')
+const { response, renderErrorView } = require('../../../utils/response')
 const bankDetailsValidations = require('./bank-details-validations')
+const { updateBankAccount } = require('../../../services/clients/stripe/stripe_client')
+const { ConnectorClient } = require('../../../services/clients/connector_client')
+const connector = new ConnectorClient(process.env.CONNECTOR_URL)
 const paths = require('../../../paths')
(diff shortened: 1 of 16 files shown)


app/controllers/stripe-setup/bank-details/post.controller.js
```diff
@@ -18,24 +22,53 @@ module.exports = (req, res) => {
...
+        return res.redirect(303, paths.dashboard.index)
+      })
+      .catch(error => {
+        // check if it is Stripe related error
+        if (error.code) {
+          if (error.code === 'routing_number_invalid') {
+            // invalid sort code
+            errors = {}
+            errors[SORT_CODE_FIELD] = fieldValidationChecks.validationErrors.invalidSortCode
+            pageData.errors = errors
```
stephencdaly: i think this would be clearer if it were
```pageData.errors = { SORT_CODE_FIELD:  fieldValidationChecks.validationErrors.invalidSortCode }```
re-using the errors object makes it a bit hard to follow

DanailMinchev: `SORT_CODE_FIELD` is constant which is evaluated as a property name, that is why we need `errors[SORT_CODE_FIELD]` which will be resolved to `errors { 'account-number': ... }`
I changed the controller similar way: `pageData.errors[SORT_CODE_FIELD] = ...` and made errors `const` again, thanks

stephencdaly: Ah yes you're right 👍

PR #1257, opened 2019-03-08
PP-4463 Add StripePerson model object for use with Stripe client

alexbishop1:
Add `StripePerson` model object, which represents the details about a responsible person in the format that the Stripe Connect Persons API expects.

### DIFF
--- a/app/services/clients/stripe/stripePerson.model.test.js
+++ b/app/services/clients/stripe/stripePerson.model.test.js
@@ -1,0 +1,235 @@
+'use strict'
+
+// NPM dependencies
+const { expect } = require('chai')
+
+// Local dependencies
+const StripePerson = require('./stripePerson.model')
+
+const firstName = 'Carol'
+const lastName = 'Danvers'
+const addressLine1 = '10 Downing Street'
+const addressLine2 = 'Marvel Cinematic Universe'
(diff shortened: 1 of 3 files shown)


app/services/clients/stripe/stripePerson.model.test.js
```diff
@@ -0,0 +1,222 @@
...
+      address_line1: addressLine1,
+      address_city: addressCity,
+      address_postcode: addressPostcode,
+      dob_day: dobDay,
+      dob_month: 0,
+      dob_year: dobYear
+    })).to.throw('StripePerson "dob_month" must be larger than or equal to 1')
+  })
+
+  it('should fail when month is less than 1000', () => {
```
DanailMinchev: probably copy pasta `should fail when year is less than 1000` ?

alexbishop1: I never copy and paste, Dan. I type everything out by hand. Your eyes are deceiving you.

alexbishop1: No, really, should be fixed now and also added a test for the month being too large.

DanailMinchev: :D

PR #1265, opened 2019-03-11
PP-4735 Resolve Connector URL

kbottla:
## WHAT
- Resolves full connector URL from pagination link (next_page) to fetch all transactions from connector
     - `next_page` reference from connector can now be full_url (http://connector/v2/...) or only context (/charges)

_**Part of removing reliance of selfservice on connector to provide URL to traverse through transactions**_

Currently, connector provides pagination urls (full URLs based on injected UriInfo) when queried for transactions. The URLs generated by connector have wrong port number (unknown to selfservice). As the selfservice is aware of connector URL it only needs to know URI context to traverse through pagination links.

## HOW 
All tests should pass

### DIFF
--- a/app/services/clients/connector_client.js
+++ b/app/services/clients/connector_client.js
@@ -2,6 +2,7 @@
 
 // NPM dependencies
 const _ = require('lodash')
+const url = require('url')
 const util = require('util')
 const EventEmitter = require('events').EventEmitter
 const logger = require('winston')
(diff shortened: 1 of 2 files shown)


rhowe-gds:
How does this cope with the current connector API? i.e. if `next.href` is a full URL

rhowe-gds:
Answering my own question: just fine
```
> url.resolve('https://foo:1234/bar', 'https://foo:5678/bar')
'https://foo:5678/bar'
```

PR #1266, opened 2019-03-11
PP-4949 Use "@pact-foundation/pact" dependency instead of deprecated "pact" dependency

DanailMinchev:
## WHAT

- Remove `pact` dependency and use `@pact-foundation/pact` instead, because pact v4.x.x is no longer being supported
- Use the new Pact object initialisation
- Use the new Pact matchers initialisation


sfount:
Will `require` not need to be updated for the test files?

DanailMinchev:
Of course, just missed that one! :)

PR #1272, opened 2019-03-12
PP-4970: Ensure function is passed to test promise

sfount:
All promise chains replaced in this PR immediately execute the method
being passed in (even before the original method will have run). This
impacts sequencing and is suspected to be causing flakeyness in build
environments.

This should make sure that all defined contracts are actually published and don't silently fail.

Updates duplicated contract descriptions to make sure they can be published.

Issue found investigating https://github.com/alphagov/pay-selfservice/pull/1271


sfount:
It looks like Pact broker may not be responding correctly -- once we have addressed that issue this refactor can be rebuilt and reviewed.

PR #1274, opened 2019-03-12
PP-4939 Fix error navigating to Organisation Details when name is set

stephencdaly:
When organisation details are incomplete, i.e. name or mandatory address fields are not present, take the user to the edit details page with existing details pre-filled rather than the view details page. This
fixes the 500 error thrown when the country code was not set on the view details page.

### DIFF
--- a/test/fixtures/service_fixtures.js
+++ b/test/fixtures/service_fixtures.js
@@ -10,36 +10,6 @@ const pactBase = require(path.join(__dirname, '/pact_base'))
 // Global setup
 const pactServices = pactBase({ array: ['service_ids'] })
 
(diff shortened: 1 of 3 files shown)


test/fixtures/service_fixtures.js
```diff
@@ -313,7 +283,16 @@ module.exports = {
...
+      service.merchant_details = _.pick(opts.merchant_details, [
+        'name',
+        'address_line1',
+        'address_line2',
+        'address_city',
+        'address_postcode',
+        'address_country',
+        'email',
+        'telephone_number'
+      ])
```
stephencdaly: Now that any merchant_details fields can be empty, just return the fields that are passed in in the opts

PR #1291, opened 2019-03-15
PP-4949 Enable Stripe setup tests and vat number company number flag

DanailMinchev:
## WHAT

- Enable the disabled tests
- Rename `organisation_details` to `vat_number_company_number` in StripeAccountSetup
[/v1/api/accounts/{accountId}/stripe-setup](https://github.com/alphagov/pay-connector/blob/20c4e66c45f96d2663884596c2be8ce24b9c1074/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountSetupResource.java#L42) endpoint now returns [`vat_number_company_number`](https://github.com/alphagov/pay-connector/blob/20c4e66c45f96d2663884596c2be8ce24b9c1074/src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeAccountSetup.java#L17)


DanailMinchev:
@sfount :
In terms of Stripe account:
it will update `vat_id` and `tax_id` ( https://stripe.com/docs/api/accounts/update -- `company` object )
In terms why it is concatenated with `and`:
as per story PP-4949 we have one "entry page / workflow" to update them and we use the flag to tell if the workflow was done or not, because the workflow includes both of the properties we use them in the flag name for clarity and we need to tell that the workflow was used no matter which property was populated.

PR #1293, opened 2019-03-15
PP-4463 Responsible person page: paths, routes and Cypress tests

alexbishop1:
- Add routes and paths so the Stripe responsible person page is generally available
- Add Cypress tests for responsible person page

I know it’s large but it’s mostly fairly short, independent Cypress tests.

### DIFF
--- a/test/cypress/integration/stripe-setup/responsible_person_spec.js
+++ b/test/cypress/integration/stripe-setup/responsible_person_spec.js
@@ -1,0 +1,518 @@
+'use strict'
+
+const commonStubs = require('../../utils/common_stubs')
+
+describe('Stripe setup: responsible person page', () => {
+  const gatewayAccountId = 42
+  const userExternalId = 'userExternalId'
+
+  const firstName = 'William'
+  const typedFirstName = 'William '
+  const lastName = 'Benn'
+  const typedLastName = ' Benn'
+  const addressLine1 = '52 Festive Road'
+  const typedAddressLine1 = ' 52 Festive Road'
+  const addressLine2 = 'Putney'
+  const typedAddressLine2 = 'Putney '
+  const city = 'London'
+  const typedCity = 'London '
+  const postcode = 'SW15 1LP'
+  const typedPostcode = 'sw151lp '
+  const dobDay = '25'
+  const typedDobDay = '25 '
+  const dobMonth = '02'
+  const typedDobMonth = ' 02'
+  const dobYear = '1971'
+  const typedDobYear = '1971 '
+  const friendlyDob = '25 February 1971'
+  const longText = 'This text is 300 ...............................................................................' +
+    '...............................................................................................................' +
+    '............................................................................ characters long'
+
+  const stubStripeAccountGet = function stubStripeAccountGet (stripeAccountId) {
+    return {
+      name: 'getStripeAccountSuccess',
+      opts: {
+        gateway_account_id: gatewayAccountId,
+        stripe_account_id: stripeAccountId
+      }
+    }
+  }
+
+  const stubStripeSetupGet = function stubStripeSetupGet (responsiblePersonCompleted) {
+    return {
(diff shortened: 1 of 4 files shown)


test/cypress/integration/stripe-setup/responsible_person_spec.js
```diff
@@ -0,0 +1,518 @@
...
+  const dobMonth = '02'
+  const typedDobMonth = ' 02'
+  const dobYear = '1971'
+  const typedDobYear = '1971 '
+  const friendlyDob = '25 February 1971'
+  const longText = 'This text is 300 ...............................................................................' +
+    '...............................................................................................................' +
+    '............................................................................ characters long'
+
+  const stubStripeAccountGet = function stubStripeAccountGet (stripeAccountId) {
```
DanailMinchev: Better naming 👍 Probably we should use similar in the other tests as well

test/cypress/integration/stripe-setup/responsible_person_spec.js
```diff
@@ -0,0 +1,518 @@
...
+  const postcode = 'SW15 1LP'
+  const typedPostcode = 'sw151lp '
+  const dobDay = '25'
+  const typedDobDay = '25 '
+  const dobMonth = '02'
+  const typedDobMonth = ' 02'
+  const dobYear = '1971'
+  const typedDobYear = '1971 '
+  const friendlyDob = '25 February 1971'
+  const longText = 'This text is 300 ...............................................................................' +
```
DanailMinchev: As far as I calculated it's 299 characters long :D

test/cypress/integration/stripe-setup/responsible_person_spec.js
```diff
@@ -0,0 +1,518 @@
...
+      name: 'getGatewayAccountStripeSetupSuccess',
+      opts: {
+        gateway_account_id: gatewayAccountId,
+        responsible_person: responsiblePersonCompleted
+      }
+    }
+  }
+
+  const stubStripeSetupGetForMultipleCalls = function stubStripeSetupGetForMultipleCalls (...responsiblePersonCompleted) {
+    const data = responsiblePersonCompleted.map(completed => (
```
DanailMinchev: Nice 👍 , we should use similar style in the other tests as well

test/cypress/integration/stripe-setup/responsible_person_spec.js
```diff
@@ -0,0 +1,518 @@
...
+      ])
+
+      cy.visit('/responsible-person')
+    })
+
+    it('should display form', () => {
+      cy.get('h1').should('contain', 'Nominate a responsible person')
+
+      cy.get('form[method=post][action="/responsible-person"]').should('exist')
+        .within(() => {
```
DanailMinchev: 👍 nice optimisation

test/cypress/integration/stripe-setup/responsible_person_spec.js
alexbishop1: It dawned on me that “stub” can be used as a verb and then I was off to the races!

test/cypress/integration/stripe-setup/responsible_person_spec.js
alexbishop1: 😲

test/cypress/integration/stripe-setup/responsible_person_spec.js
alexbishop1: I’m actually quite proud of this!

test/cypress/integration/stripe-setup/responsible_person_spec.js
alexbishop1: I know, right? I just stumbled across it in the Cypress docs and was all like:

```css
im so.happy > we [do=not][have~to] have long.selector:prefixes(everywhere)
```

PR #1299, opened 2019-03-18
PP-4949 Stripe client: update company

DanailMinchev:
## WHAT

- Add `updateCompany` to Stripe client and related `StripeCompany` model

### DIFF
--- a/app/services/clients/stripe/stripeCompany.model.js
+++ b/app/services/clients/stripe/stripeCompany.model.js
@@ -28,12 +28,12 @@ class StripeCompany {
 function build (params) {
   const stripeCompany = {
     company: {
-      vat_id: params.vat_id.replace(/\W/g, '')
+      vat_id: params.vat_id
     }
   }
 
   if (params.tax_id) {
-    stripeCompany.company.tax_id = params.tax_id.replace(/\W/g, '')
+    stripeCompany.company.tax_id = params.tax_id
   }
 
   return stripeCompany
--- a/app/services/clients/stripe/stripeCompany.model.js
+++ b/app/services/clients/stripe/stripeCompany.model.js
@@ -1,0 +1,42 @@
+'use strict'
+
+// NPM dependencies
+const Joi = require('joi')
+
+const schema = {
+  vat_id: Joi.string().required(),
+  tax_id: Joi.string().optional()
+}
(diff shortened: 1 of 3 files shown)


app/services/clients/stripe/stripeCompany.model.js
```diff
@@ -0,0 +1,42 @@
+'use strict'
+
+// NPM dependencies
+const Joi = require('joi')
+
+const schema = {
+  vat_id: Joi.string().required(),
+  tax_id: Joi.string().optional()
```
alexbishop1: Isn’t `.optional()` the default?

app/services/clients/stripe/stripeCompany.model.js
```diff
@@ -0,0 +1,42 @@
...
+
+  basicObject () {
+    return Object.assign({}, this)
+  }
+}
+
+function build (params) {
+  const stripeCompany = {
+    company: {
+      vat_id: params.vat_id.replace(/\W/g, '')
```
alexbishop1: Not sure we should be doing this normalisation here. I think the controller that takes the input should probably do it and this class should just require the input to be well-formed.

app/services/clients/stripe/stripeCompany.model.js
DanailMinchev: yes, it is, I wanted to be explicitly clear that it is optional as per docs https://github.com/hapijs/joi/blob/v14.3.1/API.md#anyoptional
```
Used to annotate the schema for readability as all keys are optional by default.
```

app/services/clients/stripe/stripeCompany.model.js
DanailMinchev: Sure, I removed the normalisation from here if it will be handled in controller, thanks 👍

app/services/clients/stripe/stripeCompany.model.js
alexbishop1: Fair enough — do you think we should also do this in `stripePerson.model.js`?

DanailMinchev: If we all agree, probably yes, for consistency. I was thinking to make it explicit as in Java `Optional`

PR #1311, opened 2019-03-20
PP-4976 Add validation for organisation address page

SandorArpa:
## WHAT

Add post controller for the organisation page with server-side validation. Currently this will not submit the organisation details if validation is not successful.
Use error messages produced by the validation in the template, rather than error messages hardcoded in the template.

Co-authored-by: @SandorArpa

### DIFF
--- a/app/controllers/request-to-go-live/organisation-address/post.controller.js
+++ b/app/controllers/request-to-go-live/organisation-address/post.controller.js
@@ -1,0 +1,97 @@
+'use strict'
+
+// NPM dependencies
+const lodash = require('lodash')
+
+// Local dependencies
+const { requestToGoLive } = require('../../../paths')
+const {
+  validateMandatoryField, validateOptionalField, validatePostcode, validatePhoneNumber
+} = require('../../../utils/validation/server-side-form-validations')
+const { renderErrorView } = require('../../../utils/response')
+
+const ADDRESS_LINE1_FIELD = 'address-line1'
+const ADDRESS_LINE2_FIELD = 'address-line2'
+const ADDRESS_CITY_FIELD = 'address-city'
+const ADDRESS_COUNTRY_FIELD = 'address-country'
+const ADDRESS_POSTCODE_FIELD = 'address-postcode'
+const TELEPHONE_NUMBER_FIELD = 'telephone-number'
+
+const validationRules = [
+  {
+    field: ADDRESS_LINE1_FIELD,
+    validator: validateMandatoryField,
+    maxLength: 200
+  },
+  {
+    field: ADDRESS_LINE2_FIELD,
+    validator: validateOptionalField,
+    maxLength: 200
+  },
+  {
+    field: ADDRESS_CITY_FIELD,
+    validator: validateMandatoryField,
+    maxLength: 100
+  },
+  {
+    field: ADDRESS_COUNTRY_FIELD,
+    validator: validateMandatoryField,
+    maxLength: 2
+  },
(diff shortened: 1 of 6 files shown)


app/controllers/request-to-go-live/organisation-address/post.controller.js
```diff
@@ -0,0 +1,97 @@
...
+  const country = formFields[ADDRESS_COUNTRY_FIELD]
+  const postCodeValidResponse = validatePostcode(postCode, country)
+  if (!postCodeValidResponse.valid) {
+    errors[ADDRESS_POSTCODE_FIELD] = postCodeValidResponse.message
+  }
+  return errors
+}
+
+module.exports = (req, res) => {
+  const trimField = (fieldName) => {
```
sfount: A small performance comment but this could be declared outside of the HTTP response controller so that it's not initialised on every request. 

It could also be slightly simplified, something like:
``` 
const trimField = (key, store) => lodash.get(store, key, '').trim()
```

With the frequency of these requests it's likely negligible so up to you.

app/controllers/request-to-go-live/organisation-address/post.controller.js
```diff
@@ -0,0 +1,97 @@
...
+    maxLength: 2
+  },
+  {
+    field: TELEPHONE_NUMBER_FIELD,
+    validator: validatePhoneNumber
+  }
+]
+
+const validate = function validate (formFields) {
+  const errors = {}
```
sfount: Potential functional simplification, `reduce` validation rules into errors. 
```js
const errors = validationRules.reduce((errors, validation) => { 
  const value = formFields[validationRule.field]
  const validationResponse = validationRule.validator(value, validationRule.maxLength)
  if (!validationResponse.valid) {
    errors[validationRule.field] = validationResponse.message
  }
  return errors
}, {})
```

jonheslop: 👍 Good shout

app/controllers/request-to-go-live/organisation-address/post.controller.js
```diff
@@ -0,0 +1,97 @@
...
+  return errors
+}
+
+module.exports = (req, res) => {
+  const trimField = (fieldName) => {
+    return lodash.get(req.body, fieldName, '').trim()
+  }
+
+  const formFields = {}
+  formFields[ADDRESS_LINE1_FIELD] = trimField(ADDRESS_LINE1_FIELD)
```
sfount: Potential functional simplification:
```js
const fields = [ ADDRESS_LINE1_FIELD, ADDRSS_LINE2_FIELD ] // ...
const formFields = fields.reduce((form, field) => { 
  form[field] = trimField(field)
  return form
}, {})
```

app/controllers/request-to-go-live/organisation-address/post.controller.js
stephencdaly: 👍 made this change

app/controllers/request-to-go-live/organisation-address/post.controller.js
stephencdaly: 🙆‍♂️

app/controllers/request-to-go-live/organisation-address/post.controller.js
stephencdaly: 👌

jonheslop:
Nice one

PR #1312, opened 2019-03-20
PP-4976 Migrate update merchant details calls to use patch endpoint

stephencdaly:
The service patch endpoint is now used for all calls to update merchant details.
It was previously just used for updating the merchant name, but now the client method has been made more generic to make the patch request for any combination of merchant details being updated

Co-authored-by: Sandor Arpa <sandor.arpa@digital.cabinet-office.gov.uk> @SandorArpa


app/controllers/edit_merchant_details/post-edit-controller.js
```diff
@@ -16,21 +16,39 @@ const ADDRESS_POSTCODE = 'address-postcode'
...
-    address_postcode: req.body[ADDRESS_POSTCODE],
-    address_country: req.body[ADDRESS_COUNTRY]
+    name: formFields[MERCHANT_NAME],
+    telephone_number: formFields[TELEPHONE_NUMBER].replace(/\s/g, ''),
+    email: formFields[MERCHANT_EMAIL],
+    address_line1: formFields[ADDRESS_LINE1],
+    address_line2: formFields[ADDRESS_LINE2],
+    address_city: formFields[ADDRESS_CITY],
+    address_postcode: formFields[ADDRESS_POSTCODE],
+    address_country: formFields[ADDRESS_COUNTRY]
```
alexbishop1: I wish it was obvious that these keys need to match what adminusers expects in the `PATCH`…

stephencdaly: hmm maybe there is a way to make this better. I'll look into this in the cleanup at the end

stephencdaly:
> This is a nice rationalisation.
> 
> Am I right in thinking we have no validation for most of the merchant details fields apart from whether they’re mandatory (e.g. no length checks)? If you’re adding this for the new page, it would be good to apply it here as well.

👍 I plan to make the validations on the organisation details page the same as the new page in a later PR

PR #1316, opened 2019-03-21
PP-4976 Use a model to construct update service requests

stephencdaly:
Replace updateMerchantDetails method with a more general updateService method. This updateService method can be used for all service update patch requests, and should replace other methods that make these requests.

Add a ServiceUpdateRequest model to construct the request payload with ops and paths defined as constants.

### DIFF
--- a/app/controllers/edit_merchant_details/post-edit-controller.js
+++ b/app/controllers/edit_merchant_details/post-edit-controller.js
@@ -6,6 +6,7 @@ const paths = require('../../paths')
 const serviceService = require('../../services/service_service')
 const { isPhoneNumber, isValidEmail } = require('../../browsered/field-validation-checks')
 const formattedPathFor = require('../../utils/replace_params_in_path')
+const { validPaths, validOps, ServiceUpdateRequest } = require('../../services/ServiceUpdateRequest.class')
 
 const MERCHANT_NAME = 'merchant-name'
 const TELEPHONE_NUMBER = 'telephone-number'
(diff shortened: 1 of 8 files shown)


app/controllers/edit_merchant_details/post-edit-controller.js
```diff
@@ -38,19 +39,22 @@ module.exports = (req, res) => {
...
-    address_line1: formFields[ADDRESS_LINE1],
-    address_line2: formFields[ADDRESS_LINE2],
-    address_city: formFields[ADDRESS_CITY],
-    address_postcode: formFields[ADDRESS_POSTCODE],
-    address_country: formFields[ADDRESS_COUNTRY]
-  }
+  formFields[TELEPHONE_NUMBER] = formFields[TELEPHONE_NUMBER].replace(/\s/g, '')
+
+  const serviceUpdateRequest = new ServiceUpdateRequest()
+    .addUpdate(validOps.replace, validPaths.merchantDetails.name, formFields[MERCHANT_NAME])
```
sfount: As this request builder understands the context of the operations would it be appropriate to add alias methods for them (mostly just syntactic sugar)? It might read well and would mean not having to import the operations.

Example: 
```js
const serviceUpdateRequest = new ServiceUpdateRequest()
  .replace(paths.merchant.name, formFields[MERCHANT_NAME])
  .replace()
  .add()
  .formatPayload()
```

app/controllers/edit_merchant_details/post-edit-controller.js
```diff
@@ -6,6 +6,7 @@ const paths = require('../../paths')
 const serviceService = require('../../services/service_service')
 const { isPhoneNumber, isValidEmail } = require('../../browsered/field-validation-checks')
 const formattedPathFor = require('../../utils/replace_params_in_path')
+const { validPaths, validOps, ServiceUpdateRequest } = require('../../services/ServiceUpdateRequest.class')
 
 const MERCHANT_NAME = 'merchant-name'
```
sfount: It might be useful in a future refactor to think about calling these something along the lines of `clientFormIds` or something. This would make it easy to understand what they are when reading the code.

Also it would then mean that
```js
requestFormBody[clientFormsIds.MERCHANT_NAME]
```

would very clearly mean the merchant name attribute from the client submitted body.

app/controllers/edit_merchant_details/post-edit-controller.js
```diff
@@ -38,19 +39,22 @@ module.exports = (req, res) => {
...
+    .addUpdate(validOps.replace, validPaths.merchantDetails.addressLine2, formFields[ADDRESS_LINE2])
+    .addUpdate(validOps.replace, validPaths.merchantDetails.addressCity, formFields[ADDRESS_CITY])
+    .addUpdate(validOps.replace, validPaths.merchantDetails.addressPostcode, formFields[ADDRESS_POSTCODE])
+    .addUpdate(validOps.replace, validPaths.merchantDetails.addressCountry, formFields[ADDRESS_COUNTRY])
+    .formatPayload()
+
   const errors = isValidForm(formFields, hasDirectDebitGatewayAccount)
   if (lodash.isEmpty(errors)) {
-    return serviceService.updateMerchantDetails(externalServiceId, reqMerchantDetails, correlationId)
+    return serviceService.updateService(externalServiceId, serviceUpdateRequest, correlationId)
```
sfount: Really nice pattern :+1:

app/services/ServiceUpdateRequest.class.js
```diff
@@ -0,0 +1,38 @@
...
+    email: 'merchant_details/email'
+  }
+}
+
+const validOps = {
+  add: 'add',
+  replace: 'replace'
+}
+
+class ServiceUpdateRequestClass {
```
sfount: If other controllers will be relying on this builder it would be good to have a few unit tests that validate the core functionality.

app/controllers/edit_merchant_details/post-edit-controller.js
stephencdaly: 👍 I'm going to refactor this class a bit at the end of the story which is to add a new page doing similar things to this page, so I'll do that.

app/models/ServiceUpdateRequest.class.test.js
```diff
@@ -0,0 +1,50 @@
...
+  it('should successfully add an "add" request', () => {
+    const payload = new ServiceUpdateRequest().add('foo', 'bar').formatPayload()
+    expect(payload).to.deep.equal([{
+      'op': 'add',
+      'path': 'foo',
+      'value': 'bar'
+    }])
+  })
+
+  it('should successfully add multiple requests', () => {
```
sfount: Nice :+1: seems to cover it

PR #1319, opened 2019-03-22
PP-4910 Fix tests

kbottla:
## WHAT

Fixes failing test `should download a csv file when the next_page is relative URI (not full connector URL)`

## HOW 
CI should pass


sfount:
Looks like this originally failed on a `timeout` issue. Have we ever seen this test timeout before? 

If necessary we can increase the tolerance for this particular test with:
```js
describe('some test', () => { 
  this.timeout(5000)
})
```

kbottla:
@sfount timeout was not related to this test but could be causing build failures intermittently. We can increase it if we see it more often.

PR #1322, opened 2019-03-22
PP-4949 add vat and company id pages

cobainc0:
## WHAT
Add Company registration number page commit : ceedc3e5e9e2145a5bd6b95bbaf2d57ae1be53cd
 - added new template for company-number page
 - created sub-directories for `vat-number` and `company-number`,
 which now live in the parent directory `vat-number-company-number`
 - the `vat-number-company-number` get controller  will now redirect to
 the first page in this umbrella, `vat-number`

Add VAT number page commit: fe83340742acf93b8646045b6d49bf67ae6f3742
- adds Controller logic and client and server-side validation


app/browsered/field-validation-checks.js
```diff
@@ -101,3 +102,12 @@ exports.isNotSortCode = value => {
...
 }
+
+exports.isNotVatNumber = value => {
+  const alphaNumeric = value.replace(/\W/g, '')
+  if (/^(GB)?([0-9]{9}([0-9]{3})?|[A-Z]{2}[0-9]{3})$/.test(alphaNumeric)) {
+    return false
+  } else {
+    return validationErrors.invalidVatNumber
+  }
+}
```
stephencdaly: I think we need some unit tests for this

app/controllers/stripe-setup/vat-number-company-number/vat-number/get.controller.js
```diff
@@ -0,0 +1,17 @@
...
+const lodash = require('lodash')
+
+// Local dependencies
+const response = require('../../../../utils/response')
+
+module.exports = (req, res) => {
+  // initialise pageData
+  let pageData = {
+    errors: lodash.get(req, 'session.stripeSetup.vatNumber.errors'),
+    vatNumber: lodash.get(req, 'session.stripeSetup.vatNumber.vatNumber', '')
```
stephencdaly: If we have session variables set, do we need to make sure we delete them after we get them?
A scenario could be that a user fills in the page for one service, then tries to fill it in for another service afterwards and finds there are pre-filled fields.

app/controllers/stripe-setup/vat-number-company-number/vat-number/post.controller.js
```diff
@@ -0,0 +1,43 @@
...
+    return res.redirect(303, stripeSetup.vatNumber)
+  } else {
+    lodash.set(req, 'session.stripeSetup.vatNumber', {
+      success: true,
+      vatNumber: vatNumber
+    })
+    return res.redirect(303, stripeSetup.companyNumber)
+  }
+
+  function validateVatNumber (vatNumber) {
```
stephencdaly: i think its nicer if this function is declared outside of the exported function, that might just be me though

stephencdaly:
Looks good, just a few things the main one being adding unit tests for the validations.
This might be a matter of moving some of the tests for the new validation class that uses it down into `field_validation_checks_test`

app/controllers/stripe-setup/vat-number-company-number/vat-number/get.controller.js
```diff
@@ -8,9 +8,14 @@ const response = require('../../../../utils/response')
...
-  let pageData = {
-    errors: lodash.get(req, 'session.stripeSetup.vatNumber.errors'),
-    vatNumber: lodash.get(req, 'session.stripeSetup.vatNumber.vatNumber', '')
+  let pageData = lodash.get(req, 'session.pageData.stripeSetup.vatNumber')
+  if (pageData) {
+    delete req.session.pageData.stripeSetup.vatNumber
+  } else {
+    pageData = {
+      errors: {},
+      vatNumber: ''
```
stephencdaly: probably not necessary to set anything in pageData in this case? but not a big deal

DanailMinchev: yes, I was wondering the same, but thought that it is good to know the structure of it if you think it is more confusing to have the blank values?

PR #1323, opened 2019-03-24
PP-4976 Improve organisation details page post controller

stephencdaly:
- Make validation pattern match that for the new request to go live organisation address page
- Use async/await in post-edit-controller and handle all errors
- Move the setting of some session variables for the get controller to a better place
- Modify tests to be unit tests rather than integration tests
- Add improved tests for checking the update of merchant details

### DIFF
--- a/app/controllers/request-to-go-live/organisation-address/post.controller.js
+++ b/app/controllers/request-to-go-live/organisation-address/post.controller.js
@@ -28,17 +28,17 @@ const validationRules = [
   {
     field: clientFieldNames.addressLine1,
     validator: validateMandatoryField,
-    maxLength: 200
+    maxLength: 255
   },
   {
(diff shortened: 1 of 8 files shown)


app/controllers/request-to-go-live/organisation-address/post.controller.js
```diff
@@ -26,27 +26,27 @@ const clientFormIds = {
 
 const validationRules = [
   {
-    field: clientFormIds.addressLine1,
+    field: clientFieldNames.addressLine1,
     validator: validateMandatoryField,
     maxLength: 200
```
alexbishop1: These seem to be out of sync with merchant_details/post-edit-controller.js

stephencdaly: Good point, I've updated them

PR #1327, opened 2019-03-25
PP-4976 Use controller errors on the edit merchant details page

stephencdaly:
The 'errors' object returned by the post controller now contains error messages as the values, where previously there was just a boolean to indicate there was an error.
Use these error messages in the template rather than the existing static text.

This modifies some of the error messages, but makes them more consistent with the design system.


stephencdaly:
It only lists the address once in the summary box. The summary box behaviour is unchanged from what it was previously. Although previously it wasn't highlighting the address line 2 input if the contents failed validation 😕

PR #1344, opened 2019-03-28
PP-4949 Stripe setup: company number page implementation

DanailMinchev:
## WHAT

- Implemented controller logic
- Added tests
- Other minor modifications

### DIFF
--- a/app/browsered/field-validation-checks.js
+++ b/app/browsered/field-validation-checks.js
@@ -9,6 +9,7 @@ const MAX_AMOUNT = 100000
 
 const validationErrors = {
   required: 'This field cannot be blank',
+  mandatoryQuestion: 'You must answer this question',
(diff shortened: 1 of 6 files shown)


app/browsered/field-validation-checks.js
```diff
@@ -116,10 +117,19 @@ exports.isNotVatNumber = value => {
 
 exports.isNotCompanyNumber = value => {
   const sanitisedCompanyNumber = value.replace(/\s/g, '').toUpperCase()
+
+  // check explicitly for empty string and make it valid,
+  // to be able to handle the validation on the client side.
+  // the empty string validation will be checked on server side.
```
stephencdaly: maybe a bit clearer:

Allow an empty string, as company number is not always required. If it is required, this will be flagged by the server side validation.

DanailMinchev: Yes, thanks, will change it in the next PR

PR #1345, opened 2019-03-28
PP-4949 Stripe setup: "VAT number / company number - check your answers" page display logic

DanailMinchev:
## WHAT

- Implement "VAT number / company number - check your answers" page with display logic
  and change of the data behaviour

### DIFF
--- a/test/cypress/integration/stripe-setup/vat-number-company-number/check_your_answers_spec.js
+++ b/test/cypress/integration/stripe-setup/vat-number-company-number/check_your_answers_spec.js
@@ -1,0 +1,334 @@
+'use strict'
+
+// Local dependencies
+const commonStubs = require('../../../utils/common_stubs')
+const {
+  stubGetGatewayAccountStripeSetupSuccess,
+  stubStripeAccountGet,
+  stubStripeSetupGetForMultipleCalls
+} = require('./support')
+
+describe('Stripe setup: "VAT number / company number - check your answers" page', () => {
+  const gatewayAccountId = 42
+  const userExternalId = 'userExternalId'
+  const validVatNumber = 'GB999 9999 73'
+  const validCompanyNumber = '01234567'
+
+  describe('Card gateway account', () => {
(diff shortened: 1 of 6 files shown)


test/cypress/integration/stripe-setup/vat-number-company-number/check_your_answers_spec.js
```diff
@@ -0,0 +1,334 @@
...
+        cy.location().should((location) => {
+          expect(location.pathname).to.eq('/vat-number-company-number/check-your-answers')
+        })
+        cy.get('dl.govuk-summary-list > div.govuk-summary-list__row:nth-child(1) > dd.govuk-summary-list__value')
+          .should('contain', validVatNumber)
+        cy.get('dl.govuk-summary-list > div.govuk-summary-list__row:nth-child(2) > dd.govuk-summary-list__value')
+          .should('contain', 'None')
+      })
+
+      it('should go to VAT number page when VAT number "change" button is clicked and change tbe value', () => {
```
stephencdaly: To avoid having to go through all the steps each time for these happy path tests with a common setup, which takes time, is there an argument that they could be in a single describe along with the first happy path test using `Cypress.Cookies.preserveOnce(names...)` to keep the cookies between the different "if"s.
This way you could combine the checking the display of the check your answers page and checking going back and changing the answers in one continuing describe block.

This would also save a lot of lines of code. What do you think?

stephencdaly: Although given time constraints I wouldn't bother changing it. Maybe an improvement for someone else to do in the future...

DanailMinchev: Yes, great idea! I was also not happy to copy & pasta a lot of code, `Cypress.Cookies.preserveOnce` is a good approach as you suggested.
As discussed this can be done later at some point.

PR #1350, opened 2019-03-29
PP-4949 Check your answers page submit implementation

DanailMinchev:
## WHAT

- Implement "VAT number / company number - check your answers" page with submission logic
- Render generic error and do not check for Stripe specific errors, as the only one Stripe error is `tax_id_invalid` and our validation should prevent it.


DanailMinchev:
🎉

PR #1351, opened 2019-03-29
BAU - Set JAVA_HOME for pact provider tests

kbottla:
## WHAT

- both connector and adminusers using Java 11.


danworth:
@alexbishop1 yes that's a good point. Looking at the consumer/provider relationships, for now during pact tests it'll be Connector running its tests with 11JDK whilst still on 8 in production. We plan to release Connector to 11 on Monday to give some time for other issues to surface with the others.

https://pact-broker-test.cloudapps.digital/groups/selfservice

PR #1352, opened 2019-03-29
PP-4997 Stripe fee/ net columns for transaction views

sfount:
* Update transaction list to show fee, net columns 
* Update transaction details to show fee, net columns 
* Add consumer contract test with transaction fee structure 
* Add Cypress validation for showing/ hiding columns appropriately

Note: PR is in draft state as no transactions will have fees associated until Stripe billing is live, the code is ready to review

### DIFF
--- a/app/controllers/transactions/transaction_list_controller.js
+++ b/app/controllers/transactions/transaction_list_controller.js
@@ -29,7 +29,6 @@ module.exports = (req, res) => {
   transactionService
     .search(accountId, filters.result, correlationId)
     .then(transactions => {
-      console.log('got all transactions', transactions)
       client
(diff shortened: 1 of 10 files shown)


mrlumbu:
Screen Shot 2019-05-01 at 12 20 32
Net column has ugly overflow for refunds. As a side note, we should consider what we mean and how we represent fees + net with refunds. I think for now this is fine - least surprise and all that

sfount:
@mrlumbu I've added the `white-space: pre` style to both amount and net amount columns to stop this overflow with minimal code changes -- the best way of presenting negative values should be investigated separately. 

Screen Shot 2019-05-01 at 16 35 26

PR #1365, opened 2019-04-04
PP-5059 Refactor update service name

SandorArpa:
## WHAT
- refactor updateServiceName to update all gateway account service names in connector.
- add relevant test


app/services/service_service.js
```diff
@@ -61,26 +61,26 @@ function getGatewayAccounts (gatewayAccountIds, correlationId) {
...
-          const gatewayAccountId = accounts[1]
-          if (gatewayAccountId && gatewayAccountId.length > 0 && !isADirectDebitAccount(gatewayAccountId)) {
+  if (!serviceExternalId) {
+    return Promise.reject(new Error(`argument: 'serviceExternalId' cannot be undefined`))
+  }
+  return getAdminUsersClient({ correlationId }).updateServiceName(serviceExternalId, serviceName, serviceNameCy)
+    .then(result => {
+      // Update gateway account service names in connector
+      const gatewayAccountIds = lodash.get(result, 'gateway_account_ids', [])
+      if (gatewayAccountIds.length <= 0) {
```
sfount: (Entirely optional) If we are always going to return the service entity here we could simplify this by performing the `Promise.all` no matter what - `Promise.all([])` will return successfully

Something like: 
```js
updateServiceName()
  .then(() => { 
    return Promise.all(...)
  })
  .then(() => { 
   return new Service(details)
  })
```

app/services/service_service.js
```diff
@@ -61,26 +61,26 @@ function getGatewayAccounts (gatewayAccountIds, correlationId) {
...
+        return new Service(result)
+      } else {
+        return Promise.all(gatewayAccountIds.map(gatewayAccountId => {
```
sfount: This `map` needs to return a result in order to wait on the result. Right now it looks like this will just `Promise.all[undefined]` which will just skip.

I.e 
```js 
map((id) => { 
  return connectorClient.patchServiceName(id)
}
```

test/unit/services/service_service_test.js
```diff
@@ -276,4 +277,61 @@ describe('service service', function () {
...
+        {
+          '../services/clients/connector_client': connectorClientStub,
+          '../services/clients/direct_debit_connector_client': directDebitClientStub,
+          './clients/adminusers_client': adminusersClientStub
+        })
+
+      serviceService.updateServiceName(externalServiceId, newServiceName, correlationId).should.be.fulfilled
+        .then((service) => {
+          setTimeout(() => {
+            expect(patchServiceName.callCount).to.equal(2)
```
sfount: 👍

PR #1370, opened 2019-04-08
PP-5013 Display Welsh specific instructions on payment link information page

stephencdaly:
- Add a supported-language enum to hold the supported language values.
- Display the "/create-payment-link/information" page with Welsh specific instructions if the query parameter "language=cy" is provided.
- Add Cypress tests covering the information page.

### DIFF
--- a/app/views/payment-links/information.njk
+++ b/app/views/payment-links/information.njk
@@ -14,7 +14,11 @@ Create a payment link - {{currentService.name}} {{currentGatewayAccount.full_typ
 
 {% block mainContent %}
 <section class="govuk-grid-column-two-thirds govuk-!-margin-top-6">
-  <h1 class="govuk-heading-l">Set payment link information</h1>
+  {% if isWelsh %}
+    <h1 class="govuk-heading-l">Set Welsh payment link information</h1>
+  {% else %}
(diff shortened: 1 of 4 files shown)


app/views/payment-links/information.njk
```diff
@@ -31,20 +35,28 @@ Create a payment link - {{currentService.name}} {{currentGatewayAccount.full_typ
     {% endif %}
     {% set confirmationLabel %}{{friendlyURL}}/{{currentService.name | removeIndefiniteArticles | slugify}}/{% endset %}
 
+    {% if isWelsh %}
+      {% set titleLabel = 'Welsh title' %}
+      {% set titleHint = 'Briefly describe what the user is paying for. For example, “Talu am drwydded barcio”. This will also be your website address.' %}
```
alexbishop1: Can we make sure the Welsh text is marked up in a `<span lang="cy">` element? I think using `html` rather than `text` in the `govukInput` macro will allow the `titleHint` to contain HTML (see https://design-system.service.gov.uk/components/text-input/#options-text-input-example--hint for details).

stephencdaly: Ah yes! I'll do that.

PR #1372, opened 2019-04-08
PP-4819: use async await to order Google Pay requests

gidsg:
## WHAT
use `async` `await` to order Google Pay PATCH requests 

Previously we were using `Promise.all` this caused a potential race condition where one PATCH happened before the other completed causing an exception in Connector.

## HOW 
Cypress tests should pass, functionality should work reliably on environment.

### DIFF
--- a/app/controllers/digital-wallet/post-enable-google-pay-controller.js
+++ b/app/controllers/digital-wallet/post-enable-google-pay-controller.js
@@ -1,28 +1,26 @@
 'use strict'
 
 // Local dependencies
+const logger = require('winston')
 const paths = require('../../paths')
 const { ConnectorClient } = require('../../services/clients/connector_client')
 const { CORRELATION_HEADER } = require('../../utils/correlation_header.js')
 const connector = new ConnectorClient(process.env.CONNECTOR_URL)
 
-module.exports = (req, res) => {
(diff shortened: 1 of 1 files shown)


oswaldquek:
i think this is good but having @sfount review it too would be a good thing

app/controllers/digital-wallet/post-enable-google-pay-controller.js
```diff
@@ -1,23 +1,26 @@
...
   const gatewayAccountId = req.account.gateway_account_id
   const correlationId = req.headers[CORRELATION_HEADER] || ''
   const gatewayMerchantId = req.body.merchantId
-  const enableGooglePayBoolean = connector.toggleGooglePay(gatewayAccountId, true, correlationId)
-  const setGatewayMerchantId = connector.setGatewayMerchantId(gatewayAccountId, gatewayMerchantId, correlationId)
-
-  Promise.all([setGatewayMerchantId, enableGooglePayBoolean]).then(() => {
+  try {
+    await connector.toggleGooglePay(gatewayAccountId, true, correlationId)
+    logger.info(`${correlationId} enabled google pay boolean for ${gatewayAccountId}`)
```
sfount: 👍 logging

PR #1378, opened 2019-04-10
PP-4997 Add `fee` `net` columns to CSV export

sfount:
Represent the fee and net colums in the self service CSV export - attributes are only available following https://github.com/alphagov/pay-selfservice/pull/1352. 

Depends on features updated in https://github.com/alphagov/pay-selfservice/pull/1352


sfount:
Closed by https://github.com/alphagov/pay-selfservice/pull/1378/commits

PR #1379, opened 2019-04-10
PP-5017 Separate Welsh payment links

stephencdaly:
- Display any Welsh payment links in a separate section on the "Manage payment links" page
- Separate out the HTML for displaying an individual product to a new template so it can be re-used for English and Welsh payment link sections.
- Add Cypress tests
- Remove get_manage_controller.test as the test cases are covered better by the Cypress tests now


app/views/payment-links/_product.njk
```diff
@@ -0,0 +1,31 @@
+<li class="pay-!-border-bottom govuk-!-padding-bottom-3 govuk-!-padding-bottom-3">
```
stephencdaly: extracted without edits from app/views/payment-links/manage.njk

test/cypress/integration/payment-links/manage_payment_links_spec.js
```diff
@@ -0,0 +1,194 @@
...
+      ])
+    })
+
+    it('should list Welsh payment links', () => {
+      cy.visit('/create-payment-link/manage')
+
+      cy.get('h1').should('contain', 'Manage payment links')
+      cy.get('.payment-links-list--header').should('contain',
+        'There are 2 payment links')
+
```
oswaldquek: could there be a way to test there are no english payment links here?

stephencdaly: good point, I could check that there is only one "ul.payment-links-list"

oswaldquek: why one and not zero?

stephencdaly: one because there is a list of Welsh payment links within 'div#welsh-payment-links' which will be picked up by that selector.

PR #1384, opened 2019-04-11
PP-4997 Connector transactions contract with fees

sfount:
Consumer contracts between self service and connector for transactions with fees.


SandorArpa:
No longer a valid test

PR #1388, opened 2019-04-12
PP-5012 Send language when creating product

stephencdaly:
Send language when creating a payment link, getting the language from whether the payment link is Welsh or not.
When no language is provided to the create product client, such as for creating a demo payment, default to English.

### DIFF
--- a/test/unit/controller/payment-links/post_review_controller.test.js
+++ b/test/unit/controller/payment-links/post_review_controller.test.js
@@ -2,19 +2,19 @@
 
 // NPM dependencies
 const supertest = require('supertest')
-const {expect} = require('chai')
+const { expect } = require('chai')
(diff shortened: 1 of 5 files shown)


test/unit/controller/payment-links/post_review_controller.test.js
```diff
@@ -35,181 +35,229 @@ const VALID_CREATE_TOKEN_REQUEST = {
 const VALID_MINIMAL_GATEWAY_ACCOUNT_RESPONSE = {
   payment_provider: 'sandbox'
```
stephencdaly: Make sure 'Diff settings' > 'Hide whitespace changes' is enabled when reviewing this as I removed a load of nested describes which will make it look like there are loads of changes if not.

PR #1410, opened 2019-04-24
PP-5108 Use true booleans instead of strings

cobainc0:
## WHAT
Connector is already capable of handling Strings and Booleans. But we would like to use true booleans in place of "boolean strings". 

Solo: @cobainc0


alexbishop1:
Can this also be used for false booleans? 😛

PR #1411, opened 2019-04-24
PP-5108 Add pact tests for wallet payments

cobainc0:
## WHAT
Add pact tests for gateway accounts supporting digital wallets
- connector now expects boolean values for `allow_google_pay` & `allow_apple_pay` flags

Solo: @cobainc0

### DIFF
--- a/test/fixtures/gateway_account_fixtures.js
+++ b/test/fixtures/gateway_account_fixtures.js
@@ -45,6 +45,22 @@ function validGatewayAccount (opts) {
 }
 
 module.exports = {
+  validGatewayAccountPatchRequest: (opts = {}) => {
+    const data = {
+      op: 'replace',
+      path: opts.path,
+      value: opts.value
+    }
+
+    return {
+      getPactified: () => {
+        return pactRegister.pactify(data)
+      },
+      getPlain: () => {
+        return _.clone(data)
+      }
+    }
+  },
   validGatewayAccountEmailRefundToggleRequest: (enabled = true) => {
     const data = {
       op: 'replace',
--- a/test/unit/clients/connector_client/connector_patch_apple_pay_toggle_test.js
+++ b/test/unit/clients/connector_client/connector_patch_apple_pay_toggle_test.js
@@ -1,0 +1,89 @@
+'use strict'
+
+// NPM dependencies
+const Pact = require('pact')
+const chai = require('chai')
+const chaiAsPromised = require('chai-as-promised')
+
+// Custom dependencies
+const path = require('path')
+const PactInteractionBuilder = require('../../../fixtures/pact_interaction_builder').PactInteractionBuilder
+const Connector = require('../../../../app/services/clients/connector_client').ConnectorClient
+const gatewayAccountFixtures = require('../../../fixtures/gateway_account_fixtures')
+
+// Constants
+const ACCOUNTS_RESOURCE = '/v1/api/accounts'
+const port = Math.floor(Math.random() * 48127) + 1024
+const connectorClient = new Connector(`http://localhost:${port}`)
+const expect = chai.expect
+const existingGatewayAccountId = 666
+
+// Global setup
+chai.use(chaiAsPromised)
+
+describe('connector client - patch apple pay toggle (enabled) request', () => {
+  const patchRequestParams = { path: 'allow_apple_pay', value: true }
+  const request = gatewayAccountFixtures.validGatewayAccountPatchRequest(patchRequestParams).getPlain()
+
+  let provider = Pact({
+    consumer: 'selfservice',
+    provider: 'connector',
+    port: port,
+    log: path.resolve(process.cwd(), 'logs', 'mockserver-integration.log'),
+    dir: path.resolve(process.cwd(), 'pacts'),
+    spec: 2,
+    pactfileWriteMode: 'merge'
+  })
+
+  before(() => provider.setup())
+  after(() => provider.finalize())
+
+  describe('apple pay toggle - supported payment provider request', () => {
+    const applePayToggleSupportedPaymentProviderState =
+      `a gateway account supporting digital wallet with external id ${existingGatewayAccountId} exists in the database`
+
+    before(() => {
+      return provider.addInteraction(
+        new PactInteractionBuilder(`${ACCOUNTS_RESOURCE}/${existingGatewayAccountId}`)
+          .withUponReceiving('a valid patch apple pay toggle (enabled) request')
+          .withState(applePayToggleSupportedPaymentProviderState)
+          .withMethod('PATCH')
+          .withRequestBody(request)
+          .withStatusCode(200)
+          .withResponseHeaders({})
+          .build())
+    })
+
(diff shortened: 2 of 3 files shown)


test/unit/clients/connector_client/connector_patch_apple_pay_toggle_test.js
```diff
@@ -0,0 +1,93 @@
...
+    dir: path.resolve(process.cwd(), 'pacts'),
+    spec: 2,
+    pactfileWriteMode: 'merge'
+  })
+
+  before(() => provider.setup())
+  after(() => provider.finalize())
+
+  describe('apple pay toggle - supported payment provider request', () => {
+    const applePayToggleSupportedPaymentProvider =
```
danworth: perhaps this should end in `State` as per the other similar names.

test/fixtures/gateway_account_fixtures.js
```diff
@@ -45,6 +45,38 @@ function validGatewayAccount (opts) {
 }
 
 module.exports = {
+  validGatewayAccountGooglePayToggleRequest: (enabled = false) => {
```
danworth: Since these two methods are incredibly similar I'm wondering if this could instead be a `validGatewayAccountPatchRequest(op, path, value)` so it can be used for any patch request? You can keep the understanding of what the request is when its used e.g. 
```
const validToggleApplePayPatchRequest = validGatewayAccountPatchRequest(......
```

test/unit/clients/connector_client/connector_patch_apple_pay_toggle_test.js
```diff
@@ -0,0 +1,93 @@
...
+        .should.be.fulfilled
+        .notify(done)
+    })
+  })
+
+  describe('apple pay toggle with unsupported payment provider request', () => {
+    const applePayToggleUnsupportedPaymentProviderState = `User ${existingGatewayAccountId} exists in the database`
+    const unsupportedPaymentProvider = 'sandbox'
+    const errorResponse = {
+      message: `Gateway ${unsupportedPaymentProvider} does not support digital wallets.`
```
danworth: @stephencdaly raised a good point about not asserting on things in pact tests which we don't actually rely upon. So if we're not using `message` then we should not assert on it. Maybe speak with Stephen as I think he's been removing these fields in pact tests.

test/unit/clients/connector_client/connector_patch_google_pay_toggle_test.js
```diff
@@ -0,0 +1,93 @@
...
+        .should.be.fulfilled
+        .notify(done)
+    })
+  })
+
+  describe('google pay toggle with unsupported payment provider request', () => {
+    const googlePayToggleUnsupportedPaymentProviderState = `User ${existingGatewayAccountId} exists in the database`
+    const unsupportedPaymentProvider = 'sandbox'
+    const errorResponse = {
+      message: `Gateway ${unsupportedPaymentProvider} does not support digital wallets.`
```
danworth: Same point about asserting on `message` element.

test/unit/clients/connector_client/connector_patch_apple_pay_toggle_test.js
cobainc0: 👍 Good shout. Thanks

test/fixtures/gateway_account_fixtures.js
cobainc0: 👍 Updated. Thanks

test/unit/clients/connector_client/connector_patch_apple_pay_toggle_test.js
cobainc0: Awesome! 👍

test/fixtures/gateway_account_fixtures.js
cobainc0: I've renamed this so that it's a "generic wallet payment toggle" request: `validGatewayAccountWalletToggleRequest`

test/unit/clients/connector_client/connector_patch_apple_pay_toggle_test.js
```diff
@@ -0,0 +1,94 @@
...
+
+    before(() => {
+      return provider.addInteraction(
+        new PactInteractionBuilder(`${ACCOUNTS_RESOURCE}/${existingGatewayAccountId}`)
+          .withUponReceiving('a valid patch apple pay toggle (enabled) request')
+          .withState(applePayToggleUnsupportedPaymentProviderState)
+          .withMethod('PATCH')
+          .withRequestBody(request)
+          .withStatusCode(400)
+          .withResponseBody(errorResponse)
```
danworth: I think since we don't use the response body, e.g. the error message then this can be removed `withResponseBody(errorResponse)`. My understanding is that this will create a pact which the pact broker will run against Connector and make sure that Connector's response includes the expected error message. Since Selfservice doesn't rely upon it, then I think the point @stephencdaly is making is that we shouldn't create a pact between SelfService and Connector that enforces it.

test/fixtures/gateway_account_fixtures.js
```diff
@@ -45,6 +45,22 @@ function validGatewayAccount (opts) {
 }
 
 module.exports = {
+  validGatewayAccountWalletToggleRequest: (opts = {}) => {
```
danworth: Since this is in `gateway_account_fixtures` and all our patch requests require a path and value (I think `op` is always `replace` then this could be called `validGatewayAccountPatchRequest` since it need not be specific to patching wallet payments (if you remove the `|| false`).

test/unit/clients/connector_client/connector_patch_google_pay_toggle_test.js
```diff
@@ -0,0 +1,94 @@
...
+const port = Math.floor(Math.random() * 48127) + 1024
+const connectorClient = new Connector(`http://localhost:${port}`)
+const expect = chai.expect
+const existingGatewayAccountId = 666
+
+// Global setup
+chai.use(chaiAsPromised)
+
+describe('connector client - patch google pay toggle (enabled) request', () => {
+  const params = { path: 'allow_google_pay', value: true }
```
danworth: Tiny point but `patchRequestParams` might be more descriptive.

test/unit/clients/connector_client/connector_patch_apple_pay_toggle_test.js
cobainc0: I understand. I guess the point here was to make it obvious that **sandbox** doesn't support digital wallets, but if you think it's clear without it, then I can remove it.

test/fixtures/gateway_account_fixtures.js
```diff
@@ -45,6 +45,22 @@ function validGatewayAccount (opts) {
 }
 
 module.exports = {
+  validGatewayAccountPatchRequest: (opts = {}) => {
```
danworth: I would be inclined to just have the arguments as `(path, value)` rather than wrapped in an object `opts` object but I don't want to hold you up.

PR #1417, opened 2019-04-25
PP-5153 Update existing Stripe person

stephencdaly:
Stripe creates a blank "person" on a test account when it is created. Therefore rather than creating a new person, we need to update the existing person.
We update the final person returned when we call Stripe to list the persons as this will be the first person created.

### DIFF
--- a/app/controllers/stripe-setup/responsible-person/post.controller.js
+++ b/app/controllers/stripe-setup/responsible-person/post.controller.js
@@ -12,7 +12,7 @@ const { response, renderErrorView } = require('../../../utils/response')
 const {
   validateMandatoryField, validateOptionalField, validatePostcode, validateDateOfBirth
 } = require('../../../utils/validation/server-side-form-validations')
-const { createPerson } = require('../../../services/clients/stripe/stripe_client')
+const { listPersons, updatePerson } = require('../../../services/clients/stripe/stripe_client')
 const { ConnectorClient } = require('../../../services/clients/connector_client')
 const connector = new ConnectorClient(process.env.CONNECTOR_URL)
 
@@ -60,7 +60,7 @@ const validationRules = [
 
 const trimField = (key, store) => lodash.get(store, key, '').trim()
 
-module.exports = (req, res) => {
+module.exports = async function (req, res) {
   const fields = [
     FIRST_NAME_FIELD,
     LAST_NAME_FIELD,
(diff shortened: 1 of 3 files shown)


app/controllers/stripe-setup/responsible-person/post.controller.js
```diff
@@ -106,17 +106,17 @@ module.exports = (req, res) => {
...
-        logger.error(`[requestId=${req.correlationId}] Error creating responsible person with Stripe - ${error.message}`)
-        return renderErrorView(req, res, 'Please try again or contact support team')
-      })
+    try {
+      const stripeAccountId = res.locals.stripeAccount.stripeAccountId
+      const persons = await listPersons(stripeAccountId)
+      const person = persons.pop()
+      await updatePerson(stripeAccountId, person.id, buildStripePerson(formFields))
+      await connector.setStripeAccountSetupFlag(req.account.gateway_account_id, 'responsible_person', req.correlationId)
+      await res.redirect(303, paths.dashboard.index)
```
alexbishop1: `await` is much more _promising_ than `then`

app/services/clients/stripe/stripe_client.js
```diff
@@ -37,8 +37,13 @@ module.exports = {
     return stripe.accounts.update(stripeAccountId, company.basicObject())
   },
 
-  createPerson: function (stripeAccountId, body) {
+  listPersons: async function (stripeAccountId) {
+    const persons = await stripe.accounts.listPersons(stripeAccountId)
+    return persons.data
```
alexbishop1: Why can’t this just do `stripe.accounts.listPersons(stripeAccountId)`?

stephencdaly: yeh it could, it returns an object of form:
{
     "object": "list",
     "data" : [ ... persons ... ]
}

I thought maybe it was clearer if we returned the actual persons list here, but if you think otherwise I can change it.

app/controllers/stripe-setup/responsible-person/post.controller.js
stephencdaly: 🤦‍♂️

app/services/clients/stripe/stripe_client.js
alexbishop1: I just thought it was a bit inconsistent that `updatePerson` returns immediately but the `listPersons` will wait for the request to Stripe to come back.

Anyway, happy to go with whatever you think is best.

stephencdaly: yeh that's fair, I can change it for consistency

PR #1423, opened 2019-04-29
PP-4889 Add metadata to transaction detail page.

danworth:
## WHAT
Add metadata to transactions details page, should not show if there is no metadata for the charge.

### DIFF
--- a/test/cypress/integration/transactions/transaction_details_spec.js
+++ b/test/cypress/integration/transactions/transaction_details_spec.js
@@ -326,4 +326,26 @@ describe('Transactions details page', () => {
     cy.visit(`${transactionsUrl}/${chargeDetails.charge.charge_id}`)
     cy.get('.transaction-details tbody').should('not.contain', 'Wallet Type')
   })
+
+  it('should display metadata when available', () => {
+    const chargeDetails = defaultChargeDetails()
+    chargeDetails.charge.metadata = {
+      key1: 123,
+      key2: true,
+      key3: 'some string'
+    }
+    cy.task('setupStubs', getStubs(chargeDetails))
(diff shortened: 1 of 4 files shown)


test/cypress/integration/transactions/transaction_details_spec.js
```diff
@@ -326,4 +326,27 @@ describe('Transactions details page', () => {
...
   })
+
+  it('should display metadata when available', () => {
+    const chargeDetails = defaultChargeDetails()
+    chargeDetails.charge.metadata = {
+      key1: 123,
+      key2: true,
+      key3: 'some string'
+    }
+    console.log(chargeDetails)
```
jonheslop: Probs can remove the log

danworth: thanks

test/cypress/integration/transactions/transaction_details_spec.js
```diff
@@ -326,4 +326,26 @@ describe('Transactions details page', () => {
...
+    chargeDetails.charge.metadata = {
+      key1: 123,
+      key2: true,
+      key3: 'some string'
+    }
+    cy.task('setupStubs', getStubs(chargeDetails))
+    cy.visit(`${transactionsUrl}/${chargeDetails.charge.charge_id}`)
+    cy.get('h2').should('contain', 'Metadata')
+    cy.get('th').contains('key1').siblings().first().should('contain', 123)
+    cy.get('th').contains('key2').siblings().first().should('contain', true)
```
alexbishop1: Strictly speaking, as these are asserting on text from the page, they should be `'123'` and `'true'` (note the quotation marks)

danworth: Fixed 👍

PR #1459, opened 2019-05-13
PP-5199 Enable transactions pact tests

stephencdaly:
* Set the provider to "selfservice" and make the Pact states match the states that exist in connector.
* Modify test fixtures so they produce the correct pacts.
* Modify Cypress tests to work with the modified fixtures.

### DIFF
--- a/test/fixtures/transaction_fixtures.js
+++ b/test/fixtures/transaction_fixtures.js
@@ -38,26 +38,16 @@ const buildChargeEventStateWithDefaults = (opts = {}) => {
   return state
 }
 
-const buildTransactionWithDefaults = (opts = {}) => {
+const buildTransactionDetailsWithDefaults = (opts = {}) => {
   const data = {
     amount: opts.amount || 20000,
-    state: {
-      finished: opts.state_finished || false,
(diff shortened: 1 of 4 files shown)


test/fixtures/transaction_fixtures.js
```diff
@@ -76,15 +66,46 @@ const buildTransactionWithDefaults = (opts = {}) => {
         line1: opts.billing_address_line1 || 'address line 1',
         line2: opts.billing_address_line2 || 'address line 2',
         postcode: opts.billing_address_postcode || 'AB1A 1AB',
-        city: opts.billing_address_city || 'GB',
-        county: opts.billing_address_county || null,
+        city: opts.billing_address_city || 'London',
+        county: opts.billing_address_county || 'Greater London',
```
alexbishop1: Not sure this is realistic: although county exists in parts of our model for some reason, it can’t ever be set, so it should probably always be null

stephencdaly: I've removed that field from the pact fixture entirely, so it won't be asserted on by the pact

PR #1475, opened 2019-05-21
PP-5193 Check permission for link gocardless account

stephencdaly:
- Resolving gateway account was failing when navigating to the link account page, as the get gateway account middleware was expecting the service to have already been resolved. It then ended up resolving the wrong gateway account. The permissions middleware resolves the service so adding this to the stack fixes the issue.
- Use a new specific permission that was added for accessing this setting.
- Simplify navigation_ui_tests by having just 2 tests to check that the correct links are shown for card/direct debit gateway accounts when the user has all permissions.

Co-authored-by: Rory Malcolm <rory.malcolm@digital.cabinet-office.gov.uk>

### DIFF
--- a/test/fixtures/user_fixtures.js
+++ b/test/fixtures/user_fixtures.js
@@ -178,6 +178,14 @@ const defaultPermissions = [
   {
     name: 'stripe-vat-number-company-number:read',
     description: 'View Stripe VAT number company number'
+  },
+  {
+    name: 'connected-gocardless-account:read',
+    description: 'View connected go cardless account'
+  },
+  {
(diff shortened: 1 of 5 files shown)


test/fixtures/user_fixtures.js
```diff
@@ -178,6 +178,14 @@ const defaultPermissions = [
   {
     name: 'stripe-vat-number-company-number:read',
     description: 'View Stripe VAT number company number'
+  },
+  {
+    name: 'connected-gocardless-account:read',
+    description: 'View connected go cardless account'
```
alexbishop1: “GoCardless”

PR #1478, opened 2019-05-22
PP-4819 - Show Digital wallets tab now we have officially launched the feature

jonheslop:


app/views/payment-types/index.njk
```diff
@@ -17,17 +17,15 @@
   <h1 class="govuk-heading-l page-title">Manage payment types</h1>
 
   <p class="govuk-body">Choose which credit and debit cards you want to accept.
-  {# TODO: uncomment this block for Digital Wallet global launch
-  {% if isDigitalWalletSupported %} You can also let users pay using a digital wallet, like Apple Pay.{% endif %} #}
+    {% if isDigitalWalletSupported %} You can also let users pay using a digital wallet, like Apple Pay.{% endif %}
```
alexbishop1: Is there a reason we don’t mention Apple Pay and Google Pay?

jonheslop: 🤷‍♂

jonheslop: Have added Google Pay

PR #1481, opened 2019-05-24
PP-5281 Remove {EPDQ,SMARTPAY}_3DS_ENABLED feature flags

mrlumbu:
We support 3DS for ePDQ and Smartpay, and no longer need these flags


oswaldquek:
🐉 dracarys

PR #1491, opened 2019-05-28
PP-5179 - 3DS settings redesign

jonheslop:
- New template
- All new async controllers
- Refactored connector client to be promise based
- Cypress tests
- Removed all the old mocha tests


app/controllers/toggle-3ds/3ds.get.controller.js
```diff
@@ -0,0 +1,31 @@
...
+      disabled3ds: req.account.disableToggle3ds,
+      supports3ds: req.account.supports3ds,
+      requires3ds: req.account.requires3ds,
+      hasCardTypeRequiring3dsEnabled: lodash.some(acceptedCards, 'requires3ds'),
+      showHelper3ds: req.account.payment_provider === 'worldpay'
+    }
+
+    return response(req, res, 'toggle-3ds/index', pageData)
+  } catch (error) {
+    return renderErrorView(req, res, error.message.message[0], error.errorCode)
```
stephencdaly: I think the error message passed in here is displayed to the user. We probably don't want to display an error that comes from connector.

app/controllers/toggle-3ds/3ds.post.controller.js
```diff
@@ -0,0 +1,31 @@
...
+        toggle_3ds: enable
+      },
+      correlationId: correlationId
+    }
+
+    await connector.update3dsEnabled(params)
+    req.flash('generic', '3D secure settings have been updated')
+    return res.redirect(paths.toggle3ds.index)
+  } catch (error) {
+    return renderErrorView(req, res, error.message.message[0], error.errorCode)
```
stephencdaly: Same with the error message here.

PR #1492, opened 2019-05-30
PP-5182 - Apple/Google Pay settings refresh

jonheslop:
- Standardising the flow for each
- Soon will introduce settings page, PP-5184 so preparing the pages for that


test/cypress/integration/settings/allow_apple_pay_spec.js
```diff
@@ -2,46 +2,86 @@ describe('Apple Pay', () => {
...
+            service_roles: [{
+              service: {
+                gateway_account_ids: [gatewayAccountId],
+                name: serviceName
+              }
+            }]
+          }
+        },
+        {
+          name: 'getGatewayAccountSuccessRepeatNTimes',
```
stephencdaly: I don't think you need to use this repeated stub, this is only intended for when the same request will return a different response on subsequent calls, whereas the response is the same both times here.

test/cypress/integration/settings/allow_apple_pay_spec.js
```diff
@@ -2,46 +2,86 @@ describe('Apple Pay', () => {
...
+          opts: [{
+            gateway_account_id: gatewayAccountId,
+            payment_provider: paymentProvider
```
stephencdaly: Would it be good to specify that `allow_apple_pay` is `false` for this response? even though this is the default it would add some clarity.

test/cypress/integration/settings/allow_google_pay_spec.js
```diff
@@ -5,46 +5,84 @@ describe('Google Pay', () => {
...
+            service_roles: [{
+              service: {
+                gateway_account_ids: [gatewayAccountId],
+                name: serviceName
+              }
+            }]
+          }
+        },
+        {
+          name: 'getGatewayAccountSuccessRepeatNTimes',
```
stephencdaly: same here, don't need to use N times

test/cypress/integration/settings/allow_apple_pay_spec.js
jonheslop: Yeah i didnt know what that did but was just there in the test already 🤷‍♂, will refactor 👍

PR #1502, opened 2019-06-07
Pp 5266 scenario 2

oswaldquek:
Show 'Connect to GoCardless' page if Gateway Account is live, uses the direct debit payment method and is not connected to GoCardless yet.

At this point I haven't wrote any tests and the html/css styling isn't quite right.

### DIFF
--- a/app/views/dashboard/index.njk
+++ b/app/views/dashboard/index.njk
@@ -5,7 +5,7 @@ Dashboard - {{currentService.name}} {{currentGatewayAccount.full_type}} - GOV.UK
 {% endblock %}
 
 {% block mainContent %}
-  {% if account.paymentMethod === 'direct debit' and account.type === 'live' and account.isConnected === false %}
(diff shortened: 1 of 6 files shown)


app/views/dashboard/index.njk
```diff
@@ -5,12 +5,18 @@ Dashboard - {{currentService.name}} {{currentGatewayAccount.full_type}} - GOV.UK
 {% endblock %}
 
 {% block mainContent %}
-  <div class="govuk-grid-column-full">
-    <h1 class="govuk-heading-l govuk-!-margin-top-6 first-steps__title">Dashboard</h1>
-  </div>
-
-  {% if paymentMethod === 'card' %}
-    {% include "./_activity.njk" %}
+  {% if account.paymentMethod === 'direct debit' and account.type === 'live' and account.isConnected === false %}
```
stephencdaly: I think we can leave out the check that the account is 'live', as we're able to connect sandbox accounts. We do need a check that the payment provider is GoCardless though.

PR #1507, opened 2019-06-10
PP-5266: Show GoCardless dashboard when account is already linked

oswaldquek:


app/views/dashboard/index.njk
```diff
@@ -10,6 +10,11 @@ Dashboard - {{currentService.name}} {{currentGatewayAccount.full_type}} - GOV.UK
       <h1 class="govuk-heading-l govuk-!-margin-top-6 first-steps__title">Connect to GoCardless</h1>
       {% include "./_link-to-gocardless.njk" %}
     </div>
+  {% elif account.paymentMethod === 'direct debit' and account.paymentProvider === 'gocardless' and account.isConnected === true %}
```
stephencdaly: would it be worthwhile simplifying this by moving 
```
{% if account.paymentMethod === 'direct debit' and account.paymentProvider === 'gocardless' %}
```
into a surrounding block. Obviously doesn't change the functionality but do you think it's easier to read?

oswaldquek: it's probably easier to read. i'll change it.

oswaldquek: fixed

PR #1511, opened 2019-06-12
PP-5185 settings page extended tests

jonheslop:
Extended settings tests to start from the settings index page not their subpages.


stephencdaly:
What about billing address? Is that already covered?

jonheslop:
Good point i missed it cos it doesnt have cypress tests, can do in another PR

PR #1512, opened 2019-06-12
PP-5266: Pact test for getting gateway accounts from directdebit conn…

oswaldquek:
…ector

I created a new direct_debit_connector_client2.js as I spent a day trying to set a pact server port on direct_debit_connector_client.js with no success.

### DIFF
--- a/app/services/clients/direct_debit_connector_client2.js
+++ b/app/services/clients/direct_debit_connector_client2.js
@@ -1,0 +1,28 @@
+'use strict'
+
+// Local dependencies
+const baseClient = require('./base_client/base_client')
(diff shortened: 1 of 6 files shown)


oswaldquek:
depends on https://github.com/alphagov/pay-direct-debit-connector/pull/497

app/services/clients/direct_debit_connector_client2.js
```diff
@@ -0,0 +1,28 @@
+'use strict'
```
stephencdaly: is this replacing the old one?

oswaldquek: eventually

PR #1514, opened 2019-06-12
PP-5185 - settings page redundant for direct debit gateways

jonheslop:
Settings page errored because it couldnt retrieve card only settings on a DD gateway, so redirect from this page to API Keys, which is basically what used to happen when you clicked settings

### DIFF
--- a/test/cypress/integration/settings/settings-index.spec.js
+++ b/test/cypress/integration/settings/settings-index.spec.js
@@ -1,0 +1,43 @@
+describe('Settings page', () => {
+  const userExternalId = 'cd0fa54cf3b7408a80ae2f1b93e7c16e'
+  const gatewayAccountId = 'DIRECT_DEBIT:42'
+  const serviceName = 'A Direct Debit Service'
+
+  describe('should redirect to API keys if Direct Debit gateway', () => {
+    beforeEach(() => {
+      cy.task('setupStubs', [
+        {
+          name: 'getUserSuccess',
+          opts: {
+            external_id: userExternalId,
+            service_roles: [{
+              service: {
+                gateway_account_ids: [gatewayAccountId],
+                name: serviceName
+              }
+            }]
+          }
+        },
+        {
+          name: 'getDirectDebitGatewayAccountSuccess',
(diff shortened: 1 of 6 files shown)


test/cypress/integration/settings/settings-index.spec.js
```diff
@@ -0,0 +1,43 @@
...
+        {
+          name: 'getDirectDebitGatewayAccountSuccess',
+          opts: {
+            gateway_account_id: gatewayAccountId
+          }
+        }
+      ])
+    })
+
+    it('should show it is disabled', () => {
```
stephencdaly: what is disabled?

test/cypress/integration/settings/settings-index.spec.js
```diff
@@ -0,0 +1,43 @@
...
+      cy.setEncryptedCookies(userExternalId, gatewayAccountId)
+      cy.request({
+        url: '/settings/',
+        followRedirect: false // turn off following redirects
+      })
+        .then((resp) => {
+          // redirect status code is 302
+          expect(resp.status).to.eq(302)
+          expect(resp.redirectedToUrl).to.contain('/api-keys')
+        })
```
stephencdaly: could you do `cy.visit('/settings'`) then check the location? would that make it more of a cypress-style test?

test/cypress/integration/settings/settings-index.spec.js
jonheslop: Oh lol left over from test I copied will update words

test/cypress/integration/settings/settings-index.spec.js
jonheslop: The problem then is it redirects to `/api-keys` which calls publicApi and there are no fixtures for that yet, so this was quicker 😬

stephencdaly: ahh ok, that's fine then

jonheslop:
Closing in favour of a much simpler solution! https://github.com/alphagov/pay-selfservice/pull/1516

PR #1526, opened 2019-06-25
PP-5366 - use inputmode and pattern attribute on numeric fields

jonheslop:
The Design System team have done some accessibility testing on `input
type="number"` and deemed them problematic. They have suggested using

`input type="text" inputmode="numeric" pattern="[0-9]*"`

instead as documented here https://github.com/alphagov/govuk-frontend/issues/1449


nickcolley:
@jonheslop make sure you pair this change with `novalidate` https://github.com/alphagov/govuk-design-system/issues/941#issuecomment-506286337

PR #1543, opened 2019-07-04
PP-5384 allow user to update phone numbers

jonheslop:
5e39c5b - Add link to update phone number
172ac9b - Add edit phone number page
cd47913 - Add post controller to actually update phone number
f74c6ce - Add phone number validation
b184e9c - Add Cypress test for updating phone number

Linting on commit also cleaned up `user_service.js`

### DIFF
--- a/app/controllers/user/phone-number/get.controller.js
+++ b/app/controllers/user/phone-number/get.controller.js
@@ -1,0 +1,14 @@
+'use strict'
+
+// Local dependencies
+const { renderErrorView } = require('../../../utils/response')
+const userService = require('../../../services/user_service')
+
+module.exports = async (req, res) => {
+  try {
+    const { telephoneNumber } = await userService.findByExternalId(req.user.externalId)
+    return res.render('team-members/edit-phone-number', { telephoneNumber })
+  } catch (error) {
+    return renderErrorView(req, res, 'Unable to retrieve user')
+  }
+}
--- a/app/controllers/user/phone-number/post.controller.js
+++ b/app/controllers/user/phone-number/post.controller.js
@@ -4,8 +4,14 @@
 const { renderErrorView } = require('../../../utils/response')
 const userService = require('../../../services/user_service')
 const paths = require('../../../paths')
+const { invalidTelephoneNumber } = require('../../../utils/validation/telephone-number-validation')
 
 module.exports = async (req, res) => {
+  if (invalidTelephoneNumber(req.body.phone)) {
+    req.flash('genericError', '<h2>There was a problem with the details you gave for:</h2><ul class="govuk-list govuk-error-summary__list"><li><a href="#phone">Invalid telephone number.</a></li></ul>')
+    return res.redirect(paths.user.phoneNumber)
+  }
+
   try {
     await userService.updatePhoneNumber(req.user.externalId, req.body.phone)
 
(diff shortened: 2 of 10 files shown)


app/controllers/user/phone-number/post.controller.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+
+// Local dependencies
+const { renderErrorView } = require('../../../utils/response')
+const userService = require('../../../services/user_service')
+const paths = require('../../../paths')
+const { invalidTelephoneNumber } = require('../../../utils/validation/telephone-number-validation')
+
+module.exports = async (req, res) => {
```
sfount: method could be named to describe what the code is doing + show a more useful stack trace

sfount: 👍  async/ await 🎉

app/controllers/user/phone-number/get.controller.js
```diff
@@ -0,0 +1,14 @@
+'use strict'
+
+// Local dependencies
+const { renderErrorView } = require('../../../utils/response')
+const userService = require('../../../services/user_service')
+
+module.exports = async (req, res) => {
```
sfount: method could be named to describe what the code is doing + show a more useful stack trace

app/services/user_service.js
```diff
@@ -211,10 +211,23 @@ module.exports = {
...
+  /**
+   * @param newPhoneNumber
+   * @returns {Promise}
+   */
+  updatePhoneNumber: (externalId, newPhoneNumber) => {
+    return new Promise((resolve, reject) => {
+      getAdminUsersClient().updatePhoneNumberForUser(externalId, newPhoneNumber)
+        .then(
+          () => resolve(),
+          error => reject(new Error(`There has been a problem updating the phone number: ${error}`))
```
sfount: I think favour `.catch` instead of `(success(), failure())` it reads much more clearly, something like: 

```
getAdminUsersClient().updatePhoneNumberForUser()
  .then(() => {})
  .catch((error) => {})
```

sfount: EDIT: 

This is using a `Promise` for the request anyway, it would be good to avoid any new Promises as it's much easier to make mistakes, I think this can be simplified with: 

```
return 
  getAdminUsersClient().updatePhoneNumberForUser(externalId, newPhoneNumber)
  .catch((error) => new Error(`There has been a problem updating the phone number: ${error}`))
```

app/controllers/user/phone-number/get.controller.js
jonheslop: Would the filename not be in the stack trace? And both these files only export one function so naming it wouldn’t make anything clearer?

app/services/user_service.js
jonheslop: Yeah nice, I had just copy/pasted something else but this is better

app/controllers/user/phone-number/get.controller.js
sfount: Yeah, don't think it needs to be strongly required here. Some quick things I can think of: 
* whether something is POST, GET, PATCH impacts what the file name can mean, it's not enforced in the naming structure
* something like `updateUserPhonePageHandler` might read clearer to someone that doesn't understand how everything fits together 
* for this simpler page this is straight forward but I think it's a good practice for more complex pages

PR #1551, opened 2019-07-09
PP-5355 Don't add null to pact

mrlumbu:
Pacts should not be written expecting values to be null. Pacts
are about the subset of a response that is necessary for the
consumer. Unless the consumer postively uses the fact that a
value is null, it should not write it in to the contract.


test/fixtures/transaction_fixtures.js
```diff
@@ -9,14 +9,19 @@ const pactBase = require(path.join(__dirname, '/pact_base'))
 const pactRegister = pactBase()
 
 const buildChargeEventWithDefaults = (opts = {}) => {
-  return {
+  let chargeEvent = {
```
sfount: `let` could be a `const` here! No problem with updating attributes, the pointer to the original object is what can't be updated

PR #1570, opened 2019-07-23
BAU Remove Snyk local dependency

sfount:
* Snyk is run in CI against PRs and `master` builds, there should be no
need to run it locally with the CLI
* Report for selfservice can be found
https://app.snyk.io/org/govuk-pay/project/987ad850-d52f-42da-a7f3-b1295702f90e/


heathd:
I've often seen different results running via the command line locally vs the scan performed by the webapp. 

I think the local execution scans the contents of node_modules wheras the web based one only looks at some combination of `package.json` and `package_lock.json`. I haven't verified this.

Also `npm audit` may do a better job than snyk these days.

`snyk-protect` also applies patches which may address security vulnerabilities where fixes haven't been released yet. I'm not sure how often this happens in real life https://snyk.io/docs/cli-protect/

sfount:
@heathd it looks like `snyk-protect` only applies patches specified in `.snyk`, all of our rules here expired last year. 

It does look like this functionality could be used though, I'll look at what it takes to add that but remove from local testing. 

Developers would still be able to run snyk locally fine, they would just install it globally rather than at this package level.

iqbalgds:
Closing this - as Snyk is being removed as part of this PR: 
https://github.com/alphagov/pay-scripts/pull/995

PR #1571, opened 2019-07-23
BAU Remove cypress local dependency

sfount:
* Cypress is a massive dependency and requires unzipping, give CI a break
* Ref: https://github.com/alphagov/pay-toolbox/pull/53


sfount:
Closed alongside https://github.com/alphagov/pay-jenkins-library/pull/111

PR #1580, opened 2019-08-05
PP-5512: Strip out email search param for Google Analytics

oswaldquek:
Sending (and hence recording) emails breaks our agreement with Google and is a
breach of our obligations under GDPR.

## WHAT

## HOW 
_Steps to test or reproduce:_

### DIFF
--- a/app/views/includes/head.njk
+++ b/app/views/includes/head.njk
@@ -14,7 +14,9 @@
   ga('set', 'displayFeaturesTask', null);
   ga('set', 'transport', 'beacon');
 
-  ga('send', 'pageview');
+  var filtered_page_path = document.location.pathname + document.location.search.replace(/&email=[\w%.-]*/, '');
+  ga('set', 'page', filtered_page_path);
+  
 </script>
 <!-- End Google Analytics -->
 {% endif %}


app/views/includes/head.njk
```diff
@@ -14,7 +14,13 @@
   ga('set', 'displayFeaturesTask', null);
   ga('set', 'transport', 'beacon');
 
-  ga('send', 'pageview');
```
jankowiakmaria: As far as I understand (https://developers.google.com/analytics/devguides/collection/analyticsjs/command-queue-reference#method-details), without `send` command we won't trace the traffic at all.

app/views/includes/head.njk
```diff
@@ -14,7 +14,13 @@
...
+  var url = new URL(document.location.href);
+  var search_params = new URLSearchParams(url.search);
+  search_params.delete('email');
+  url.search = search_params.toString();
+  var filtered_page_path = url.pathname + url.search;
+  ga('set', 'page', filtered_page_path);
```
jonheslop: Assuming `ga('set', 'page', filtered_page_path);` is equivalent to `ga('send', 'pageview');`

Then this looks good to me

app/views/includes/head.njk
```diff
@@ -14,7 +14,13 @@
   ga('set', 'displayFeaturesTask', null);
   ga('set', 'transport', 'beacon');
 
-  ga('send', 'pageview');
+  var url = new URL(document.location.href);
+  var search_params = new URLSearchParams(url.search);
```
jankowiakmaria: You should be able to retrieve searchParams from url itself https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams

jonheslop: Ahh hang on `URLSearchParams` [isn’t available](https://caniuse.com/#feat=urlsearchparams) in Internet Explorer 😩 

Just checked and 27% of traffic to selfservice is through IE so will have to write a fallback for it otherwise this fix won’t work on any queries in IE

app/views/includes/head.njk
oswaldquek: Yes this is what the guy from Merkle suggested. I'll forward you the email.

app/views/includes/head.njk
oswaldquek: 😩 do you have any suggestions?

app/views/includes/head.njk
```diff
@@ -14,7 +14,13 @@
...
+  search_params.delete('email');
+  url.search = search_params.toString();
+  var filtered_page_path = url.pathname + url.search;
```
jankowiakmaria: didn't want to spend too much time on review, but **probably** there is a nicer way of doing it.

app/views/includes/head.njk
oswaldquek: Ah ok should it be 
```
ga('set', 'page', filtered_page_path);
ga('send', 'pageview');
```
then?

app/views/includes/head.njk
oswaldquek: probably. but i'm no javascript developer. if it works, then i'd rather do this to get something out so we can meet our GDPR legal requirements asap.

app/views/includes/head.njk
oswaldquek: @jonheslop will `(new URL(document.location)).searchParams` work in IE?

app/views/includes/head.njk
jankowiakmaria: I think so.

app/views/includes/head.njk
jonheslop: No, no support for that either https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams#Browser_compatibility 

Maybe we just go with a simpler regex

jonheslop: Like this maybe? https://jsbin.com/kemacav/edit?js,console

jonheslop: Just updated it to this - `/&email=[\w%.-]*/` which is a little more robust

app/views/includes/head.njk
oswaldquek: the analytics consultant from Merkle says it isn't needed so let's go with what he says first.

jonheslop: I don’t see the bit in the email where they say you dont need the send line 🤔 I could be wrong hard to read big threads

I wonder if we should keep it and have it like you say above
```
ga('set', 'page', filtered_page_path);
ga('send', 'pageview');
```

PR #1582, opened 2019-08-07
PP-5500 Ledger service client

sfount:
Add ledger services client using a singleton pattern, the same client should be offered to all modules requiring it, only exposing methods to make requests to ledger.

* update test environment to include ledger URL
* update README with upcoming feature flag

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ See [About Cypress tests in selfservice](./test/cypress/cypress_testing.md) for 
 | HTTPS_PROXY                 |   |      | HTTPS proxy url |
 | NO_PROXY                    |   |      | host:port(s) that need to be by passed by the proxy. Supports comma separated list |
 | NODE_WORKER_COUNT           |   | 1 | The number of worker threads started by node cluster when run in production mode |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -57,6 +57,7 @@ See [About Cypress tests in selfservice](./test/cypress/cypress_testing.md) for
 | HTTPS_PROXY                 |   |      | HTTPS proxy url |
 | NO_PROXY                    |   |      | host:port(s) that need to be by passed by the proxy. Supports comma separated list |
 | NODE_WORKER_COUNT           |   | 1 | The number of worker threads started by node cluster when run in production mode |
+| FEATURE_USE_LEDGER_PAYMENTS |   | false/undefined | Use the Ledger service as the source of payment data in favour of card connector |
```
kbottla: Not sure if we want to document it given it is temporary

sfount: Self service has particularly poor visibility on feature flags -- there's no centralised configuration (like in Toolbox/ Java apps). 

Without this it's up to devs to figure out each of the different files that implements the feature flag. I think it's okay to add it temporarily -- it can be removed with the feature flag.

PR #1585, opened 2019-08-08
PP-5426 Transaction view uses ledger (feature flag)

sfount:
* Transaction details page feature flag to point to ledger or connector
* Depends on (forked from) https://github.com/alphagov/pay-selfservice/pull/1582
* Depends on https://github.com/alphagov/pay-infra/pull/2282

### DIFF
--- a/app/models/charge.js
+++ b/app/models/charge.js
@@ -12,7 +12,7 @@ const connector = new ConnectorClient(process.env.CONNECTOR_URL)
 const Ledger = require('../services/clients/ledger_client')
 
 const { FEATURE_USE_LEDGER_PAYMENTS } = process.env
-const useLedgerTransactions = FEATURE_USE_LEDGER_PAYMENTS !== 'true'
+const useLedgerTransactions = FEATURE_USE_LEDGER_PAYMENTS === 'true'
 
 module.exports = function (correlationId) {
   correlationId = correlationId || ''
@@ -63,7 +63,7 @@ module.exports = function (correlationId) {
 
       return transactionView.buildPaymentView(charge, transactionEvents, users)
     } catch (error) {
-      throw 'CLIENT_UNAVAILBLE' // eslint-disable-line no-throw-literal
+      throw getStatusCodeForError(error)
     }
   }
 
@@ -110,11 +110,17 @@ module.exports = function (correlationId) {
     })
   }
 
-  function findWithEventsError (err, response, reject) {
+  function getStatusCodeForError (err, response) {
+    let status = 'CLIENT_UNAVAILABLE'
     const code = (response || {}).statusCode || (err || {}).errorCode
-    if (code === 404) return reject('NOT_FOUND')
-    if (code > 200) return reject('GET_FAILED')
-    reject('CLIENT_UNAVAILABLE')
+    if (code > 200) status = 'GET_FAILED'
+    if (code === 404) status = 'NOT_FOUND'
+    return status
+  }
+
+  function findWithEventsError (err, response, reject) {
+    const statusCode = getStatusCodeForError(err, response)
+    reject(statusCode)
   }
 
   return {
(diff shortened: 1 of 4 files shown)


app/models/charge.js
```diff
@@ -86,15 +110,21 @@ module.exports = function (correlationId) {
...
 
-  function findWithEventsError (err, response, reject) {
+  function getStatusCodeForError (err, response) {
+    let status = 'CLIENT_UNAVAILABLE'
     const code = (response || {}).statusCode || (err || {}).errorCode
-    if (code === 404) return reject('NOT_FOUND')
-    if (code > 200) return reject('GET_FAILED')
-    reject('CLIENT_UNAVAILABLE')
+    if (code > 200) status = 'GET_FAILED'
+    if (code === 404) status = 'NOT_FOUND'
```
jankowiakmaria: This line will never be reached, because `404` will be caught by the first condition `if (code > 200)`.
That's why they've been in the previous order.

app/services/clients/ledger_client.js
```diff
@@ -8,6 +8,44 @@ const defaultOptions = {
...
+    transaction.refund_summary.amount_submitted = transaction.refund_summary.amount_refunded
+  }
+  transaction.charge_id = transaction.transaction_id
+  return transaction
+}
+
+const transaction = function transaction (id, gatewayAccountId, options = {}) {
+  const configuration = Object.assign({
+    url: `/v1/transaction/${id}`,
+    qs: { account_id: gatewayAccountId },
```
jankowiakmaria: it will be changed to `gateway_account_id` so it might be worth adding this one as well as query parameter (later - after the actual changes the `account_id` will be removed)

app/services/clients/ledger_client.js
```diff
@@ -8,6 +8,44 @@ const defaultOptions = {
...
+  const legacyConnectorEventsParity = (transactionEvents) => {
+    transactionEvents.events = transactionEvents.events.map(event => {
+      return Object.assign({
+        type: event.resource_type.toLowerCase(),
+        updated: event.timestamp,
+        ...(event.data.refunded_by) && { submitted_by: event.data.refunded_by }
+      }, event)
+    })
+    return transactionEvents
+  }
```
jankowiakmaria: [minor] for `events` the `legacyConnector...` is internal function, but for `transaction` it's not. Would it be possible to be more consistent?

app/models/charge.js
sfount: A `let` can be overridden - this isn't a terminal operation

app/services/clients/ledger_client.js
sfount: `transaction` will be used to map both an individual transaction and the transactions picked up in transaction search -- events will only ever be transformed within this one end point. I see how that's not clear until the second end point has been added

app/services/clients/ledger_client.js
sfount: I would rather be explicit? It seems strange to commit multiple possible options to master, this will have to be changed when `account_id` is removed, I'll add a reminder to the ticket

sfount:
@jankowiakmaria there is no `getStatusCodeForError` bug as far as I'm aware -- agreed that legacy connector mappings could have some unit tests to make sure sensible inputs are parsed correctly. 

I think it would be overkill to test the branching behaviour specifically as the idea is that it should be uniform and with the feature flag turned on, pass all of the upstream tests.

sfount:
Sensible inputs are parsed correctly https://github.com/alphagov/pay-selfservice/pull/1585/commits/9beddbd5e369aae138f5740a7819b014e140fa4e

app/services/clients/ledger_client.js
jankowiakmaria: The order will need to be anyway: add a new one and then remove the old, otherwise it would break the functionality.

app/models/charge.js
jankowiakmaria: my bad while reading.
I still think it would make more sense to return from the if statement (and it would require the change of the order back to the previous one).

PR #1606, opened 2019-08-19
PP-5537 add get single transaction pact test

SandorArpa:
## WHAT

- add ledger transaction fixtures
- refactor ledger client to to use optional transaction_type query
- add get transaction in created state details pact test

### DIFF
--- a/test/fixtures/ledger_transaction_fixtures.js
+++ b/test/fixtures/ledger_transaction_fixtures.js
@@ -1,0 +1,227 @@
+'use strict'
+
+// NPM dependencies
+const path = require('path')
+const lodash = require('lodash')
+
+// Global setup
+const pactBase = require(path.join(__dirname, '/pact_base'))
+const pactRegister = pactBase()
+
+const buildChargeEventWithDefaults = (opts = {}) => {
+  const chargeEvent = {
+    type: opts.type || 'PAYMENT',
+    state: buildChargeEventStateWithDefaults(opts.state),
+    amount: opts.amount || 100,
+    updated: opts.updated || '2018-05-01T13:27:00.063Z'
+  }
+  if (opts.refund_reference) {
+    chargeEvent.refunds_reference = opts.refund_reference
+  }
+  if (opts.submitted_by) {
+    chargeEvent.submitted_by = opts.submitted_by
+  }
+  return chargeEvent
+}
+
+const buildChargeEventStateWithDefaults = (opts = {}) => {
+  let state
+  if (opts.status === 'failed') {
+    state = {
+      status: opts.status,
+      finished: true,
+      code: opts.code || 'P0010',
+      message: opts.message || 'Payment method rejected'
+    }
+  } else {
+    state = {
+      status: opts.status || 'started',
+      finished: opts.finished || false
+    }
+  }
+
+  return state
+}
+
+const buildTransactionDetails = (opts = {}) => {
+  const data = {
+    amount: opts.amount || 20000,
+    state: buildChargeEventStateWithDefaults(opts.state),
+    description: opts.description || 'ref1',
+    reference: opts.reference || 'ref188888',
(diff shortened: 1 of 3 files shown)


test/fixtures/ledger_transaction_fixtures.js
```diff
@@ -0,0 +1,287 @@
...
+      successful_payments: { count: 1, total_in_pence: 2 },
+      refunded_payments: { count: 3, total_in_pence: 4 },
+      net_income: { total_in_pence: 5 }
+    }
+
+    return {
+      getPactified: () => {
+        return pactRegister.pactify(data)
+      },
+      getPlain: () => {
```
sfount: slightly more succinctly
```suggestion
      getPlain: () => data
```

test/fixtures/ledger_transaction_fixtures.js
```diff
@@ -0,0 +1,287 @@
...
+module.exports = {
+  validTransactionSummaryResponse: () => {
+    let data = {
+      successful_payments: { count: 1, total_in_pence: 2 },
+      refunded_payments: { count: 3, total_in_pence: 4 },
+      net_income: { total_in_pence: 5 }
+    }
+
+    return {
+      getPactified: () => {
```
sfount: Looks like a lot of these are anonymous method that could directly 'return' the values
```suggestion
      getPactified: () => pactRegister.pactify(data)
```

test/fixtures/ledger_transaction_fixtures.js
```diff
@@ -0,0 +1,287 @@
...
+    state = {
+      status: opts.status || 'started',
+      finished: opts.finished || false
+    }
+  }
+
+  return state
+}
+
+const buildTransactionCreatedDetailsWithDefaults = (opts = {}) => {
```
sfount: As these three methods (`buildTransactionCreatedDetailsWithDefaults`, `buildTransactionDetailsWithDefaults` and `buildTransactionSearchResultWithDefaults`) are all 'builders' would it be possible to have one method that can then add details as needed? 

For example there could be an `opts.includeSettlementSummary` which would only add that part of the object is it was true. 

This might reduce the lines of code needed for the fixture and mean the fixture is only defined once instead of three different times.

test/unit/clients/ledger_client/ledger_get_transaction_details_test.js
```diff
@@ -0,0 +1,80 @@
...
+        new PactInteractionBuilder(`${TRANSACTION_RESOURCE}/${params.transaction_id}`)
+          .withQuery('account_id', params.account_id)
+          .withQuery('transaction_type', 'PAYMENT')
+          .withUponReceiving('a valid created transaction details request')
+          .withState(defaultTransactionState)
+          .withMethod('GET')
+          .withStatusCode(200)
+          .withResponseBody(pactified)
+          .build()
+      ).then(() => done())
```
sfount: As nothing is done with this 'done' we could simplify this by just returning the Promise, something like: 

```js
before(() => { 
  //...
  return provider.addInteractionBuilder()
}
```

test/unit/clients/ledger_client/ledger_get_transaction_details_test.js
```diff
@@ -0,0 +1,80 @@
...
+    })
+
+    afterEach(() => provider.verify())
+
+    it('should get transaction details successfully', function (done) {
+      const getCreatedTransactionDetails = validCreatedTransactionDetailsResponse.getPlain()
+      ledgerClient.transaction(params.transaction_id, params.account_id, { transaction_type: 'PAYMENT' })
+        .then((ledgerResponse) => {
+          expect(ledgerResponse).to.deep.equal(getCreatedTransactionDetails)
+          done()
```
sfount: I think this could also just be returned: 
```js 
it('...', () => { 
  //...
  return ledgerClient.transaction()
    .then((response) => { 
      expected()
    })
  })
})
```

PR #1613, opened 2019-08-27
PP-5537 add get events and search ledger transactions pact tests

SandorArpa:
## WHAT
- add pact tests
- create a pact provider helper class. This is needed so the tests can share the pre-configured ledger_client's port.
- update transaction fixtures accordingly

### DIFF
--- a/test/fixtures/ledger_transaction_fixtures.js
+++ b/test/fixtures/ledger_transaction_fixtures.js
@@ -43,6 +43,20 @@ const buildChargeEventStateWithDefaults = (opts = {}) => {
   return state
 }
 
+const buildPaymentEvents = (opts = {}) => {
+  let events = []
+  if (opts.payment_states) {
+    opts.payment_states.forEach(paymentState => {
+      events.push({
+        state: buildChargeEventStateWithDefaults(paymentState),
+        resource_type: 'PAYMENT',
+        data: {}
+      })
+    })
+  }
+  return events
+}
+
 const buildTransactionDetails = (opts = {}) => {
   const data = {
     amount: opts.amount || 20000,
@@ -73,7 +87,7 @@ const buildTransactionDetails = (opts = {}) => {
         city: opts.billing_address_city || 'London',
         country: opts.billing_address_country || 'GB'
       },
-      card_brand: opts.card_brand || 'Visa'
+      card_brand: opts.card_brand || 'visa'
     }
   }
 
@@ -91,7 +105,7 @@ const buildTransactionDetails = (opts = {}) => {
       status: opts.refund_summary_status || 'unavailable',
       user_external_id: opts.user_external_id || null,
       amount_available: opts.refund_summary_available || 20000,
-      amount_submitted: opts.refund_summary_submitted || 0
+      amount_refunded: opts.amount_refunded || 0
     }
   }
 
@@ -112,6 +126,19 @@ const buildTransactionDetails = (opts = {}) => {
   if (opts.metadata) data.metadata = opts.metadata
 
   return data
+}
+
+const buildRefundDetails = (opts = {}) => {
+  return {
+    gateway_account_id: opts.gateway_account_id || '1',
+    amount: opts.amount || 10,
+    state: buildChargeEventStateWithDefaults(opts),
+    reference: opts.reference || 'aaba9756-15af-4ab0-b2c2-8696bc47655e',
+    created_date: opts.created_date || '2019-08-20T14:39:49.536Z',
+    transaction_type: 'REFUND',
+    transaction_id: opts.transaction_id || '1b5kia0u28ll2ic4obv26r5e4h',
+    parent_transaction_id: opts.parent_transaction_id || 'puuhl0gu7egigin7oh9c75p4m1'
+  }
 }
 
 module.exports = {
@@ -189,12 +216,8 @@ module.exports = {
     }
 
     return {
-      getPactified: () => {
-        return pactRegister.pactify(data)
-      },
-      getPlain: () => {
-        return data
-      }
+      getPactified: () => pactRegister.pactify(data),
+      getPlain: () => data
     }
   },
   validTransactionRefundRequest: (opts = {}) => {
@@ -205,12 +228,8 @@ module.exports = {
     }
 
     return {
-      getPactified: () => {
-        return pactRegister.pactify(data)
-      },
-      getPlain: () => {
-        return data
-      }
+      getPactified: () => pactRegister.pactify(data),
+      getPlain: () => data
     }
   },
   invalidTransactionRefundResponse: (opts = {}) => {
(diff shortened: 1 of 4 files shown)


test/unit/clients/ledger_client/ledger_pact_test_provider.js
```diff
@@ -0,0 +1,35 @@
...
+  log: path.resolve(process.cwd(), 'logs', 'mockserver-integration.log'),
+  dir: path.resolve(process.cwd(), 'pacts'),
+  spec: 2,
+  pactfileWriteMode: 'merge'
+})
+module.exports = {
+  getPort: function () {
+    return port
+  },
+  interaction: function (pact) {
```
jankowiakmaria: `addInteraction` would be more consistent as a name of this function.

test/unit/clients/ledger_client/ledger_pact_test_provider.js
```diff
@@ -0,0 +1,35 @@
+'use strict'
+
+// NPM dependencies
+const Pact = require('pact')
+
+// Custom dependencies
+const path = require('path')
+const port = parseInt(process.env.LEDGER_URL.match(/\d+(\.\d+)?$/g)[0], 10)
```
jankowiakmaria: In other pact tests we set random port. Shouldn't we be doing the same for ledger? Or is there anything special about it?

test/fixtures/ledger_transaction_fixtures.js
```diff
@@ -43,6 +43,20 @@ const buildChargeEventStateWithDefaults = (opts = {}) => {
   return state
 }
 
+const buildPaymentEvents = (opts = {}) => {
+  let events = []
+  if (opts.payment_states) {
+    opts.payment_states.forEach(paymentState => {
+      let statuses = {}
```
jankowiakmaria: This is an event, so it would be better to name it this way.

This could be more readable:
```js
const event = {
    state: buildChargeEventStateWithDefaults(paymentState),
    resource_type: 'PAYMENT',
    data: {}
}
events.push(event)
```
or even:
```js
events.push({
    state: buildChargeEventStateWithDefaults(paymentState),
    resource_type: 'PAYMENT',
    data: {}
})
```

test/fixtures/ledger_transaction_fixtures.js
```diff
@@ -91,7 +105,7 @@ const buildTransactionDetails = (opts = {}) => {
       status: opts.refund_summary_status || 'unavailable',
       user_external_id: opts.user_external_id || null,
       amount_available: opts.refund_summary_available || 20000,
-      amount_submitted: opts.refund_summary_submitted || 0
+      amount_refunded: opts.refund_summary_submitted || 0
```
jankowiakmaria: Maybe the option should be renamed as well (unless it's not ledger test specific).

test/fixtures/ledger_transaction_fixtures.js
```diff
@@ -114,6 +128,22 @@ const buildTransactionDetails = (opts = {}) => {
   return data
 }
 
+const buildRefundDetails = (opts = {}) => {
+  const state = buildChargeEventStateWithDefaults(opts)
+  const data = {
```
jankowiakmaria: return the object instead

test/fixtures/ledger_transaction_fixtures.js
```diff
@@ -114,6 +128,22 @@ const buildTransactionDetails = (opts = {}) => {
...
+    gateway_account_id: opts.gateway_account_id || '1',
+    amount: opts.amount || 10,
+    state: state,
```
jankowiakmaria: to simplify
```suggestion
    state: buildChargeEventStateWithDefaults(opts),
```

test/fixtures/ledger_transaction_fixtures.js
```diff
@@ -219,12 +241,47 @@ module.exports = {
...
-      getPlain: () => {
-        return data
+      getPactified: () => pactRegister.pactify(data),
+      getPlain: () => data
+    }
+  },
+  validTransactionEventsResponse: (opts = {}) => {
+    const data = {
+      transaction_id: opts.transaction_id || 'ht439nfg2l1e303k0dmifrn4fc',
+      events: (opts.payment_states) ? buildPaymentEvents(opts) : buildChargeEventStateWithDefaults(opts)
```
jankowiakmaria: Is it correct?
`buildChargeEventStateWithDefaults` returns a state object, not an array of events.

test/fixtures/ledger_transaction_fixtures.js
```diff
@@ -219,12 +241,47 @@ module.exports = {
...
+    }
+
+    return {
+      getPactified: () => pactRegister.pactify(data),
+      getPlain: () => data
+    }
+  },
+  validTransactionSearchResponse: (opts = {}) => {
+    let results = []
```
jankowiakmaria: `transactionSearchResult` or something more meaningful would be better.

test/fixtures/ledger_transaction_fixtures.js
```diff
@@ -219,12 +241,47 @@ module.exports = {
...
+      transaction.gateway_account_id = opts.gateway_account_id
+      if (transaction.type === 'payment') {
+        transaction.includeCardDetails = true
+        transaction.includeRefundSummary = true
+        transaction.includeSettlementSummary = true
+        result = buildTransactionDetails(transaction)
+        results.push(result)
+      } else if (transaction.type === 'refund') {
+        result = buildRefundDetails(transaction)
+        results.push(result)
```
jankowiakmaria: current version doesn't increase readability so it might as well be:
```suggestion
        results.push(buildRefundDetails(transaction))
```

test/fixtures/ledger_transaction_fixtures.js
```diff
@@ -219,12 +241,47 @@ module.exports = {
...
+    let results = []
+    opts.transactions.forEach(transaction => {
+      let result
+      transaction.gateway_account_id = opts.gateway_account_id
+      if (transaction.type === 'payment') {
+        transaction.includeCardDetails = true
+        transaction.includeRefundSummary = true
+        transaction.includeSettlementSummary = true
+        result = buildTransactionDetails(transaction)
+        results.push(result)
```
jankowiakmaria: current version doesn't increase readability so it might as well be:
```suggestion
        results.push(buildTransactionDetails(transaction))
```

test/fixtures/ledger_transaction_fixtures.js
```diff
@@ -219,12 +241,47 @@ module.exports = {
...
+
+    return {
+      getPactified: () => pactRegister.pactify(data),
+      getPlain: () => data
+    }
+  },
+  validTransactionSearchResponse: (opts = {}) => {
+    let results = []
+    opts.transactions.forEach(transaction => {
+      let result
```
jankowiakmaria: `transaction` (or something else that's more meaningful given the transaction is already taken)

test/unit/clients/ledger_client/ledger_pact_test_provider.js
SandorArpa: In other clients the port is configurable. After a discussion with Steven we decided that instead of refactoring ledger's client and use random ports, we add a 'wrapper' that uses ledger client's port.

test/fixtures/ledger_transaction_fixtures.js
SandorArpa: Everything is called either xxxRequest or xxxResponse thus I believe it'd confusing naming this anything else

PR #1616, opened 2019-08-29
Support non-HTTPS URLs for confirmation pages during testing

rhowe-gds:
## WHAT
During testing we sometimes disable HTTPS but we require that confirmation page
URLs are HTTPS. Unfortunately, the `DISABLE_INTERNAL_HTTPS` environment
variable is set to true during tests and so if we key off that, we'll break the
test which checks for validation that URLs are HTTPS. Add an extra environment
variable to use during tests - `ALLOW_HTTP_CONFIRMATION_PAGES` to allow us to
bypass this validation check.

### DIFF
--- a/app/browsered/field-validation-checks.js
+++ b/app/browsered/field-validation-checks.js
@@ -15,6 +15,7 @@ const validationErrors = {
   currency: 'Choose an amount in pounds and pence using digits and a decimal point. For example “10.50”',
   phoneNumber: 'Must be a 11 digit phone number',
   validEmail: 'Please use a valid email address',
+  isURL: 'Please use a valid URL',
   isHttps: 'URL must begin with https://',
   isAboveMaxAmount: `Choose an amount under £${MAX_AMOUNT.toLocaleString()}`,
   isPasswordLessThanTenChars: 'Choose a Password of 10 characters or longer',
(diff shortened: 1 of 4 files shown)


app/browsered/field-validation-checks.js
```diff
@@ -61,11 +63,7 @@ exports.isPhoneNumber = function (value) {
 }
 
 exports.isHttps = function (value) {
-  if (value.substr(0, 8) !== 'https://') {
-    return validationErrors.isHttps
-  } else {
-    return false
-  }
+  return new url.URL(value).protocol === 'https:'
```
jonheslop:
```suggestion
  return new url.URL(value).protocol !== 'https:' ? validationErrors.isHttps : false
```

app/controllers/test_with_your_users/submit_controller.js
```diff
@@ -27,7 +27,7 @@ module.exports = (req, res) => {
     req.flash('genericError', `<h2>Use valid characters only</h2> ${isCurrency(paymentAmountInPence)}`)
   } else if (isAboveMaxAmount(penceToPounds(paymentAmountInPence))) {
     req.flash('genericError', `<h2>Enter a valid amount</h2> ${isAboveMaxAmount(penceToPounds(paymentAmountInPence))}`)
-  } else if (!confirmationPage || isHttps(confirmationPage)) {
+  } else if (!confirmationPage || (process.env.DISABLE_INTERNAL_HTTPS !== 'true' && isHttps(confirmationPage))) {
```
rhowe-gds: Oh no, `DISABLE_INTERNAL_HTTPS` is set to `true` in the tests

rhowe-gds:
I give up on getting this to work

PR #1622, opened 2019-09-02
PP-5611 Remove Partial Reference Pact Test

RoryMalcolm:
- Removes the pact test that matches with a partial reference on search


RoryMalcolm:
> I assume you've made the change in connector. A link to that PR would be nice!

https://github.com/alphagov/pay-connector/pull/1667 - Sorry, didn't know etiquette!

PR #1627, opened 2019-09-03
PP-5537 add get events and search ledger transactions pact test

SandorArpa:
## WHAT

- add pact tests
- create a pact provider helper class. This is needed so the tests can
  share the pre-configured ledger_client port.
- update transaction fixtures


SandorArpa:
superseded

PR #1631, opened 2019-09-06
bau: Send analytics to cross-gov Google Analytics

oswaldquek:
See https://docs.google.com/document/d/1Scq-axCUknouQhKkf5ppqcqljeWSo62njL4p9qGcIjo/edit# to help verify this PR is correct.

### DIFF
--- a/app/views/includes/head.njk
+++ b/app/views/includes/head.njk
@@ -14,9 +14,15 @@
   ga('set', 'displayFeaturesTask', null);
   ga('set', 'transport', 'beacon');
 
+  ga('create', 'UA-145652997-1', 'auto', 'govuk_shared', {'allowLinker': true});
+  ga('govuk_shared.require', 'linker');
+  ga('govuk_shared.linker.set', 'anonymizeIp', true);
+  ga('govuk_shared.linker:autoLink', [‘www.gov.uk’]);
+
   var filtered_page_path = document.location.pathname + document.location.search.replace(/&email=[\w%.-]*/, '');
   ga('set', 'page', filtered_page_path);
   ga('send', 'pageview');
+  ga('govuk_shared.send', 'pageview');
 
 </script>
 <!-- End Google Analytics -->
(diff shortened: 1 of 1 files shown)


jonheslop:
Looks legit although I am confused about the `autoLink` bit, in the docs it says

> This field should include any domains to be included in the cross domain tracking. This should always include ‘www.gov.uk’ plus any other services that your service links to. For example, it could be:
> analytics.js: `[‘www.gov.uk’, ‘another.service.gov.uk’, ‘something.service.gov.uk’]`

And then it says 

> Note that the content of this field needs to be very specific as it will match subdomains automatically. For example if ‘gov.uk’ was included, it would match any subdomain of gov.uk. 

So in the example it explicitly mentions some domains that are all subdomains of `gov.uk` and it says that subdomains are done automatically?

So what you have done seems correct but I’d sorta say the docs are contradictory?

oswaldquek:
@jonheslop I think it just means if we put 'gov.uk' instead of 'www.gov.uk' it (the x-gov GA?) will match subdomains automatically. I'm not sure what this means though...does it mean other GAs on subdomains with .gov.uk will see our page visits?

oswaldquek:
Will try to get someone with a better understanding to review

app/views/includes/head.njk
```diff
@@ -14,9 +14,15 @@
   ga('set', 'displayFeaturesTask', null);
   ga('set', 'transport', 'beacon');
 
+  ga('create', 'UA-145652997-1', 'auto', 'govuk_shared', {'allowLinker': true});
+  ga('govuk_shared.require', 'linker');
+  ga('govuk_shared.linker.set', 'anonymizeIp', true);
+  ga('govuk_shared.linker:autoLink', [‘www.gov.uk’]);
```
jonheslop:
```suggestion
  ga('govuk_shared.linker:autoLink', ['www.gov.uk']);
```

PR #1655, opened 2019-09-23
PP-5657 Map Payment fields to Refunds

kbottla:
## WHAT 
- When searching transactions, payment related fields (card details, reference, gateway_transaction_id, and so on) are displayed for refund transaction.
  Ledger now returns payment transaction as nested object for refund

### DIFF
--- a/app/services/clients/utils/ledger_legacy_connector_parity.js
+++ b/app/services/clients/utils/ledger_legacy_connector_parity.js
@@ -13,7 +13,21 @@ const legacyConnectorTransactionParity = (transaction) => {
   if (transaction.refund_summary && transaction.refund_summary.amount_refunded) {
     transaction.refund_summary.amount_submitted = transaction.refund_summary.amount_refunded
   }
+
   transaction.charge_id = transaction.transaction_id
+
+  if (transaction.transaction_type && transaction.transaction_type.toLowerCase() === 'refund') {
+    if (transaction.parent_transaction !== undefined && transaction.parent_transaction !== null) {
+      let charge = transaction.parent_transaction
+      transaction.charge_id = charge.transaction_id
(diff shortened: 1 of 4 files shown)


app/services/clients/utils/ledger_legacy_connector_parity.js
```diff
@@ -13,7 +13,21 @@ const legacyConnectorTransactionParity = (transaction) => {
   if (transaction.refund_summary && transaction.refund_summary.amount_refunded) {
     transaction.refund_summary.amount_submitted = transaction.refund_summary.amount_refunded
   }
+
   transaction.charge_id = transaction.transaction_id
+
+  if (transaction.transaction_type && transaction.transaction_type.toLowerCase() === 'refund') {
+    if (transaction.parent_transaction !== undefined && transaction.parent_transaction !== null) {
+      let charge = transaction.parent_transaction
```
sfount: As all of the keys are the same, I think this would be equivalent: 
```js
Object.assign(transaction, transaction.parent_transaction) 
```
(except that it would be done before the `transaction.charge_id = transaction.transaction_id`)

There are some unit tests for the parity transform in the confusingly named https://github.com/alphagov/pay-selfservice/blob/master/test/unit/clients/ledger_client/legacy_connector_parity_util_test.js -- having a quick test here would make sure this has the same behaviour.

sfount: Just noticed that there is a test in the referenced file 🤦‍♂ 👍

kbottla: Not all keys are same. Would only want specific values copied to `refund` transaction. Rest of them (state, created_date, amount, issueBy) will stay same as they are specific to refund

PR #1666, opened 2019-10-07
Add Hadolint configuration from pay-code-analysis-config

alexbishop1:
🎉


alexbishop1:
Following discussions, the PMD config has been removed from this PR. We can add it back if we ever make it do anything useful for JavaScript apps.

PR #1680, opened 2019-10-18
PP-5716 Switch to structured logging

stephencdaly:
- Bump Winston to v3
- Log in JSON format, including the standard keys included for all log lines

- Also remove approximate package version numbers (`^`), as this was causing some npm install issues.

`^` allows npm to install an approximate package, this has caused issues when sharing package lock and is resolved by dependabot updating packages manually. Remove all approximate version numbers.


app/controllers/static_controller.js
```diff
@@ -1,7 +1,7 @@
-var logger = require('winston')
+const logger = require('../utils/logger')(__filename)
```
danworth: 👍

app/middleware/get_gateway_account.js
```diff
@@ -28,7 +28,7 @@ module.exports = function (req, res, next) {
         next()
       })
       .catch(err => {
-        winston.error(`${req.correlationId} - Error when attempting to retrieve direct debit gateway account: ${err}`)
+        logger.error(`${req.correlationId} - Error when attempting to retrieve direct debit gateway account: ${err}`)
```
danworth: 👍

app/services/clients/old_base_client.js
```diff
@@ -6,7 +6,6 @@ const https = require('https')
 const http = require('http')
 const urlParse = require('url').parse
 const _ = require('lodash')
-const logger = require('winston')
```
danworth: Looks like it was simply unused hence no replacement

server.js
```diff
@@ -132,7 +132,7 @@ function initialiseErrorHandling (app) {
 function listen () {
   const app = initialise()
   app.listen(port)
-  logger.log('Listening on port ' + port)
+  logger.info('Listening on port ' + port)
```
danworth: 👍 lolz

start.js
```diff
@@ -1,15 +1,15 @@
...
-  var path = require('path')
-  var fs = require('fs')
-  var logger = require('winston')
-  var throng = require('throng')
-  var server = require('./server')
-  var environment = require('./app/services/environment')
-  var pidFile = path.join(__dirname, '/.start.pid')
-  var fileOptions = { encoding: 'utf-8' }
-  var pid
+  const path = require('path')
```
danworth: `const` `const` `const` `const` `const` 💯

PR #1689, opened 2019-10-21
PP-5603 switch cypress tests to ledger for transaction searches

SandorArpa:
## WHAT
- switch transaction_search_spec to use ledger data
- switch transaction_details_spec to use ledger data
- fix: get_query_string_for_params -> delete cardbrand when
  flattenCardBrandsParam is true


test/cypress/integration/transactions/transaction_details_spec.js
```diff
@@ -20,45 +25,50 @@ function formatDate (date) {
...
+    finished: false,
+    status: 'created'
+  },
+  resource_type: 'PAYMENT',
+  event_type: 'PAYMENT_CREATED',
+  timestamp: '2019-09-18T10:06:17.152Z',
+  data: {}
+}]
+
+function defaultTransactionDetails (opts = {}) {
```
sfount: One small thing, this method feels like a fixture at the level of the `spec` (test), at this level we are really just providing data that the fixture will then build. 

The advantage of this is that the fixture is then also guaranteed against pacts through contracts with other services. 

Basically I think instead of receiving an opts this can just accept one or two configurations and be over-ridden by tests if the data needs to be different. As this logic is done by the fixture I don't think it need to happen here too.

PR #1691, opened 2019-10-24
PP-5755 Add Sentry

danworth:
Send error level log messages to Sentry. Using the log level to
control when messages are sent to Sentry seems intuitive and since we now
require a common logger throughout the application it offers a
convenient point to add Sentry.

Unhandled exceptions will be caught by the Sentry error middleware added
within server.js as per the guidlines when using Sentry with Express.

When initialising Sentry a beforeSend filter is added to remove the
request element from the event if it exists. The request element contains the
data from a failed request and this can include sensitive information such
as card numbers. We may want to introduce more refined filtering to allow
non-sensitive information from a failed request to go to Sentry at a
later date.

app/utils/sentry.js and more elements of app/utils/logger.js can
probably move to pay-js-commons once we're happy with the implementation.

## HOW 
Local testing results sent to Sentry can be viewed here:
https://pay-sentry.cloudapps.digital/sentry/selfservice/?environment=Dan_Testing_SELFSERVICE

### DIFF
--- a/app/utils/sentry.js
+++ b/app/utils/sentry.js
@@ -1,0 +1,32 @@
+const Sentry = require('@sentry/node')
+
+function initialiseSentry () {
+  Sentry.init({
+    dsn: process.env.SENTRY_DSN,
+    environment: process.env.ENVIRONMENT,
+    beforeSend (event) {
+      if (event.request) {
+        delete event.request // This can include sensitive data such as card numbers
+      }
+      return event
+    }
+  })
+  return Sentry
+}
+
+const addSentryToErrorLevel = originalLogger => {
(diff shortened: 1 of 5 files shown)


app/utils/sentry.js
```diff
@@ -0,0 +1,32 @@
...
+      if (event.request) {
+        delete event.request // This can include sensitive data such as card numbers
+      }
+      return event
+    }
+  })
+  return Sentry
+}
+
+const addSentryToErrorLevel = (originalLogger) => {
```
mrlumbu: total nitpick, but no need for brackets around `(originalLogger)`

danworth: nitpick away 👍fixed

PR #1702, opened 2019-10-30
PP-5427 switch activity dashboard to ledger

SandorArpa:
## WHAT
PP-5427 switch activity dashboard to ledger

- refactor dashboard-activity-controller to use ledger
- update relevant tests
- refactor cypress tests
- code clean up: remove 'custom' date range (it has been removed from the dashboard ui)

### DIFF
--- a/app/controllers/dashboard/dashboard-activity-controller.js
+++ b/app/controllers/dashboard/dashboard-activity-controller.js
@@ -10,7 +10,7 @@ const getNamespace = require('continuation-local-storage').getNamespace
 const logger = require('../../utils/logger')(__filename)
 const response = require('../../utils/response').response
 const CORRELATION_HEADER = require('../../utils/correlation_header').CORRELATION_HEADER
-const ConnectorClient = require('../../services/clients/connector_client').ConnectorClient
+const LedgerClient = require('../../services/clients/ledger_client')
 const { isADirectDebitAccount } = require('../../services/clients/direct_debit_connector_client.js')
 const auth = require('../../services/auth_service.js')
 const { datetime } = require('@govuk-pay/pay-js-commons').nunjucksFilters
@@ -29,9 +29,6 @@ const {
   LIVE,
   DENIED
 } = require('../../models/go-live-stage')
-
-const connectorClient = () => new ConnectorClient(process.env.CONNECTOR_URL)
-const getTimespanDays = (fromDateTime, toDateTime) => moment(toDateTime).diff(moment(fromDateTime), 'days')
 
 // Constants
 const clsXrayConfig = require('../../../config/xray-cls')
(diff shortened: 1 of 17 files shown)


app/controllers/dashboard/dashboard-activity-controller.js
```diff
@@ -180,9 +164,10 @@ module.exports = (req, res) => {
   }
 }
 
-function getTransactionDateRange (period, customRange) {
-  let fromDateTime
-  let toDateTime = moment().tz('Europe/London').format() // Today is the default
+function getTransactionDateRange (period) {
```
sfount: I'm a little concerned about such a big change to this method without any unit tests validating behaviour, are there some that I've just missed?

Also it looks like this removes any option for custom dates, is that definitely not used by anything?

app/services/clients/ledger_client.js
```diff
@@ -65,11 +65,16 @@ const allTransactionPages = async function allTransactionPages (gatewayAccountId
...
 }
 
-const transactionSummary = async function transactionSummary (gatewayAccountId, fromDate, toDate, options = {}) {
-  const path = `/v1/report/transactions-summary?account_id=${gatewayAccountId}&from_date=${fromDate}&to_date=${toDate}`
+const transactionSummary = function transactionSummary (gatewayAccountId, fromDate, toDate, options = {}) {
+  const path = '/v1/report/transactions-summary'
   const configuration = Object.assign({
     url: path,
-    description: 'Transaction summary for a given gateway account ID',
+    qs: {
```
sfount: 👍 really nice improvement using `qs`. Much more readable!

app/services/clients/utils/ledger_legacy_connector_parity.js
```diff
@@ -40,6 +40,9 @@ const legacyConnectorTransactionParity = (transaction) => {
 }
 
 const legacyConnectorTransactionSummaryParity = (transactionSummaryResult) => {
+  if (transactionSummaryResult === 'undefined') {
```
sfount: This will test if the returned value is the exact string 'undefined', if this is testing if this doesn't exist it would need to test the type of the object. 

In this case a falsy comparison would probably work (as it expects an object otherwise, an object will always be truthy)
```js 
if (!transactionSummaryResult) { 
  return {}
}
```

Or to follow the current pattern
```js
if (typeof transactionSummaryResult === 'undefined') { 
  return {}
}
```

SandorArpa: 👍

app/controllers/dashboard/dashboard-activity-controller.js
SandorArpa: > I'm a little concerned about such a big change to this method without any unit tests validating behaviour, are there some that I've just missed?
`dashboard-activity-controller.test` 
> Also it looks like this removes any option for custom dates, is that definitely not used by anything?
It was going to be a feature 2-3 years ago but it was halted. The decision was that if the user wants specific dates then they should use the transaction page. I had a discussion with @jonheslop about it.

SandorArpa: > I'm a little concerned about such a big change to this method without any unit tests validating behaviour, are there some that I've just missed?
> 
> Also it looks like this removes any option for custom dates, is that definitely not used by anything?
 
It's tested under `dashboard-activity-controller.test`

The customRange was going to be a feature about 2-3 years ago, but it was halted by the designers. The decision was that if the user wants specific dates then they should use the transaction page. I had a discussion with @jonheslop about it.

PR #1733, opened 2019-11-11
PP-5760 Specify status code when rendering error page

stephencdaly:
When we display an error page for a BAU error rather than an internal error, specify a non-500 status code so that it doesn't send a 500 status by default. This also prevents an error from being logged for these cases.

The first commit in the PR has some refactoring to improve the way we require the response methods.

I'd recommend looking at the 2 commits separately, as it will be easier to see the changes.

### DIFF
--- a/app/controllers/create_service_controller.js
+++ b/app/controllers/create_service_controller.js
@@ -4,11 +4,11 @@
 const lodash = require('lodash')
 
 // Local Dependencies
-const responses = require('../utils/response')
+const { renderErrorView, response } = require('../utils/response')
 const paths = require('../paths')
 const serviceService = require('../services/service_service')
 const userService = require('../services/user_service')
-const {validateServiceName} = require('../utils/service_name_validation')
+const { validateServiceName } = require('../utils/service_name_validation')
 
 exports.get = (req, res) => {
   let pageData = lodash.get(req, 'session.pageData.createServiceName')
@@ -24,7 +24,7 @@ exports.get = (req, res) => {
   pageData.submit_link = paths.serviceSwitcher.create
   pageData.my_services = paths.serviceSwitcher.index
 
-  return responses.response(req, res, 'services/add_service', pageData)
+  return response(req, res, 'services/add_service', pageData)
 }
 
 exports.post = (req, res) => {
@@ -36,7 +36,7 @@ exports.post = (req, res) => {
   const validationErrorsCy = validateServiceName(serviceNameCy, 'service_name_cy', false)
   if (Object.keys(validationErrors).length || Object.keys(validationErrorsCy).length) {
     lodash.set(req, 'session.pageData.createServiceName', {
-      errors: {...validationErrors, ...validationErrorsCy},
+      errors: { ...validationErrors, ...validationErrorsCy },
       current_name: serviceName,
       current_name_cy: serviceNameCy
     })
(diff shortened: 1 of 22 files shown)


app/controllers/forgotten_password_controller.js
```diff
@@ -40,8 +40,8 @@ module.exports = {
   newPasswordGet: (req, res) => {
     const id = req.params.id
     const render = (user) => {
-      if (!user) return errorView(req, res)
-      res.render('forgotten_password/new_password', {id: id})
+      if (!user) return renderErrorView(req, res)
```
danworth: Somewhere in our style guide is to always use `{}` might be nice to add them since its updating the line but not necessary for approval.

app/controllers/create_service_controller.js
```diff
@@ -4,11 +4,11 @@
 const lodash = require('lodash')
 
 // Local Dependencies
-const responses = require('../utils/response')
+const { renderErrorView, response } = require('../utils/response')
```
danworth: This is nice, makes it easier to follow down below 👍

app/controllers/service_roles_update_controller.js
```diff
@@ -60,7 +60,7 @@ module.exports = {
     }
 
     if (req.user.externalId === externalUserId) {
-      renderErrorView(req, res, 'Not allowed to update self permission')
+      renderErrorView(req, res, 'Not allowed to update self permission', 400)
```
danworth: Maybe 403 forbidden?

app/controllers/partnerapp/handle_gocardless_connect_get.js
```diff
@@ -43,7 +43,7 @@ function processPayload (req, res, getPayload) {
     .catch(err => {
       console.log(err)
       if (err.errorIdentifier === GO_CARDLESS_ACCOUNT_ALREADY_LINKED_TO_ANOTHER_ACCOUNT) {
-        renderErrorView(req, res, 'This GoCardless account is already connected to a GOV.UK Pay account. You’ll need to use a different account.')
+        renderErrorView(req, res, 'This GoCardless account is already connected to a GOV.UK Pay account. You’ll need to use a different account.', 400)
```
danworth: Maybe 409 conflict?

app/controllers/service_users_controller.js
```diff
@@ -142,7 +142,7 @@ module.exports = {
     const correlationId = req.correlationId
 
     if (userToRemoveExternalId === removerExternalId) {
-      renderErrorView(req, res, 'Not allowed to delete a user itself')
+      renderErrorView(req, res, 'Not allowed to delete a user itself', 400)
```
danworth: Maybe 403 or 409?

app/controllers/login/send-again-post-controller.js
```diff
@@ -18,6 +18,6 @@ module.exports = (req, res) => {
         logger.error(err)
       })
   } else {
-    renderErrorView(req, res, 'You do not use text messages to sign in')
+    renderErrorView(req, res, 'You do not use text messages to sign in', 400)
```
danworth: Maybe 409?

stephencdaly: agreed we'll leave this as 400

app/controllers/service_users_controller.js
stephencdaly: updated to 403

app/controllers/partnerapp/handle_gocardless_connect_get.js
stephencdaly: agreed we'd leave this as 400

app/controllers/service_roles_update_controller.js
stephencdaly: updated to 403

PR #1741, opened 2019-11-14
PP-5790 ability to add flex credentials

jonheslop:
- Added Flex credentials to Your PSP page
- Added Cypress tests for whole lot

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -225,6 +225,8 @@ module.exports.bind = function (app) {
 
   // YOUR PSP
   app.get(yourPsp.index, xraySegmentCls, permission('gateway-credentials:read'), getAccount, paymentMethodIsCard, yourPspController.getIndex)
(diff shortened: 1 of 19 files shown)


app/routes.js
```diff
@@ -225,6 +225,8 @@ module.exports.bind = function (app) {
 
   // YOUR PSP
   app.get(yourPsp.index, xraySegmentCls, permission('gateway-credentials:read'), getAccount, paymentMethodIsCard, yourPspController.getIndex)
+  app.get(yourPsp.flex, xraySegmentCls, permission('gateway-credentials:update'), getAccount, paymentMethodIsCard, yourPspController.getFlex)
```
RoryMalcolm: does this need to have update creds?

jonheslop: Well I guess its like, why show someone a page if they cant change it?
but yeah technically we could just protect the post route…
but that might be frustrating to the user

PR #1747, opened 2019-11-20
Bau govuk frontend 3.4 bump

jonheslop:
Update to the new GOV.UK Frontend and following their introduction of a new feature tackle our custom width pages in a different manner https://github.com/alphagov/govuk-frontend/releases/tag/v3.4.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "csrf": "3.1.x",
     "express": "4.17.x",
     "google-libphonenumber": "3.2.6",
-    "govuk-frontend": "3.3.0",
+    "govuk-frontend": "3.4.0",
     "http-proxy": "1.18.x",
     "https-proxy-agent": "3.0.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -84,7 +84,7 @@
     "csrf": "3.1.x",
     "express": "4.17.x",
     "google-libphonenumber": "3.2.6",
-    "govuk-frontend": "3.3.0",
+    "govuk-frontend": "^3.4.0",
```
sfount: I think we're moving away from using wildcard version in package json in favour of depndabot updating these to fixed version. 

It means `npm install` will always give you the same package in all environments and won't ever try to update things.

jonheslop: Ah ok will amend, I had ran `npm i -S govuk-frontend@latest` guess I should have used `npm i -S govuk-frontend@3.4.0`

PR #1757, opened 2019-11-27
Fix capitalisation on add service button

quis:
GOV.UK style is to use _Sentence case_ not _Title Case_ for buttons etc.


alexbishop1:
Closing in favour of #1758 for the reasons explained in the code review.

PR #1770, opened 2019-12-04
BAU Remove reference from transaction details title

sfount:
* `reference` is being used by services to include personal information
like paying user name
* we send our page titles to google analytics
* patch the page title to use transaction ID (in our control) vs.
reference (out of our control)


sfount:
Superseded by https://github.com/alphagov/pay-selfservice/pull/1772

PR #1785, opened 2019-12-16
zd3866092: Redact entire email

oswaldquek:
We send analytics to the cross gov google analytics app too. They (GOV.UK) have some
script that flags when personal information isn't redacted. While we have
redacted the front part of the email (i.e. everything before the "@"), it's
possible the GOV.UK script is still flagging the emails even though the front
part is redacted. Removing the entire email might solve the problem.


heathd:
it would be better to extract a function and unit test the function than rely on manual testing

PR #1809, opened 2020-01-14
PP-5996 Add cypress test for changing user permissions

stephencdaly:
Replaces the `PermissionsIT.adminUserCanEditServiceRoleOfAnotherMemberOfTheTeam` end to end test

### DIFF
--- a/test/cypress/integration/manage-team-members/edit_permissions_spec.js
+++ b/test/cypress/integration/manage-team-members/edit_permissions_spec.js
@@ -1,0 +1,83 @@
+'use strict'
+
+const { getUserWithServiceRoleStubOpts } = require('../../utils/user_stubs')
+
+const SERVICE_EXTERNAL_ID = 'service_abc_123'
+const AUTHENTICATED_USER_ID = 'authenticated-user-id'
+const EDITING_USER_ID = 'user-we-are-editing-id'
+
+describe('Edit service user permissions', () => {
+  beforeEach(() => {
+    // keep the same session for entire describe block
+    Cypress.Cookies.preserveOnce('session')
+    Cypress.Cookies.preserveOnce('gateway_account')
+
+    const authenticatedUserStubOpts = getUserWithServiceRoleStubOpts(AUTHENTICATED_USER_ID, 'logged-in-user@example.com', SERVICE_EXTERNAL_ID, 'admin')
+    const userWeAreEditingStubOpts = getUserWithServiceRoleStubOpts(EDITING_USER_ID, 'other-user@example.com', SERVICE_EXTERNAL_ID, 'admin')
+
+    cy.task('setupStubs', [
+      {
+        name: 'getUserSuccess',
+        opts: authenticatedUserStubOpts
+      },
+      {
+        name: 'getUserSuccess',
+        opts: userWeAreEditingStubOpts
+      },
+      {
+        name: 'getServiceUsersSuccess',
+        opts: {
+          serviceExternalId: SERVICE_EXTERNAL_ID,
+          users: [
+            authenticatedUserStubOpts,
+            userWeAreEditingStubOpts
+          ]
+        }
+      },
+      {
+        name: 'getInvitedUsersSuccess',
+        opts: {
+          serviceExternalId: SERVICE_EXTERNAL_ID,
+          invites: []
+        }
+      },
+      {
+        name: 'postUpdateServiceRoleSuccess',
+        opts: {
+          role: 'view-and-refund',
+          external_id: EDITING_USER_ID,
+          serviceExternalId: SERVICE_EXTERNAL_ID
+        }
+      }
+    ])
+  })
+
+  it('should display team members page', () => {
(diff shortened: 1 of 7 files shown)


test/cypress/integration/manage-team-members/edit_permissions_spec.js
```diff
@@ -0,0 +1,123 @@
...
+    cy.get('a').contains('Edit permissions').click()
+    cy.get('#role-admin-input').should('exist').should('have.attr', 'checked')
+  })
+
+  it('should update permission', () => {
+    cy.get('#role-view-and-refund-input').click()
+    cy.get('button').contains('Save changes').click()
+    cy.get('.flash-container--good').contains('Permissions have been updated')
+  })
+})
```
oswaldquek: should there not be another step where we go back to `/service/${SERVICE_EXTERNAL_ID}` and see the user who's permissions that was just edited appears in the right view-and-refund list?

test/cypress/integration/manage-team-members/edit_permissions_spec.js
```diff
@@ -0,0 +1,123 @@
...
+          users: [
+            authenticatedUserStubOpts,
+            userWeAreEditingStubOpts,
+            viewOnlyUserStubOpts,
+            viewAndRefundUserStubOpts
+          ]
+        }
+      },
+      {
+        name: 'getInvitedUsersSuccess',
```
oswaldquek: why do we need this `getInvitedUsersSuccess` block? they seem redundant as the user we're editing is the one with email address other-user@example.com

test/cypress/integration/manage-team-members/edit_permissions_spec.js
```diff
@@ -0,0 +1,123 @@
...
+describe('Edit service user permissions', () => {
+  beforeEach(() => {
+    // keep the same session for entire describe block
+    Cypress.Cookies.preserveOnce('session')
+    Cypress.Cookies.preserveOnce('gateway_account')
+
+    const authenticatedUserStubOpts = getUserStubOpts(AUTHENTICATED_USER_ID, 'logged-in-user@example.com', 'admin', 'Administrator')
+    const userWeAreEditingStubOpts = getUserStubOpts(EDITING_USER_ID, 'other-user@example.com', 'admin', 'Administrator')
+    const viewOnlyUserStubOpts = getUserStubOpts('view-only-user-id', 'view-only-user@example.com', 'view-only', 'View Only')
+    const viewAndRefundUserStubOpts = getUserStubOpts('view-and-refund-user-id', 'view-and-refund-user@example.com', 'view-and-refund', 'View And Refund')
```
oswaldquek: do we really need `viewOnlyUserStubOpts` and `viewAndRefundUserStubOpts` for this test? we're just testing we can change the permissions for the user that relates to `viewOnlyUserStubOpts`

test/cypress/plugins/stubs.js
```diff
@@ -111,6 +112,53 @@ module.exports = {
...
+            headers: {
+              'Content-Type': 'application/json'
+            },
+            body: aValidServiceUsersResponse
+          }
+        }]
+      }
+    ]
+  },
+  getInvitedUsersSuccess: (opts = {}) => {
```
oswaldquek: not sure if this is relevant to what is being tested here

test/cypress/integration/manage-team-members/edit_permissions_spec.js
stephencdaly: Because the tests are driven using stubs rather than any real interaction with adminusers, this would just check that a user with the view-and-refund permission is in the correct list which we're already checking

test/cypress/integration/manage-team-members/edit_permissions_spec.js
stephencdaly: maybe not, I just thought I should do some basic tests on the pages we're viewing as they don't already exist as cypress tests. I could split these out into different spec files if you think this makes sense.
or if you don't think its worthwhile at all i could remove

test/cypress/integration/manage-team-members/edit_permissions_spec.js
stephencdaly: this was to test the team members page, as thought we may as well, but as i said I could split this out if you don't think it makes sense as part of this spec

test/fixtures/invite_fixtures.js
```diff
@@ -8,12 +8,31 @@ const pactBase = require('./pact_base')
 // Global setup
 const pactInvites = pactBase()
 
+function buildInviteWithDefaults (opts) {
+  const data = _.defaults(opts, {
```
oswaldquek: what does `_.defaults` do?

stephencdaly: it constructs an object using the values for the keys in `opts` if they exist, otherwise uses the  values in the second object given. https://lodash.com/docs/4.17.15#defaults

PR #1813, opened 2020-01-16
PP-6008 Add ability to verify stubs have been called in Cypress tests

stephencdaly:
In Cypress tests, it is sometimes desired to verify that a HTTP request
to an external service is made to verify the behaviour of a task. The
default behaviour of Mountebank, which is the stub server used by
Cypress tests, is to return a 200 response with an empty body in
response to all requests a matching stub does not exist for. Therefore,
tests will only fail if a stub we expect to be called is not called if
we expect a non-200 response or the application uses the body of the
response in some way.

The example this is implemented for in this commit is adding a new
service. We want to verify that the appropriate calls are made to
adminusers and connector to perform the actions that add the service.
Without verifying these requests are made with the expected request
body, we cannot be confident that a service and gateway account would
have been created in the real world.

The verification that requests are made in the Cypress test in
combination with contract (Pact) tests for the endpoints called serves
as an adequate alternative to end-to-end tests.

Verification of the number of times a stub is called is done by running
mountebank with the `--debug` flag, which records the matches that are
made against each stub. After each test is run, we can then look at the
size of the matches array and compare against the expectation which is
stored in a new `verifyCalledTimes` property on the stubs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "lint": "./node_modules/.bin/standard --fix",
     "lint-sass": "./node_modules/.bin/sass-lint -v",
     "test": "npm run snyk-protect && rm -rf ./pacts && ./node_modules/mocha/bin/mocha '!(node_modules)/**/*+(.|_)+(test|tests)'.js",
-    "cypress:server": "mb | node --inspect -r dotenv/config start.js dotenv_config_path=test/cypress/test.env",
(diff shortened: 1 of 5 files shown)


stephencdaly:
I was also thinking it might be desirable in the future to always verify that stubs that are set up are called at least once, like how mockito works.
However, due to the fact that we often set up stubs in a `beforeEach` for all tests in a spec file which might not be used for each `it`, this may take a bit of editing of tests to achieve.

oswaldquek:
LGTM

PR #1821, opened 2020-01-20
PP-6008-Update-dd-service-name-cypress test

nimalank7:
Description:
- Replaces MyServicesIT.editServiceName_shouldSucceed_ForADirectDebitAccount in self-service end-to-end with Cypress test update_service_name.js
- Adds Cypress test for updating Welsh service name

### DIFF
--- a/test/cypress/integration/my-services/update_service_name.js
+++ b/test/cypress/integration/my-services/update_service_name.js
@@ -1,0 +1,116 @@
+'use strict'
+
+const { getUserStubWithServiceName, getGatewayAccountsStub } = require('../../utils/common_stubs')
+const authenticatedUserId = 'authenticated-user-id'
+const serviceExternalId = 'service-external-id'
+const newServiceName = 'Updated Service'
+const serviceName = {
+  en: 'My Service'
+}
+const welshServiceName = {
+  en: 'My Service',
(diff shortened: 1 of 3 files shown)


test/cypress/integration/my-services/update_service_name.js
```diff
@@ -0,0 +1,99 @@
...
+      setupStubs(serviceName)
+      cy.get('a').contains('Edit name').click()
+
+      cy.title().should('eq', 'Edit service name - GOV.UK Pay')
+      cy.get('#service-name').should('have.attr', 'value', 'My Service')
+      cy.get('#checkbox-service-name-cy').should('have.attr', 'aria-expanded', 'false')
+    })
+
+    it('should update service name to Updated Service', () => {
+      setUpStubWithWelshServiceName()
```
stephencdaly: I think this is ever so slightly confusing, as if we don't provide a `welshServiceName` it will setup the `getUserStubWithServiceName` to return an empty string for the `service_name` in the user response.

I think it would be better if this function just returned an object for the `patchUpdateServiceNameSuccess` which we then pass into the `setupStubs` call within the test.

nimalank7: Done

PR #1823, opened 2020-01-20
PP-6008-Add-delete-payment-link-cypress-test

nimalank7:
Description:
- Moves PrototypeProductPaymentIT.shouldDeleteATestProductPayment from selfservice endtoend as a cypress test

### DIFF
--- a/test/cypress/integration/payment-links/delete_payment_links_spec.js
+++ b/test/cypress/integration/payment-links/delete_payment_links_spec.js
@@ -1,0 +1,38 @@
+const { getUserStub, getGatewayAccountStub } = require('../../utils/common_stubs')
+const { getProductsStub, getProductByExternalIdStub, deleteProductStub } = require('../../utils/products_stubs')
+const userExternalId = 'a-user-id'
+const gatewayAccountId = 42
+const productExternalId = 'a-product-id'
+
+const product = {
+  external_id: productExternalId
+}
+
+describe('Should delete payment link', () => {
+  beforeEach(() => {
+    cy.setEncryptedCookies(userExternalId, gatewayAccountId)
+    cy.task('setupStubs', [
+      getUserStub(userExternalId, [gatewayAccountId]),
+      getGatewayAccountStub(gatewayAccountId, 'test', 'worldpay'),
+      getProductsStub([product], gatewayAccountId),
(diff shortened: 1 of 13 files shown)


test/cypress/integration/payment-links/delete_payment_links_spec.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const product = buildPaymentLinkOpts('Pay for a parking permit', 'example.com/my-service/pay-for-a-parking-permit', 'en')
+
+describe('Should delete payment link', () => {
+  beforeEach(() => {
+    cy.setEncryptedCookies(userExternalId, gatewayAccountId)
+    cy.task('setupStubs', [
+      commonStubs.getUserStub(userExternalId, [gatewayAccountId]),
+      commonStubs.getGatewayAccountStub(gatewayAccountId, 'test', 'worldpay'),
```
stephencdaly: i know this was copy+pasted but could we use destructuring for these imports too for consistency.

test/cypress/integration/payment-links/delete_payment_links_spec.js
```diff
@@ -0,0 +1,52 @@
...
+    language: language,
+    type: 'ADHOC',
+    links: [
+      {
+        'rel': 'friendly',
+        'method': 'GET',
+        'href': href
+      }
+    ]
+  }
```
stephencdaly: I think we don't have to provide any of these options apart from `external_id`, we can just use the defaults given to us by the fixture hopefully as `external_id` is the only thing used by any of our tests.

test/cypress/integration/payment-links/delete_payment_links_spec.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const product = buildPaymentLinkOpts('Pay for a parking permit', 'example.com/my-service/pay-for-a-parking-permit', 'en')
```
stephencdaly: As we only call this function once, I don't think its worthwhile having a function at all, just make it a constant. I.e.
```
const product = { external_id: productExternalId }
```

test/cypress/integration/payment-links/delete_payment_links_spec.js
nimalank7: I just removed it but it broke the test

PR #1837, opened 2020-01-27
BAU: Make pact a dev dependency

richardTowers:
WHAT
----

In 2f5e2e06 we switched from the old version of pact to the new version,
but it looks to me like we accidentally (?) made it a production
dependency, when it should be a dev dependency (?).

This needs reviewing by someone who understands pact a bit more than I do so
they can confirm my suspicion that it should only be a dev dependency.

HOW
---

Run `rm -r node_modules` and `npm install --production`, observe that
pact doesn't get installed.

Confirm that all the tests still pass.


richardTowers:
Just going through closing my old PRs. It looks like this has been fixed in a separate commit on the default branch.

PR #1864, opened 2020-02-14
PP-6127 Map secrets to env vars

mrlumbu:
Map secrets from selfservice secret service to env vars.
Have removed gocardless env vars as they don't appear to
be used


manifest.yml
```diff
@@ -14,22 +14,25 @@ applications:
   command: npm start
```
rjbaker: Just needs binding to the `selfservice-secret-service`

mrlumbu: 👍

PR #1871, opened 2020-02-19
PP-6152 Add metadata to payment link page

sfount:
* resource controllers for GETting add metadata to payment link page
* initial template for column header and cell value form
* restrict access to pages and POST routes by `experimental_features_enabled` flag

### DIFF
--- a/app/controllers/payment-links/metadata/resource-controller.js
+++ b/app/controllers/payment-links/metadata/resource-controller.js
@@ -1,0 +1,10 @@
+const { response } = require('../../../utils/response.js')
+
+const addMetadataPage = function addMetadataPage (req, res, next) {
+  // @TODO(sfount) rethink how response works, get repsonse context from library and render _in_ the contrller
+  response(req, res, 'payment-links/metadata/edit-metadata')
+}
+
+module.exports = {
+  addMetadataPage
+}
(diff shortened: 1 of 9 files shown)


app/controllers/payment-links/metadata/resource-controller.js
```diff
@@ -0,0 +1,36 @@
+const { response } = require('../../../utils/response')
```
kbottla: Would be useful to have tests for this controller ?

app/paths.js
```diff
@@ -150,7 +150,10 @@ module.exports = {
     edit: '/create-payment-link/manage/edit/:productExternalId',
     editInformation: '/create-payment-link/manage/edit/information/:productExternalId',
     editReference: '/create-payment-link/manage/edit/reference/:productExternalId',
-    editAmount: '/create-payment-link/manage/edit/amount/:productExternalId'
+    editAmount: '/create-payment-link/manage/edit/amount/:productExternalId',
+    metadata: {
+      add: '/create-payment-link/manage/edit/:productExternalId/metadata'
```
kbottla: I guest path is `create-payment-link` due to existing convention ?

app/controllers/payment-links/metadata/resource-controller.js
sfount: I think so! it's currently used by every controller though and don't think it fits into the scope of this PR. We could do this is a BAU improvement (although I would like to stop passing `req`, `res` etc. down into this library and refactor it to just provide a context).

app/paths.js
sfount: yeah, maybe we should rethink that once we get all the 'management' pages in

PR #1881, opened 2020-02-25
PP-6164 Show All Service Transactions

rorymalcolm:
- Shows transactions for all services a user has permissions to view them for to a user via a link in the 'My Services' page if a feature flag is enabled.

### DIFF
--- a/app/controllers/all-service-transactions/get-controller.js
+++ b/app/controllers/all-service-transactions/get-controller.js
@@ -1,0 +1,57 @@
+'use strict'
+
+// NPM dependencies
+const _ = require('lodash')
+
+const { response } = require('../../utils/response')
+const { ConnectorClient } = require('../../services/clients/connector_client.js')
+const serviceService = require('../../services/service_service')
+const transactionService = require('../../services/transaction_service')
+const { buildPaymentList } = require('../../utils/transaction_view.js')
+const { getFilters, describeFilters } = require('../../utils/filters.js')
+const router = require('../../routes.js')
+const states = require('../../utils/states')
+const client = new ConnectorClient(process.env.CONNECTOR_URL)
+
+const { CORRELATION_HEADER } = require('../../utils/correlation_header.js')
+
+module.exports = async (req, res) => {
+  const correlationId = req.headers[CORRELATION_HEADER] || ''
+  const servicesRoles = _.get(req, 'user.serviceRoles', [])
+  const filters = getFilters(req)
+  try {
+    const aggregatedGatewayAccountIds = servicesRoles
+      .flatMap(servicesRole => servicesRole.service.gatewayAccountIds)
+      .reduce((accumulator, currentValue) => accumulator.concat(currentValue), [])
+    const aggregatedGatewayAccounts = await serviceService.getGatewayAccounts(aggregatedGatewayAccountIds, req.correlationId)
+    const csList = _.flatMap(aggregatedGatewayAccounts, x => x.id).join(',')
+    const searchResultOutput = await transactionService.search(csList, filters.result)
+    const cardTypes = await client.getAllCardTypesPromise(correlationId)
+    const model = buildPaymentList(searchResultOutput, cardTypes, null, filters.result, router.paths.allServiceTransactions.download)
+    model.search_path = router.paths.allServiceTransactions.index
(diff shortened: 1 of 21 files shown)


app/views/all_service_transactions/filter.njk
```diff
@@ -0,0 +1,217 @@
+
+<div class="transactions-filter govuk-grid-row">
```
sfount: Just a reminder to myself (as discussed), if any of this filter/ table template was re-usable between the transactions and all transactions page (hopefully all of it with a condition or two for specific changes) it would mean we don't have to maintain two different pages. Looking good so far!

app/controllers/all-service-transactions/get-controller.js
```diff
@@ -0,0 +1,63 @@
...
+          if (_.has(filters.result, 'brand')) {
+            model.cardBrands.forEach(brand => {
+              brand.selected = filters.result.brand.includes(brand.value)
+            })
+          }
+          model.filterRedirect = router.paths.allServiceTransactions.index
+          return response(req, res, 'all_service_transactions/index', model)
+        })
+      } catch (err) {
+        logger.error(err)
```
SandorArpa: What is going to happen on the user's end if this thrown an error?

app/controllers/all-service-transactions/get-controller.js
```diff
@@ -0,0 +1,63 @@
...
+const { CORRELATION_HEADER } = require('../../utils/correlation_header.js')
+
+module.exports = async (req, res) => {
+  const correlationId = req.headers[CORRELATION_HEADER] || ''
+  const servicesRoles = _.get(req, 'user.serviceRoles', [])
+  const filters = getFilters(req)
+  const aggregatedGatewayAccountIds = servicesRoles
+    .flatMap(servicesRole => servicesRole.service.gatewayAccountIds)
+    .reduce((accumulator, currentValue) => accumulator.concat(currentValue), [])
+  serviceService.getGatewayAccounts(aggregatedGatewayAccountIds, req.correlationId)
```
stephencdaly: use `await`

app/controllers/all-service-transactions/get-controller.js
rorymalcolm: I've changed this to render the error view instead in response

app/controllers/all-service-transactions/get-controller.js
rorymalcolm: no longer async

app/controllers/all-service-transactions/get-controller.js
```diff
@@ -0,0 +1,62 @@
...
+          if (_.has(filters.result, 'brand')) {
+            model.cardBrands.forEach(brand => {
+              brand.selected = filters.result.brand.includes(brand.value)
+            })
+          }
+          model.filterRedirect = router.paths.allServiceTransactions.index
+          return response(req, res, 'all_service_transactions/index', model)
+        })
+      } catch (err) {
+        response.renderErrorView(req, res, 'Unable to fetch trasnaction information')
```
SandorArpa: Trasnaction? ;)

app/controllers/all-service-transactions/get-controller.js
```diff
@@ -0,0 +1,58 @@
...
+  const servicesRoles = _.get(req, 'user.serviceRoles', [])
+  const filters = getFilters(req)
+  try {
+    const aggregatedGatewayAccountIds = servicesRoles
+      .flatMap(servicesRole => servicesRole.service.gatewayAccountIds)
+      .reduce((accumulator, currentValue) => accumulator.concat(currentValue), [])
+    const aggregatedGatewayAccounts = await serviceService.getGatewayAccounts(aggregatedGatewayAccountIds, req.correlationId)
+    const csList = _.flatMap(aggregatedGatewayAccounts, x => x.id).join(',')
+    const searchResultOutput = await transactionService.search(csList, filters.result)
+    client.getAllCardTypes({ correlationId }, allCards => {
```
stephencdaly: can we use `getAllCardTypesPromise`?  does it return the same?

rorymalcolm: done

stephencdaly:
I think the service switching stuff should be in a different PR to make this easier to review, we should also have tests alongside the code changes.

PR #1882, opened 2020-02-26
PP-6156 Add metadata to page for individual payment link

alexbishop1:
On the page for each individual payment link, add a section for reporting columns (metadata associated with the payment link), complete with links to add a new reporting column and edit the
existing ones. Reporting columns are sorted alphabetically (case-insensitively) by metadata key.

The reporting columns section is only shown if the `experimentalFeaturesEnabled` flag is set on the service.

While the reporting columns section uses the [summary list component from the GOV.UK Design System](https://design-system.service.gov.uk/components/summary-list/), it uses HTML rather than the Nunjucks macro. This is because the number of rows in the table is dynamic (it’s produced by iterating over the metadata keys) and there’s no nice way to tell the summary list component Nunjucks macro that we want a variable number of rows (we’d have to pull the construction of the `rows` object out of the template itself, which is complicated).

### DIFF
--- a/app/controllers/payment-links/get-edit-controller.js
+++ b/app/controllers/payment-links/get-edit-controller.js
@@ -17,7 +17,10 @@ module.exports = (req, res) => {
     self: formattedPathFor(paths.paymentLinks.edit, req.params.productExternalId),
     editInformation: formattedPathFor(paths.paymentLinks.editInformation, req.params.productExternalId),
     editReference: formattedPathFor(paths.paymentLinks.editReference, req.params.productExternalId),
-    editAmount: formattedPathFor(paths.paymentLinks.editAmount, req.params.productExternalId)
+    editAmount: formattedPathFor(paths.paymentLinks.editAmount, req.params.productExternalId),
+    addMetadata: formattedPathFor(paths.paymentLinks.metadata.add, req.params.productExternalId),
+    formattedPathFor,
+    paths
   }
 
   const gatewayAccountId = auth.getCurrentGatewayAccountId(req)
--- a/app/views/payment-links/edit.njk
+++ b/app/views/payment-links/edit.njk
@@ -1,7 +1,7 @@
 {% extends "../layout.njk" %}
 
 {% block pageTitle %}
-  Edit your payment link - {{currentService.name}} {{currentGatewayAccount.full_type}} - GOV.UK Pay
+  Payment link - {{currentService.name}} {{currentGatewayAccount.full_type}} - GOV.UK Pay
 {% endblock %}
 
 {% block side_navigation %}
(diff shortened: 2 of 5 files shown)


app/views/payment-links/edit.njk
```diff
@@ -120,6 +121,46 @@
...
   }}
 
+{% if currentService.experimentalFeaturesEnabled %}
+  <h2 class="govuk-heading-m">Reporting columns</h2>
+  <p class="govuk-body">Add extra columns to your transactions, for example cost centre code or business area.<p>
+  <p class="govuk-body">This can make it easier to match up payments with other systems.</p>
+
+  {% if product.metadata %}
+    <dl class="govuk-summary-list">
+      {% for metadataKey, metadataValue in product.metadata | dictsort %}
```
sfount: Nice 👍

app/controllers/payment-links/get-edit-controller.js
```diff
@@ -17,7 +17,9 @@ module.exports = (req, res) => {
     self: formattedPathFor(paths.paymentLinks.edit, req.params.productExternalId),
     editInformation: formattedPathFor(paths.paymentLinks.editInformation, req.params.productExternalId),
     editReference: formattedPathFor(paths.paymentLinks.editReference, req.params.productExternalId),
-    editAmount: formattedPathFor(paths.paymentLinks.editAmount, req.params.productExternalId)
+    editAmount: formattedPathFor(paths.paymentLinks.editAmount, req.params.productExternalId),
+    addMetadata: formattedPathFor(paths.paymentLinks.metadata.add, req.params.productExternalId),
+    editMetadata: formattedPathFor(paths.paymentLinks.metadata.edit, req.params.productExternalId)
```
sfount: Could `formattedPathFor` be used to also calculate the metatada key specific URL? that would avoid doing any mutation in the template. 

Two options for this might be: 
1. pass in `formattedPathFor` into the template -- this could then look like
```html
<a class="govuk-link" href="{{ formattedPathFor(paths.paymentLinks.metadata.edit, product.external_id, metadataKey) }}">
``` 

2. An easy to unit test method to add the edit path to the metadata itself before passing it into the template
```
product.metadata.forEach((metadata) => metadata.editUrl = formattedPathFor(...))
```
```
<a class="govuk-link" href="{{ metadata.editUrl }}">
``` 

The current option looks fine to me as well, happy to let the template do this just a little harder to test.

alexbishop1: The latter option is tricky because the metadata is just an object (e.g `{"key1": "value1", "key2": "value2"}`) so there’s nowhere to really put the edit URL. We could convert it to something else (an array of objects, say) but that would be a bit of a pain, so I went with the first option, even though it’s not as great.

test/unit/utils/replace_params_in_path_test.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
+
+// NPM dependencies
+const { expect } = require('chai')
+
+// Local dependencies
+const formattedPathFor = require('../../../app/utils/replace_params_in_path')
+
+describe('When a path is formatted', () => {
```
sfount: 👏 for covering this, can't believe it wasn't done before

PR #1883, opened 2020-02-26
PP-6160 Show associated metadata on Manage Payment links page

SandorArpa:
## WHAT
- For each Payment Link, display the associated metadata

with @iqbalgds

### DIFF
--- a/app/assets/sass/application.scss
+++ b/app/assets/sass/application.scss
@@ -39,3 +39,5 @@ $govuk-page-width: 1200px;
 @import "components/error-label";
 @import "components/digital-wallet";
 @import "components/padding";
+@import "components/reporting-columns";
+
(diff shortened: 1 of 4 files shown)


app/assets/sass/components/padding.scss
```diff
@@ -1,3 +1,21 @@
+// .reporting-columns{
```
sfount: Doesn't look like we need these classes

app/views/payment-links/manage.njk
```diff
@@ -16,6 +16,7 @@
   </aside>
 {% endif %}
   <h1 class="govuk-heading-l govuk-!-margin-top-6">{% if permissions.tokens_create %}Manage{% else %}View{% endif %} payment links</h1>
+  <h1 class="govuk-heading-l govuk-!-margin-top-6">Test</h1>
```
sfount: Oops I missed this first time, looks like it was temporary as well?

PR #1884, opened 2020-02-26
PP-6153 Payment link metadata add/edit page validation

sfount:
* validates for empty inputs
* parses error codes from products and specifically targets form elements 
* follows design system guidelines for forms, specific feedback provided on inputs as well as in an error summary https://design-system.service.gov.uk/components/error-summary/

### DIFF
--- a/app/controllers/payment-links/metadata/metadata-form.js
+++ b/app/controllers/payment-links/metadata/metadata-form.js
@@ -31,9 +31,7 @@ class MetadataForm {
     const errorMaps = {}
     for (const fieldKey in fields) {
       const field = fields[fieldKey]
-      console.log(field)
       field.validation.some((validationEntry) => {
-        console.log(validationEntry, this.values[field.id])
         const valid = validationEntry.validator(this.values[field.id])
         if (!valid) {
           errors.push({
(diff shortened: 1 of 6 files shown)


app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -0,0 +1,83 @@
...
+}
+
+const submissionErrorCodes = {
+  DUPLICATE_METADATA_KEYS: {
+    field: fields.metadataKey,
+    href: `#${fields.metadataKey.id}`,
+    text: 'Column header must be unique for this payment link'
+  }
+}
+
```
SandorArpa: I know it's unlikely but in theory we could have the 'NON_EXISTENT_METADATA_KEY' on edit submission

sfount: As discussed, thinking through scenarios that the server could reply with is great -- I think this one is covered by: 

* if the key doesn't exist when the user clicks edit the page to edit the key will `404`

This just leaves the scenarios that the key was deleted by another user while the user was editing it, in this case there's no guidance that we can give as we don't know what happened, hopefully this is edge case enough to not model it here.

PR #1887, opened 2020-02-27
PP-6146 Add missing gocardless env vars

danworth:
The following env vars are being provisioned by the selfservice secret
service but are not in the env-map or manifest.

`GOCARDLESS_TEST_OAUTH_BASE_URL`
`GOCARDLESS_LIVE_OAUTH_BASE_URL`
`GOCARDLESS_TEST_CLIENT_ID`
`GOCARDLESS_LIVE_CLIENT_ID`

## WHAT
Found these whilst cleaning up our staging config.


danworth:
@mrlumbu said this isn't actually needed by Selfservice so I'll remove the values form tfvars instead...

PR #1909, opened 2020-03-13
PP-6268 Implement Service Switcher

rorymalcolm:
- Implements the service switcher, allowing routes to provide a
serviceId through the serviceSwitcher middleware, which will then
redirect users to that service before loading the page
- Tests to ensure that functionality works as expected.

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -41,6 +41,7 @@ const transactionsDownloadController = require('./controllers/transactions/trans
 const transactionsListController = require('./controllers/transactions/transaction_list_controller')
 const transactionDetailController = require('./controllers/transactions/transaction_detail_controller')
 const transactionRefundController = require('./controllers/transactions/transaction_refund_controller')
+const transactionDetailRedirectController = require('./controllers/transactions/transaction_detail_redirect_controller')
 const credentialsController = require('./controllers/credentials_controller')
 const loginController = require('./controllers/login')
 const dashboardController = require('./controllers/dashboard')
@@ -223,6 +224,7 @@ module.exports.bind = function (app) {
   app.get(transactions.download, xraySegmentCls, permission('transactions-download:read'), getAccount, paymentMethodIsCard, transactionsDownloadController)
   app.get(transactions.detail, xraySegmentCls, permission('transactions-details:read'), getAccount, paymentMethodIsCard, transactionDetailController)
   app.post(transactions.refund, xraySegmentCls, permission('refunds:create'), getAccount, paymentMethodIsCard, transactionRefundController)
+  app.get(transactions.redirectDetail, xraySegmentCls, permission('transactions-details:read'), getAccount, paymentMethodIsCard, transactionDetailRedirectController)
 
   // YOUR PSP
   app.get(yourPsp.index, xraySegmentCls, permission('gateway-credentials:read'), getAccount, paymentMethodIsCard, yourPspController.getIndex)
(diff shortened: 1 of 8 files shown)


app/middleware/service-switcher.js
```diff
@@ -0,0 +1,20 @@
...
+const logger = require('../utils/logger')(__filename)
+const validAccountId = require('../utils/valid_account_id.js')
+const { renderErrorView } = require('../utils/response.js')
+
+const notAuthorised = (req, res) => {
+  return renderErrorView(req, res, 'You do not have the rights to access this service.', 403)
+}
+
+module.exports = (req, res, next) => {
+  let newAccountId = req.params.serviceId
```
stephencdaly: Does the path contain a service id or a gateway account id? I'm confused.
Also this should be a const.

app/routes.js
```diff
@@ -223,6 +224,7 @@ module.exports.bind = function (app) {
   app.get(transactions.download, xraySegmentCls, permission('transactions-download:read'), getAccount, paymentMethodIsCard, transactionsDownloadController)
   app.get(transactions.detail, xraySegmentCls, permission('transactions-details:read'), getAccount, paymentMethodIsCard, transactionDetailController)
   app.post(transactions.refund, xraySegmentCls, permission('refunds:create'), getAccount, paymentMethodIsCard, transactionRefundController)
+  app.get(transactions.serviceSwitchDetail, xraySegmentCls, permission('transactions-details:read'), switchService, getAccount, paymentMethodIsCard, transactionDetailController)
```
stephencdaly: I think the `permission` middleware should come after the `switchService` middleware, as we want to check the user has the permission for the service we're switching to.

test/integration/service_switcher_ft_tests.js
```diff
@@ -0,0 +1,171 @@
...
+      })
+  })
+
+  describe('The transaction view endpoint', function () {
+    it('should return the correct transaction view through the service switching middleware', function (done) {
+      let transactionId = 888
+      ledgerMock.get(ledgerTransactionPathFor(transactionId))
+        .reply(200,
+          {
+            'gateway_account_id': gatewayAccountId,
```
stephencdaly: can you use a fixture from `ledger_transaction_fixtures` to produce this response.

test/integration/service_switcher_ft_tests.js
```diff
@@ -0,0 +1,171 @@
...
+    let user = session.getUser({
+      gateway_account_ids: [gatewayAccountId],
+      permissions: [{ name: permissions }]
+    })
+    app = session.getAppWithLoggedInUser(getApp(), user)
+
+    userCreator.mockUserResponse(user.toJson(), done)
+    connectorMock.get(CONNECTOR_ACCOUNT_PATH)
+      .reply(200, {
+        'payment_provider': 'sandbox',
```
stephencdaly: can you use a fixture from `gateway_account_fixtures` to construct this response.

app/services/auth_service.js
```diff
@@ -145,12 +145,6 @@ function getCurrentGatewayAccountId (req) {
     logger.error(`Could not resolve the gatewayAccountId for user: ${lodash.get(req, 'user.externalId')}`)
     return null
   }
-  // check if we don't have Cookie value
-  // or if it's different user  / different userGatewayAccountIds
-  if ((!currentGatewayAccountId) ||
-    (currentServiceGatewayAccountIds.indexOf(currentGatewayAccountId) === -1)) {
-    currentGatewayAccountId = currentServiceGatewayAccountIds[0]
-  }
```
stephencdaly: do we not still need this behaviour in other circumstances?

app/middleware/service-switcher.js
```diff
@@ -0,0 +1,20 @@
...
+  if (validAccountId(newAccountId, req.user)) {
```
stephencdaly: Could this be handled by the `permission` middleware if we change the ordering?

app/routes.js
rorymalcolm: oops

PR #1918, opened 2020-03-18
PP-6303 All Services Transactions View

rorymalcolm:
- Adds all services transactions view, a url, hidden behind a feature
flag, that displays all the transactions across all services in a
downloadable filterable format for users to access.

### DIFF
--- a/app/controllers/all-service-transactions/get-controller.js
+++ b/app/controllers/all-service-transactions/get-controller.js
@@ -1,0 +1,58 @@
+'use strict'
+
+// NPM dependencies
+const _ = require('lodash')
+
+const { response, renderErrorView } = require('../../utils/response')
+const { ConnectorClient } = require('../../services/clients/connector_client.js')
+const serviceService = require('../../services/service_service')
+const transactionService = require('../../services/transaction_service')
+const { buildPaymentList } = require('../../utils/transaction_view.js')
+const { getFilters, describeFilters } = require('../../utils/filters.js')
+const router = require('../../routes.js')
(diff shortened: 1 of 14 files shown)


app/controllers/all-service-transactions/get-controller.js
```diff
@@ -0,0 +1,59 @@
...
+const { CORRELATION_HEADER } = require('../../utils/correlation_header.js')
+
+module.exports = async (req, res) => {
+  const correlationId = req.headers[CORRELATION_HEADER] || ''
+  const servicesRoles = _.get(req, 'user.serviceRoles', [])
+  const filters = getFilters(req)
+  try {
+    const aggregatedGatewayAccountIds = servicesRoles
+      .flatMap(servicesRole => servicesRole.service.gatewayAccountIds)
+      .reduce((accumulator, currentValue) => accumulator.concat(currentValue), [])
```
sfount: Similarly I think this is the only unique behaviour in this controller, we can probably abstract this out -- this would also mean that the current test coverage covers these new controllers (as there is only one)

app/controllers/transactions/transaction_detail_controller.js
```diff
@@ -18,6 +18,8 @@ module.exports = (req, res) => {
     .findWithEvents(accountId, chargeId)
     .then(data => {
       data.indexFilters = req.session.filters
+      console.log(`redirect: ${req.session.backLink}`)
```
sfount: debugging line?

PR #1944, opened 2020-04-09
PP-6333 Add Node Debug launch file and clean up code

iqbalgds:
- Add debug launch.json file
- Clean up start.js file

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,16 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Self Service Debug",
+            "protocol": "inspector",
+            "type": "node",
+            "request": "attach",
+            "port": 9401,
+            "address": "localhost",
+            "restart": true,
+            "localRoot": "${workspaceRoot}",
+            "remoteRoot": "/app"
+        },
+    ]
+}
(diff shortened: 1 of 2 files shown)


sfount:
This to clarify that this changes the entry file that self service uses in the production environment and so should be reviewed/ released very carefully. 

I have paired/ reviewed the changes and can't see any major downsides to removing the multiple thread handling per process; we already scale services based on containers so shouldn't need this additional layer of load balancing. Self services isn't required to serve many connections at once and likely never even uses the additional threads capacity.

To clarify this change has also been made as multiple threads has interfered with attaching a debugger to the Node process however isn't required for this work as a separate script could be used for development. 

The only consideration from me is if the log lines that have been removed are used in any monitoring tools however they aren't as far as I'm aware.

iqbalgds:
The Node multi-threading in Node was done using Throng:
https://github.com/alphagov/pay-frontend/pull/1483
This library has not been updated for 4 years and it not that popular in Github.
This was blocking us from debugging our node apps.
Therefore, this library and relevant code was removed from the `start.js`.
This should also have an added benefit of making the Node apps start up quicker.

iqbalgds:
Agreed - called the file `start-dev.js` instead of `dev_start.js`
Let me know if you agree

iqbalgds:
Not needed anymore - closing.

PR #1950, opened 2020-04-17
PP-6345 All service transactions support optional headers (stripe, moto)

sfount:
* for all service transactions an individual gateway account
configuration cannot be used to determine if optional headers should be
shown on the transaction list or transaction download
* instead reduce all gateway accounts available to the session so that
if any of them meet the criteria the headers will be shown

### DIFF
--- a/app/utils/valid_account_id.js
+++ b/app/utils/valid_account_id.js
@@ -10,15 +10,38 @@ const userServicesContainsGatewayAccount = function userServicesContainsGatewayA
 }
 
 const liveUserServicesGatewayAccounts = async function liveUserServicesGatewayAccounts (user) {
-  const emptyAccountsString = '[]'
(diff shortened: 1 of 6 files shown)


app/utils/valid_account_id.js
```diff
@@ -10,15 +10,38 @@ const userServicesContainsGatewayAccount = function userServicesContainsGatewayA
 }
 
 const liveUserServicesGatewayAccounts = async function liveUserServicesGatewayAccounts (user) {
```
stephencdaly, line 12: Could this file be renamed from `valid_account_id` while you're here, as it does a bit more than validate an account id. `service_gateway_accounts.js` or something, I dunno?

stephencdaly: Also there's no unit test for this function, reckon we should add one now that it does more?

PR #1955, opened 2020-04-20
PP-6285 Add cookie banner and analytics

kbottla:
## WHAT
- Added cookie banner, analytics and relevant tests
- Sets common cookie `gov_pay_cookie_policy` for analytics for consent choice
- Added new dev dependency `jsdom` for simulating DOM/window for testing cookie-banner/analytics functions


app/assets/sass/components/cookie-message.scss
```diff
@@ -1,19 +1,112 @@
-.cookie-banner {
-  box-sizing: border-box;
```
iqbalgds, line 2: Will the old banner work for now as well?

app/browsered/cookie-banner.js
```diff
@@ -0,0 +1,265 @@
+'use strict'
+
+const analytics = require('./analytics')
```
iqbalgds, line 3: Can we split the cookie functions out?
And require it in?

kbottla: Added new file for cookie functions

app/assets/sass/components/cookie-message.scss
kbottla: we don't use this class any more

PR #1957, opened 2020-04-22
PP-6285 Add IDs to buttons

kbottla:
## WHAT
- Added ID attributes to buttons to be used across tests. Using css class makes the tests break easily if more components with same class are added to pages

### DIFF
--- a/app/views/api-keys/_keys.njk
+++ b/app/views/api-keys/_keys.njk
@@ -15,7 +15,10 @@
       govukButton({
         text: "Create a new API key",
         href: routes.apiKeys.create,
-        classes: "generate-key"
+        classes: "generate-key",
+        attributes: {
+          id: "create-api-key"
+        }
       })
     }}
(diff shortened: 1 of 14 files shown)


app/views/api-keys/_keys.njk
```diff
@@ -15,7 +15,10 @@
       govukButton({
         text: "Create a new API key",
         href: routes.apiKeys.create,
-        classes: "generate-key"
+        classes: "generate-key",
+        attributes: {
+          id: "create-api-key"
```
sfount, line 20: adding `id`s to elements looks fine with me, in the future when thinking about automated tests we should consider using `data-action="create"` more -- this is a valid HTML5 meta tag that describes what will happen but doesn't impact the actual structure of the DOM. 

In this was the test can look something like
`buttonAction('create').click()`

which will scale no matter how many buttons we have, as long as they all do a unique type of action

PR #1978, opened 2020-05-06
PP 6466 Fix Google pay settings page

stephencdaly:
At some point, we stopped sending the merchant ID to connector. This means that although we were changing the `allow_google_pay` setting on the gateway account, it never appeared as enabled because when connector returns the setting, it checks that the merchant ID is there.

Fix the routes that the form on the apple/google pay settings page post to.

Also fix an issue where it was not possible to turn off google pay due to the `required: "true"` attribute on the merchant ID field. Instead do server side validation for this.

The tests need to be improved for this, but as it was already broken, this PR just aims to get it working again.

### Also

Stop alignment being messed up when a flash message shown. This was happening when errors were shown for the google pay settings page.

Added some padding to the settings navigation as a quick fix for this. Although we need to tinker with the layout to fix this properly.

### DIFF
--- a/app/controllers/digital-wallet/post-google-pay-controller.js
+++ b/app/controllers/digital-wallet/post-google-pay-controller.js
@@ -2,6 +2,7 @@
 
 // Local dependencies
 const paths = require('../../paths')
+const logger = require('../../utils/logger')(__filename)
 const { renderErrorView } = require('../../utils/response')
 const { ConnectorClient } = require('../../services/clients/connector_client')
 const { CORRELATION_HEADER } = require('../../utils/correlation_header')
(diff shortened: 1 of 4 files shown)


app/controllers/digital-wallet/post-google-pay-controller.js
```diff
@@ -11,14 +12,36 @@ module.exports = async (req, res) => {
...
   const correlationId = req.headers[CORRELATION_HEADER] || ''
   const enable = req.body['google-pay'] === 'on'
+  const gatewayMerchantId = req.body.merchantId
+
+  if (enable && !gatewayMerchantId) {
+    req.flash('genericError', '<h2>Please enter a valid Merchant ID</h2>')
+    return res.redirect(paths.digitalWallet.googlePay)
+  }
+
+  if (enable) {
```
sfount, line 22: I know this is following the previous code convention (which is a good thing!) but the logic here is quite hard to follow and might be simpler if it just did error handling once? 

Something like: 
```js
try {
  if (enablingGooglePay) {
    await connector.setGatewayMerchantId(gatewayAccountId, gatewayMerchantId, correlationId)
  }
  await connector.toggleGooglePay(gatewayAccountId, enable, correlationId)
  logger.info('...')
  req.flash('generic', '...')
} catch (error) {
  logger.error('...')
  req.flash('error', '...')
}
```

stephencdaly: The reason why I did it separately is because we rely on connector for the validation, so wanted to be sure that we're only showing a validation error when appropriate

PR #1990, opened 2020-05-18
BAU: local dev experience improvements

sfount:
Once a consensus is reached, the `README.md` will be updated to reflect sensible defaults and guidance for local development.

* add simple start file that won't try to manage threads or add overhead
* scripts to connect to different environments running locally
  * `tunnel`: any AWS environment that the developer has tunneled into
  * `local`: microservices running locally, usually managed by Docker
  * `docker`: this expects the service to running inside of Docker and is
  connecting to a network in Docker compose

Optimisations
* only include google analytics if we're in a production environment;
there's no need to include empty values locally


sfount:
These look like they've been incorporated in subsequent changes

PR #1992, opened 2020-05-18
PP-6490 Payout search client wrapper

sfount:
API resource wrapper utility for ledger payout search. Client call will
be verified when the provider has added this interaction.

Depends on https://github.com/alphagov/pay-selfservice/pull/1991


sfount:
This functionality will be included in another story

PR #2005, opened 2020-05-26
PP-6491 add ledger payout search pact test

SandorArpa:
## WHAT

- add a new search payouts contract test

### DIFF
--- a/test/fixtures/payout_fixtures.js
+++ b/test/fixtures/payout_fixtures.js
@@ -4,14 +4,14 @@ const pact = pactBase()
 
 function buildPayout (opts) {
   return {
-    gateway_payout_id: 'some-gateway-payout-id',
+    gateway_payout_id: opts.gatewayPayoutId || 'some-gateway-payout-id',
     gateway_account_id: opts.gatewayAccountId || 1,
-    amount: 10000,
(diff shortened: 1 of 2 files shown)


test/unit/clients/ledger_client/ledger_search_payout_test.js
```diff
@@ -0,0 +1,56 @@
+'use strict'
+
+// NPM dependencies
```
sfount, line 3: I don't think we need these comments in all of our files, this is just a semantic comment but I think it's plenty to separate the (npm) library imports by a new line from the local imports

test/fixtures/payout_fixtures.js
```diff
@@ -4,10 +4,10 @@ const pact = pactBase()
 
 function buildPayout (opts) {
   return {
-    gateway_payout_id: 'some-gateway-payout-id',
+    gateway_payout_id: opts.gateway_payout_id || 'some-gateway-payout-id',
```
sfount, line 7: It would be good to keep the options for this builder consistent, it looks like at the moment they use camelCase (`paidOutDate`). 

I think it's fine to use either but if we're changing to snake_case it should be the same for the other params

PR #2022, opened 2020-06-03
PP-6591 Payouts log user activity

sfount:
Add info logs that can be used to monitor feature usage: 
* payout page feature
* download transaction for a given payout feature

### DIFF
--- a/app/controllers/all-service-transactions/download-transactions.js
+++ b/app/controllers/all-service-transactions/download-transactions.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const logger = require('../../utils/logger')(__filename)
+const { keys } = require('@govuk-pay/pay-js-commons').logging
 const date = require('../../utils/dates')
 const transactionService = require('../../services/transaction_service')
 const Stream = require('../../services/clients/stream_client')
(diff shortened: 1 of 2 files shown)


app/controllers/all-service-transactions/download-transactions.js
```diff
@@ -29,6 +29,8 @@ module.exports = (req, res) => {
           time_taken: timestampStreamEnd - timestampStreamStart,
           from_date: filters.fromDate,
           to_date: filters.toDate,
+          gateway_payout_id: filters.gatewayPayoutId,
+          user_id: req.user && req.user.externalId,
```
stephencdaly, line 33: We should really be using the logging keys in pay-js-commons for structured logs. The idea is that we use these for keys we log in multiple places to ensure they all use the same key. For keys that only refer to the single log line, we can just define them here.

stephencdaly: https://github.com/alphagov/pay-js-commons/blob/master/src/logging/keys.js

sfount: Great suggestion 👍  didn't realise it had so many keys like https://github.com/alphagov/pay-js-commons/blob/master/src/logging/keys.js#L102. I'll update this

PR #2023, opened 2020-06-04
PP-6591 Add link to list payouts on "My services" page

stephencdaly:
Display the link if the user has any live stripe accounts.

Refactor the my-services/get-index-controller.js to use async await and be easier to follow.

### DIFF
--- a/app/controllers/my-services/get-index-controller.js
+++ b/app/controllers/my-services/get-index-controller.js
@@ -8,53 +8,61 @@ const { response } = require('../../utils/response')
 const serviceService = require('../../services/service_service')
 const getHeldPermissions = require('../../utils/get_held_permissions')
 
-module.exports = (req, res) => {
+const hasAccountWithPayouts = function hasLiveStripeAccount (gatewayAccounts) {
+  return gatewayAccounts.some(gatewayAccount =>
+    !isDirectDebitAccount(gatewayAccount) &&
+    gatewayAccount.payment_provider === 'stripe' &&
+    gatewayAccount.type === 'live')
+}
+
+const isDirectDebitAccount = function isDirectDebitAccount (gatewayAccount) {
+  return gatewayAccount.payment_method === 'direct debit'
+}
+
+module.exports = async (req, res) => {
   const servicesRoles = _.get(req, 'user.serviceRoles', [])
   const newServiceId = _.get(req, 'query.s')
-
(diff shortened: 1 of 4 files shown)


app/controllers/my-services/get-index-controller.js
```diff
@@ -8,53 +8,61 @@ const { response } = require('../../utils/response')
...
-      env: process.env
-    }
-    if (newServiceId) {
-      servicesData.filter(serviceData => {
-        return serviceData.external_id === newServiceId
-      }).forEach(service => {
-        data.new_service_name = service.name
-      })
-    }
+const isDirectDebitAccount = function isDirectDebitAccount (gatewayAccount) {
```
sfount, line 18: I believe this helper already exists (for some reason) in the direct debit connector client, there are other places in the code that we use that so I think it would be good to use it here. (Understood that this is just a good moving around of the legacy code though!)

app/controllers/my-services/get-index-controller.js
```diff
@@ -8,53 +8,61 @@ const { response } = require('../../utils/response')
 const serviceService = require('../../services/service_service')
 const getHeldPermissions = require('../../utils/get_held_permissions')
 
-module.exports = (req, res) => {
-  const servicesRoles = _.get(req, 'user.serviceRoles', [])
-  const newServiceId = _.get(req, 'query.s')
+const hasAccountWithPayouts = function hasLiveStripeAccount (gatewayAccounts) {
```
sfount, line 11: A small unit test to cover the happy path for this might be good, we have fixtures for services + gateway accounts so hopefully should be just writing one or two scenarios

test/cypress/integration/my-services/payouts_link_spec.js
```diff
@@ -0,0 +1,54 @@
...
+  beforeEach(() => {
+    // keep the same session for entire describe block
+    Cypress.Cookies.preserveOnce('session')
+    Cypress.Cookies.preserveOnce('gateway_account')
+  })
+
+  it('should display link to view payouts', () => {
+    cy.task('setupStubs', [
+      getUserStubWithServiceName(authenticatedUserId, [1]),
+      getGatewayAccountsStub(1, 'live', 'stripe')
```
sfount, line 17: Oh just realised the link logic is covered here

PR #2026, opened 2020-06-06
PP-6145 Bump stripe api version to 2020-03-02

sfount:
No major changes should have happened between our latest stripe api version and
the latest version. Bump for parity with out other services.


sfount:
We should keep up to date with features and fixes on the Stripe API, this should be done when the referenced ticket is prioritised.

PR #2032, opened 2020-06-08
BAU Payouts page add specific content for release date

sfount:
As the payouts tracking feature was released after services started
taking payments with Stripe, there are a number of services (who
potentially took payments before and are coming back to self service)
who would see no payouts listed.

This adds more specific content that will only be shown if the
environment specifies what the date the payouts processing began.

**Current copy** - this will be shown by default if there are no payouts and we have no release date in the environment.

Screenshot 2020-06-08 at 12 12 40

**Draft copy** - this will only be shown if there are no payouts AND we've specified a release date in the environment, this can be turned off if we think it's no longer relevant to communicate when to expect to see data from.

Screenshot 2020-06-08 at 12 14 35


christopherthomasdesign:
Looks good @sfount 🙂 only a couple of very minor stylistic things I'd recommend:

- removing 0 from the front of single digit numbers, makes them more natural to read and consistent with the content style guide
- putting the line about contacting support in a new paragraph – is distinct enough from the previous sentence to be worth calling out on its own IMO

sfount:
Thanks @christopherthomasdesign! Addressed both of your comments in https://github.com/alphagov/pay-selfservice/pull/2032/commits/1b73908a8464a7c312f1d3308d75eba76e7a6ace

PR #2033, opened 2020-06-08
 PP-6633 Add account status panel

kbottla:
## WHAT

- Added account status panel to Dashboard which is shown when feature flag `ENABLE_ACCOUNT_STATUS_PANEL` is `true`
      and Stripe account is not fully setup
- Add details button in the panel points to `routes.stripeSetup.accountSetup` which is to be added when Stripe onboarding
      design is finalised and implemented
- Message is tentative and to be updated after design is finalised
- Also removed Xray tracing for dashboard controller as we are using it (removed from frontend) and could potentially be removed from other places too.

Panel looks like below (content to be updated after the design is finalised)
Screenshot 2020-06-08 at 15 03 43

### DIFF
--- a/app/controllers/dashboard/dashboard-activity-controller.js
+++ b/app/controllers/dashboard/dashboard-activity-controller.js
@@ -9,6 +9,8 @@ const logger = require('../../utils/logger')(__filename)
 const response = require('../../utils/response').response
 const CORRELATION_HEADER = require('../../utils/correlation_header').CORRELATION_HEADER
 const LedgerClient = require('../../services/clients/ledger_client')
+const { ConnectorClient } = require('../../services/clients/connector_client')
+const connectorClient = new ConnectorClient(process.env.CONNECTOR_URL)
 const { isADirectDebitAccount } = require('../../services/clients/direct_debit_connector_client.js')
 const auth = require('../../services/auth_service.js')
 const { datetime } = require('@govuk-pay/pay-js-commons').nunjucksFilters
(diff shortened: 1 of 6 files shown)


app/controllers/dashboard/dashboard-activity-controller.js
```diff
@@ -120,36 +135,36 @@ module.exports = (req, res) => {
...
-          }))
+    const stripeAccountSetup = await getStripeAccountSetup(req.account, correlationId)
+    if (stripeAccountSetup) {
+      Object.assign(model, {
+        stripeAccountSetup: stripeAccountSetup
+      })
+    }
+
+    LedgerClient.transactionSummary(gatewayAccountId, fromDateTime, toDateTime, { correlationId: correlationId })
+      .then(result => {
```
stephencdaly, line 146: reckon we should use `await` here to be consistent within the function?

app/views/dashboard/_stripe_account_setup_banner.njk
```diff
@@ -0,0 +1,31 @@
+{% if stripeAccountSetup and (not stripeAccountSetup.bankAccount or not stripeAccountSetup.vatNumberCompanyNumber or not stripeAccountSetup.responsiblePerson) %}
```
stephencdaly, line 1: Do we need to ensure we only show to admin users in this PR or will that come in a further PR?

kbottla: Good spot! I am going raise a separate PR for this

app/controllers/dashboard/dashboard-activity-controller.js
kbottla: Agree! left it so there won't be much refactoring in this PR. Will raise a separate PR or include in next PR (if small)

PR #2049, opened 2020-06-16
PP-6638 Add Stripe account details flow controller

stephencdaly:
- Added '/stripe/add-psp-account-details' route and controller which
redirects to the next page of details to fill out depending on what
has already been filled out.
- If all details have been filled out, render the setup complete page.
The copy for this has not been added as part of this commit.

### DIFF
--- a/app/controllers/stripe-setup/add-psp-account-details/get-controller.js
+++ b/app/controllers/stripe-setup/add-psp-account-details/get-controller.js
@@ -1,0 +1,21 @@
+'use strict'
+
+const paths = require('../../../paths')
+const { response } = require('../../../utils/response')
+const { ConnectorClient } = require('../../../services/clients/connector_client')
+const connectorClient = new ConnectorClient(process.env.CONNECTOR_URL)
+
+module.exports = async (req, res) => {
+  const stripeAccountSetup = await connectorClient.getStripeAccountSetup(
+    req.account.gateway_account_id, req.correlationId)
+
+  if (!stripeAccountSetup.bankAccount) {
+    res.redirect(303, paths.stripeSetup.bankDetails)
+  } else if (!stripeAccountSetup.responsiblePerson) {
+    res.redirect(303, paths.stripeSetup.responsiblePerson)
+  } else if (!stripeAccountSetup.vatNumberCompanyNumber) {
+    res.redirect(303, paths.stripeSetup.vatNumberCompanyNumber)
+  } else {
+    response(req, res, 'stripe-setup/go-live-complete')
+  }
+}
--- a/app/controllers/stripe-setup/add-psp-account-details/get-controller.js
+++ b/app/controllers/stripe-setup/add-psp-account-details/get-controller.js
@@ -1,21 +1,27 @@
 'use strict'
 
 const paths = require('../../../paths')
-const { response } = require('../../../utils/response')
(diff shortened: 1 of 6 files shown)


app/controllers/stripe-setup/add-psp-account-details/get-controller.test.js
```diff
@@ -0,0 +1,86 @@
...
+
+const getController = function getController (stripeAccountSetupResponse) {
+  return proxyquire('./get-controller', {
+    '../../../services/clients/connector_client': {
+      ConnectorClient: function () {
+        this.getStripeAccountSetup = (gatewayAccountId, correlationId) => {
+          return new Promise(resolve => {
+            resolve(stripeAccountSetupResponse)
+          })
+        }
```
sfount, line 22: The powers that be in Javascript land anticipated the need to do this!

```suggestion
        this.getStripeAccountSetup = (gatewayAccountId, correlationId) => Promise.resolve(stripeAccountSetupResponse)
```

app/controllers/stripe-setup/add-psp-account-details/get-controller.test.js
```diff
@@ -0,0 +1,86 @@
...
+  })
+
+  it('should redirect to bank account setup page', async () => {
+    const controller = getController({
+      bankAccount: false,
+      responsiblePerson: false,
+      vatNumberCompanyNumber: false
+    })
+    await controller(req, res)
+    expect(res.redirect.calledWith(303, paths.stripeSetup.bankDetails)).to.equal(true)
```
sfount, line 54: I've found that `sinon.assert` is useful here as when it passes/ fails it will tell you the exact difference between what was called with and what was expected

```
sinon.assert.calledWith(spy, 303, paths.stripeSetup.bankDetails)
```

With the current method it will just tell you that it expected `calledWith` to be `true` but was `false`.

stephencdaly: good tip, have found the output was lacking

app/controllers/stripe-setup/add-psp-account-details/get-controller.js
```diff
@@ -0,0 +1,27 @@
...
+const paths = require('../../../paths')
+const logger = require('../../../utils/logger')(__filename)
+const { response, renderErrorView } = require('../../../utils/response')
+const { ConnectorClient } = require('../../../services/clients/connector_client')
+const connectorClient = new ConnectorClient(process.env.CONNECTOR_URL)
+
+module.exports = async (req, res) => {
+  try {
+    const stripeAccountSetup = await connectorClient.getStripeAccountSetup(
+      req.account.gateway_account_id, req.correlationId)
```
sfount, line 12: _Very minor alert_, if the method call goes over a line it's useful to have it consistent just when skimming through code and looking for the other bracket, something like:
```js
const stripeAccountSetup = await connectorClient.getStripeAccountSetup(
      req.account.gateway_account_id, req.correlationId
)
```

or (potentially preferable here): 
```js
const stripeAccountSetup = await connectorClient.getStripeAccountSetup(
    req.account.gateway_account_id,
    req.correlationId
)
```

stephencdaly: cool will change this in the next PR to save having to wait for another build

PR #2050, opened 2020-06-16
PP-6494 Trim and validate Worldpay 3DS Flex credentials

alexbishop1:
Trim leading and trailing whitespace from entered Worldpay 3DS Flex credentials and validate them according to the following rules:

- Organisational unit ID: 24 lower-case hexadecimal characters e.g. “1234567890abcdef12345678”
- Issuer: 24 lower-case hexadecimal characters e.g. “1234567890abcdef12345678”
- JWT MAC key: a UUID ID in its lowercase canonical textual representation e.g. “abcdef12-3456-7890-abCd-ef1234567890”.

Trimming is because users will almost certainly be copying and pasting these values, so may unintentionally copy leading ortrailing whitespace.

If validation fails, display a standard error summary with error messages saying “Enter your organisational unit ID in the format you received it” or similar, and represent the form with the trimmed organisational unit ID and issuer (but not the JWT MAC key because we treat that like a password) entered by the user prefilled.

Remove the existing client-side validation that checks that a value is entered for each field to avoid mixing client-side and server-side validation (the new server-side validation ensures that no fields are left blank).

Add hint text to each form field saying it’s “Provided to you by Worldpay” with an example of valid input (taken from Worldpay’s test values).

Add parenthical “API ID” and “API key” notes to the issuer and JWT MAC key labels to more closely match how Worldpay describe them. Don’t bother adding Worldpay’s “org unit ID” abbreviation to the organisational unit ID because it’s obvious.

Consistently use regular sentence case when displaying terms like “organisational unit ID” in the admin tool, even though Worldpay use title case, to match our style.

GOV UK Pay admin tool Worldpay 3DS Flex validation errors

### DIFF
--- a/app/controllers/your-psp/post-flex-controller.js
+++ b/app/controllers/your-psp/post-flex-controller.js
@@ -1,25 +1,54 @@
 'use strict'
 
+// NPM dependencies
+const lodash = require('lodash')
+
 // Local dependencies
 const paths = require('../../paths')
 const { renderErrorView } = require('../../utils/response')
 const { ConnectorClient } = require('../../services/clients/connector_client')
 const { correlationHeader } = require('../../utils/correlation_header')
+const worldpay3dsFlexValidations = require('./worldpay-3ds-flex-validations')
+
+// Constants
+const ORGANISATIONAL_UNIT_ID_FIELD = 'organisational-unit-id'
+const ISSUER_FIELD = 'issuer'
+const JWT_MAC_KEY_FIELD = 'jwt-mac-key'
 
 module.exports = async (req, res) => {
   const connector = new ConnectorClient(process.env.CONNECTOR_URL)
+
   const correlationId = req.headers[correlationHeader] || ''
+
   const accountId = req.account.gateway_account_id
+
   const removeCredentials = req.body['remove-credentials'] === 'true'
+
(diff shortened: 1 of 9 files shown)


app/controllers/your-psp/post-flex-controller.js
```diff
@@ -1,25 +1,54 @@
 'use strict'
```
iqbalgds, line 1: Minor point, wondering if it is worth having a test for this file?
You have good coverage already - so not sure if it is worth it.

iqbalgds:
Looks really good - 1 minor point.

app/controllers/your-psp/post-flex-controller.js
alexbishop1: I did consider this. There was no test for the controller before, just the Cypress tests for the whole thing. Obviously, it’s gotten more complex but the actual validation logic itself is well tested in unit tests (the validation rules) and the Cypress tests (the overall validation functionality). The main thing that isn’t really tested is the posting to connector and that hasn’t changed and would best be done as a Pact test. While I’m a little uncomfortable at the amount of testing for the 3DS Flex credentials setting overall, it is going to be used extremely rarely, so maybe it isn’t really worth going beyond what we already have.

iqbalgds: 👍

PR #2053, opened 2020-06-16
PP-6635 Redirect to live account dashboard route

sfount:
e9d3ecf96d694719efd42f6fc08edcf15168ca5f
> - add new controller that sets the live gateway account id to that of
  the  given service then redirects to the dashboard
> - if the user has no access to the given service the it redirects to
  the dashboard without setting the gateway account

### DIFF
--- a/app/controllers/stripe-setup/stripe-setup-link/stripe_setup_dashboard_redirect_controller.js
+++ b/app/controllers/stripe-setup/stripe-setup-link/stripe_setup_dashboard_redirect_controller.js
@@ -1,0 +1,31 @@
+'use strict'
+
+const paths = require('../../../../app/paths')
+const { ConnectorClient } = require('../../../services/clients/connector_client')
+const connectorClient = new ConnectorClient(process.env.CONNECTOR_URL)
+const logger = require('../../../utils/logger')(__filename)
+const { keys } = require('@govuk-pay/pay-js-commons').logging
+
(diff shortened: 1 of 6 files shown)


app/controllers/stripe-setup/stripe-setup-link/stripe_setup_dashboard_redirect_controller.js
```diff
@@ -0,0 +1,29 @@
...
+
+function getTargetServiceForRedirect (user, externalServiceId) {
+  return user.serviceRoles.filter((serviceRole) => serviceRole.service.externalId === externalServiceId)[0]
+}
+
+module.exports = async (req, res) => {
+  const externalServiceId = req.params.externalServiceId
+  const targetServiceRoleForRedirect = getTargetServiceForRedirect(req.user, externalServiceId)
+  if (targetServiceRoleForRedirect) {
+    const result = await connectorClient.getAccounts(targetServiceRoleForRedirect.service.gatewayAccountIds)
```
sfount, line 17: I think this function takes a `param` object so I don't think this would work, I think this means the happy path test will need to be adjusted. 

From what I can see the usage is `{ gatewayAccountIds: [ ] }`

sfount:
Tested this locally with @SandorArpa, following scenarios seem good: 

* Redirecting to the only live account for a given service ID
* Not redirecting if there multiple live accounts
* Not redirecting if there are no live accounts

PR #2056, opened 2020-06-17
PP-6652 Account status tag "setup" status

sfount:
Add account setup details to the session account object IFF this is a
live Stripe account.

* use `govuk-tag` to be inline with the design system
* grey for test
* light blue for information required
* blue for live

Screenshot 2020-06-17 at 15 29 33

Screenshot 2020-06-17 at 15 29 44

Screenshot 2020-06-17 at 15 29 58

### DIFF
--- a/app/middleware/get_gateway_account.js
+++ b/app/middleware/get_gateway_account.js
@@ -1,19 +1,11 @@
 'use strict'
 
-// npm dependencies
 const _ = require('lodash')
-const AWSXRay = require('aws-xray-sdk')
-const getNamespace = require('continuation-local-storage').getNamespace
-
-// local dependencies
 const logger = require('../utils/logger')(__filename)
 const auth = require('../services/auth_service.js')
 const Connector = require('../services/clients/connector_client.js').ConnectorClient
 const connectorClient = new Connector(process.env.CONNECTOR_URL)
 const directDebitConnectorClient = require('../services/clients/direct_debit_connector_client.js')
-
-// Constants
-const clsXrayConfig = require('../../config/xray-cls')
 
 module.exports = function (req, res, next) {
   const accountId = auth.getCurrentGatewayAccountId(req)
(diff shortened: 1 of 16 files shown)


app/views/includes/phase_banner.njk
```diff
@@ -4,7 +4,26 @@
...
+            govuk-tag--blue
+          {% else %}
+
+          {% endif %}
+        {% endset %}
+        {% set tagText %}
+          {% if currentGatewayAccount.full_type === "Sandbox test" %}
+            Test
+          {% elseif currentGatewayAccount.payment_provider === 'stripe' and not stripeAccountIsSetup %}
+            {{ currentGatewayAccount.full_type }} - information needed
```
stephencdaly, line 21: what are the different values of `full_type`?

app/views/includes/phase_banner.njk
```diff
@@ -4,7 +4,26 @@
...
       {% if currentGatewayAccount %}
-        <strong class="environment-tag environment-tag-{{currentGatewayAccount.type}}">{% if currentGatewayAccount.full_type === 'Sandbox test' %}Test{% else %}{{currentGatewayAccount.full_type}}{% endif %}</strong>
+        {% set stripeAccountIsSetup = (currentGatewayAccount.accountSetupDetails.bankAccount and currentGatewayAccount.accountSetupDetails.responsiblePerson and currentGatewayAccount.accountSetupDetails.vatNumberCompanyNumber) %}
+        {% set tagClass %}
+          {% if currentGatewayAccount.type === "test" %}
+            govuk-tag--grey
+          {% elseif currentGatewayAccount.payment_provider === 'stripe' and not stripeAccountIsSetup %}
+            govuk-tag--blue
+          {% else %}
+
```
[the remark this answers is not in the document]
sfount: Good question! I don't think there's a change here yet?

The GOV.UK design system `tag` component uses the ` $govuk-brand-colour` if no modifier classes are added. That means omitting a modifier class will use the 'live' proposed colour as per the design spec.

Happy to try and clarify that through naming or something.

app/views/includes/phase_banner.njk
sfount: It's very curious, this just moves the logic that's already there.

I think it all needs a rethink when we're looking at cleaning up/ potentially restructuring the admin tool: 
https://github.com/alphagov/pay-selfservice/blob/a3e26b0fc1a08f628c8aa6ae814e5ffdcc91beaa/app/utils/display_converter.js#L78-L80

app/views/includes/phase_banner.njk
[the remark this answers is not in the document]
iqbalgds: I see - that all makes sense.
Can we remove the empty `else` tag?
Not sure if Nunjucks allows that.

app/views/includes/phase_banner.njk
stephencdaly: Dh pretty curious. I guess it's going to show "Worldpay test" in some cases?
Do we want the tag to be always in full caps now?

Could we maybe move all this display formatting into the template to make it clearer what's going on?

sfount: Yeah I think that's the idea 'Test' is used for sandbox accounts but if it's a test account with an integration it's more specific. 

I think all caps looks much better and is in-line with the design system but would be happy to discuss it.

Yeah totally agreed with moving the formatting into the template, we'll just need to dig into how many different downstream things use the `full_type` attribute.

app/middleware/get_gateway_account.js
```diff
@@ -33,23 +25,24 @@ module.exports = function (req, res, next) {
...
-        next()
-      })
-      .catch(err => {
-        subsegment.close(err)
-        logger.error(`${req.correlationId} - Error when attempting to retrieve card gateway account: ${err}`)
-        next()
+  return connectorClient.getAccount(params)
+    .then(data => {
+      req.account = _.extend({}, data, {
+        supports3ds: ['worldpay', 'stripe', 'epdq', 'smartpay'].includes(_.get(data, 'payment_provider')),
```
iqbalgds, line 31: Nice 👍

app/views/dashboard/_stripe_account_setup_banner.njk
```diff
@@ -1,4 +1,5 @@
 {% if permissions.stripe_account_details_update %}
```
stephencdaly, line 1: I think we need an `{% if showFeatureStripeOnboarding %}` now. Before we were checking that in the controller.

sfount:
@stephencdaly thanks for spotting that: 
1. make the test unhappy (looks like it wasn't configured to throw issues)
2. add logic to feature flag this
3. make sure the test passes

https://github.com/alphagov/pay-selfservice/pull/2056/commits/5cdadf0040399efa6fc6251fc1b1f875fe320186

PR #2075, opened 2020-06-24
PP-6644 Set default from date for unfiltered transaction search

stephencdaly:
- When a user visits the `/transactions` page, the search that is run automatically only lists transactions from 1 month ago. This will use the same day in the previous month, or if that date does not exist then the last day of the previous month.
- When a user clears the "From date" filter and clicks "Filter", the search is performed with no from date limit.
- The "Clear Filter" button has been renamed to "Reset filters" and will revert to searching for unfiltered transactions limited to 1 month ago.

### DIFF
--- a/app/utils/filters.js
+++ b/app/utils/filters.js
@@ -5,6 +5,7 @@ const check = require('check-types')
 const Paginator = require('../utils/paginator.js')
 const states = require('../utils/states')
 const _ = require('lodash')
+const moment = require('moment-timezone')
 
 function validateFilters (filters) {
   let pageSizeIsNull = !check.assigned(filters.pageSize)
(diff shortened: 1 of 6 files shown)


app/utils/filters.js
```diff
@@ -17,6 +18,14 @@ function validateFilters (filters) {
 
 function getFilters (req) {
   let filters = qs.parse(req.query)
+
+  // If a search is being performed without any filters specified, detault the from date filter to
+  // 30 days ago. We do this because the count query that is currently required by the search is
+  // very inefficient. If the search has been filtered, still allow the from date to be blank.
+  if (_.isEmpty(filters)) {
+    filters.fromDate = moment().tz('Europe/London').subtract(30, 'days').format('DD/MM/YYYY')
+  }
```
iqbalgds, line 27: Will this show up in the Front end?  I.e. will the user the `from date` being set?

stephencdaly: Yep
Screenshot 2020-06-24 at 13 12 10

iqbalgds:
Good work - I had a look over it and looks good.  Some Cypress bits I am not sure on so will be good for someone else to take a look as well

test/cypress/integration/transactions/transaction_search_spec.js
```diff
@@ -133,395 +82,237 @@ describe('Transactions', () => {
...
-          cy.get('#filter').click()
-
-          // Ensure the right number of transactions is displayed
-          cy.get('#transactions-list tbody').find('tr').should('have.length', filteredByToDateTransactions.length)
-
-          // Ensure the expected transactions are shown
-          cy.get('#transactions-list tbody').find('tr').first().find('th').should('contain', filteredByToDateTransactions[0].reference)
-        })
-
-        it('should have the right number of transactions when filtering by multiple card brands and a partial email', () => {
```
kbottla, line 315: Good cleanup 👏  to a single test for filter by fields

PR #2089, opened 2020-06-30
PP-6709 Only show all service transactions link for live service users

sfount:
* remove feature flag that has been enabled in all environments.
* only show all service transactions link if a user has access to live
services that they would have permissions to see.


sfount:
👍 It uses already existing/ presumably tested methods to do the filtering logic so I figured it was mostly covered.

Passing the value through to the template correctly isn't covered by that.

sfount:
@stephencdaly thanks for the comments; added that coverage in https://github.com/alphagov/pay-selfservice/pull/2089/commits/6b12baf74be46497533a07da7c5fa04a9f310596

PR #2091, opened 2020-07-01
BAU - Firebreak vscode debug

iqbalgds:
* Firebreak - Add VSCode debug mode
    * Add a VSCode launch debug  config so the app can run in debug mode directly from VScode.
    * Have the following configs:
      * Regular debug the app while it is running
      * Debug current unit test
      * Debug all unit test files
      * Update README

- BAU - scripts to configure local developemnt env
    * add simple start file that won't try to manage threads or add overhead
    * scripts to connect to different environments running locally
      * tunnel: any AWS environment that the developer has tunneled into
      * local: microservices running locally, usually managed by Docker
      * docker: this expects the service to running inside of Docker and is
      connecting to a network in Docker compose
    
- Optimisations
    * only include google analytics if we're in a production environment;
    there's no need to include empty values locally


app/middleware/permission.js
```diff
@@ -6,8 +6,9 @@ const resolveService = require('./resolve_service')
  * For the moment if undefined, the check is skipped.
  */
 module.exports = function (permission) {
+  const isDevelopment = process.env.NODE_ENV === 'development'
   return [resolveService, function (req, res, next) {
-    if (!permission) {
+    if (!permission || isDevelopment) {
```
iqbalgds, line 11: Remove this

app/views/includes/cookie-message.njk
```diff
@@ -1,3 +1,4 @@
+{% if not isDevelopment %}
```
iqbalgds, line 1: Revert this change

app/middleware/permission.js
iqbalgds: Done

app/views/includes/cookie-message.njk
iqbalgds: Done

README.md
```diff
@@ -23,6 +23,21 @@ You can watch Nodemon do it’s thing if you run `docker logs -f selfservice`.
 
 If you’re making changes to client-side JS or Sass files (anything within [`/browsered/`](https://github.com/alphagov/pay-selfservice/tree/BAU-update-README-to-explain-livereload/app/browsered) or [`/assets/`](https://github.com/alphagov/pay-selfservice/tree/BAU-update-README-to-explain-livereload/app/assets)) then running `npm run watch-live-reload` will watch for changes and recompile. Nodemon does not do anything here as that’s not necessary. If you install the [livereload browser plugin](http://livereload.com/extensions/) then it will refresh your page once the assets have been compiled to the `/public` folder.
 
+### Steps to run locally - new method
```
sfount, line 26: This can be done as a separate PR but I'd recommend removing the "old method" and including all the necessary instructions inline here

README.md
```diff
@@ -23,6 +23,21 @@ You can watch Nodemon do it’s thing if you run `docker logs -f selfservice`.
...
+The new method is faster and easier to run in debug mode.
+* `./scripts/generate-dev-environment.sh local` - this will create a local `.env` file.
+* Repeat the same steps above.
+* Run the app with: `npm run start:dev`
+* Go to: http://localhost:3000 -> Note the diff port number.
+
+#### Run in debug mode
```
sfount, line 33: Now that we've added configurations to launch individual mocha tests in debug mode it would be great to document that behaviour. 

I would think about organising all of these under a 'VSCode' heading as they're specific to that IDE

PR #2104, opened 2020-07-09
PP-6688 Remove `check your answers` pages

kbottla:
## WHAT
- Removed `Check your answers` pages for all onboarding tasks. Data is submitted to Stripe as soon as form is submitted.
- Renamed button 'Continue' to 'Save and continue` to reflect the action
- Splits VAT number and Company number to individual tasks and data is submitted to Stripe individually.
     - Defaults `Company number` to `NOTAPPLI` as Stripe needs some data (which is being addressed by Stripe) even when company number is not available.
- Removed loads of cypress tests for combined `vat_number_company_number` task as they are no longer needed.


app/controllers/stripe-setup/vat-number-company-number/company-number/get.controller.js
```diff
@@ -1,27 +1,8 @@
...
-  } else {
-    pageData = {
-      errors: {},
-      companyNumberDeclaration: '',
-      companyNumber: ''
-    }
-  }
-
-  return response(req, res, 'stripe-setup/vat-number-company-number/company-number', pageData)
+  return response(req, res, 'stripe-setup/vat-number-company-number/company-number')
```
iqbalgds, line 7: Minor point - Seeing as we are splitting VAT and Company number - is it worth changing the URL to: 
`stripe-setup/company-number`?

iqbalgds:
Amazing work - left 1 minor point.  Everything else looks good to me.
As its such a big PR - will let someone else check it as well.

app/controllers/stripe-setup/vat-number-company-number/company-number/get.controller.js
kbottla: yes, planning to change it separately as updating here could make the PR big

app/models/StripeAccountSetup.class.js
```diff
@@ -5,6 +5,8 @@ class StripeAccountSetup {
     this.bankAccount = opts.bank_account
     this.vatNumberCompanyNumber = opts.vat_number_company_number
```
sfount, line 6: I think uses of this have been removed so it can probably be removed from the model

kbottla: 👍 this is going into separate PR to remove here and all references

PR #2107, opened 2020-07-10
PP-6688 Change vat_number & company number tasks path

kbottla:
## WHAT
- Moves vat-number & company-number paths out of /vat-number-company-number/*** path as these are now independent tasks


iqbalgds:
Nice work 👍

PR #2124, opened 2020-07-20
PP-6771 Use structured arguments when logging outgoing requests

stephencdaly:
Use structured logging arguments for all log lines produced when making outgoing requests and consistently include the available information for all log lines.

### DIFF
--- a/app/utils/request_logger.js
+++ b/app/utils/request_logger.js
@@ -1,34 +1,53 @@
 const logger = require('./logger')(__filename)
+const { keys } = require('@govuk-pay/pay-js-commons').logging
 
 module.exports = {
   logRequestStart: context => {
-    logger.debug(`Calling ${context.service}  ${context.description}-`, {
+    const logContext = {
       service: context.service,
       method: context.method,
-      url: context.url
-    })
+      url: context.url,
+      description: context.description
+    }
+    logContext[keys.CORRELATION_ID] = context.correlationId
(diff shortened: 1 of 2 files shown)


app/utils/request_logger.js
```diff
@@ -1,34 +1,53 @@
...
+  logRequestEnd: (context, response) => {
     let duration = new Date() - context.startTime
-    logger.info(`[${context.correlationId}] - ${context.method} to ${context.url} ended - elapsed time: ${duration} ms`)
+    const logContext = {
+      service: context.service,
+      method: context.method,
+      url: context.url,
+      description: context.description,
+      response_time: duration,
+      status: response && response.statusCode
```
sfount, line 24: It's a bit confusing that until you added `status` here this would have been guaranteed logged exactly the same thing for both failures and successes (so a positive change!). 

We call `logRequestEnd` no matter if it's a success, "failure" or "error". I wonder if would be easier to follow if `logRequestFailure` and `logRequestEnd` just did their own `duration` calculation and each state only had one self contained log line. 

In that case `logRequestEnd` would be more specifically `logRequestSucceeded`.

stephencdaly: I see what you mean, but also i think it's more useful to have a single line that we can see the status code from.

Maybe that would mean just getting rid of `logRequestFailure`?

PR #2137, opened 2020-07-28
PP-6820 Improve hint for org name

kbottla:
## WHAT
- Improved hint text for org name to make it clear for service users entering org name.
- Uses html instead of nunjucks template as html line breaks are escaped if added to nunjucks text

## New pages

# **Onboarding flow**
Screenshot 2020-07-28 at 09 41 40

# **Organisation details**
Screenshot 2020-07-28 at 12 08 27

### DIFF
--- a/app/views/merchant_details/edit_merchant_details.njk
+++ b/app/views/merchant_details/edit_merchant_details.njk
@@ -50,45 +50,35 @@
   {% endif %}
   <h1 class="govuk-heading-l page-title">Organisation details</h1>
   {% if has_card_and_dd_gateway_account %}
-    <p class="govuk-body" id="merchant-details-info">Payment card schemes and Direct Debit require the details of the organisation taking payment to be shown to the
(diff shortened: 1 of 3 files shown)


app/views/merchant_details/edit_merchant_details.njk
```diff
@@ -56,39 +56,29 @@
     <p class="govuk-body" id="merchant-details-info">Direct Debit requires the details of the organisation taking payment to be shown to the payer. This information is
       displayed in emails and on payment pages.</p>
   {% elif has_card_gateway_account %}
-    <p class="govuk-body" id="merchant-details-info">Payment card schemes require the details of the organisation taking payment to be shown on payment pages.</p>
+    <p class="govuk-hint" id="merchant-details-info">Payment card schemes require the details of the organisation taking payment to be shown on payment pages.</p>
   {% endif %}
-
+  <br>
```
stephencdaly, line 61: We generally use spacing classes to control the amount of vertical space, rather than line breaks.

https://design-system.service.gov.uk/styles/spacing/

kbottla: 👍  Updated

PR #2142, opened 2020-08-01
PP-6716 Add integrity checks for CDN scripts

sfount:
Add an integrity check to ensurethe file we receive from the CDN is the 
one expected.

Future work may see these files removed entirely (in favour of HTML
standards) or bundled and served by the application.

## To test this change
- [x] Verify that the SHA-384 signatures for the file line up with what
is being served by the CDN
- [ ] Ensure the transaction search date selection still works on common
browsers and no console errors or warnings are shown

### DIFF
--- a/app/views/all-service-transactions/index.njk
+++ b/app/views/all-service-transactions/index.njk
@@ -13,10 +13,10 @@ Transactions for all live services - GOV.UK Pay
 
 {% block pageSpecificJS %}
 <script src="/public/js/components/link-follower.js"></script>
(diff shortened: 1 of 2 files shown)


alexbishop1:
Any particular reason why it’s SHA-384?

sfount:
@alexbishop1 that seems to be the trend in sub resource integrity land. 

Currently allowed hashes are `sha256`, `sha384`, and `sha512`. 

https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

PR #2158, opened 2020-08-06
PP-6848 Remove additional call for stripe setup

kbottla:
## WHAT
- We have getAccount (get-gateway-account.js) middleware where we get make a call to connector to get current stripe setup and is added to req.account object.
  We can remove getStripeSetup calls from controllers as it is taken care in middle and information available on req object

### DIFF
--- a/app/controllers/stripe-setup/add-psp-account-details/get-controller.js
+++ b/app/controllers/stripe-setup/add-psp-account-details/get-controller.js
@@ -1,29 +1,24 @@
 'use strict'
 
 const paths = require('../../../paths')
-const logger = require('../../../utils/logger')(__filename)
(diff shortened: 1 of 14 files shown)


test/cypress/integration/transactions/transaction-details.cy.test.js
```diff
@@ -507,7 +507,7 @@ describe('Transaction details page', () => {
     })
   })
 
-  describe('refunds', () => {
+  describe.only('refunds', () => {
     it('should fail when an invalid refund amount is specified', () => {
```
iqbalgds, line 511: A 'only' left in.

kbottla: 🤦  Good spot! updated

PR #2174, opened 2020-08-12
PP-6828 Refactor cypress stubs

kbottla:
## WHAT

- Refactored stubs setup for multiple stubs ( includes stripe account setup, stripe account, and some others )


test/cypress/utils/stripe-account-setup-stub.js
```diff
@@ -0,0 +1,69 @@
...
+    stubOptions.company_number = opts.companyNumber
+  }
+
+  return {
+    name: 'getGatewayAccountStripeSetupSuccess',
+    opts: stubOptions
+  }
+}
+
+const getGatewayAccountStripeSetupFlagForMultipleCalls = function (opts) {
```
stephencdaly, line 25: I think this stub is a bit overcomplicated, but I think we have a recommendation to replace the test that requires it, so should be able to get rid of it in not too long hopefully.

PR #2176, opened 2020-08-12
PP-6864 merge responsible person validation tests

SandorArpa:
## WHAT
- merge appropriate validation test into one: long text for address line 2 and invalid birth date

### DIFF
--- a/test/cypress/integration/stripe-setup/responsible-person.cy.test.js
+++ b/test/cypress/integration/stripe-setup/responsible-person.cy.test.js
@@ -15,17 +15,13 @@ describe('Stripe setup: responsible person page', () => {
   const typedLastName = ' Benn'
   const addressLine1 = '52 Festive Road'
   const typedAddressLine1 = ' 52 Festive Road'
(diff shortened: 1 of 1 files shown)


test/cypress/integration/stripe-setup/responsible-person.cy.test.js
```diff
@@ -117,14 +114,22 @@ describe('Stripe setup: responsible person page', () => {
           cy.get('input#home-address-postcode[name=home-address-postcode][autocomplete=postal-code]').should('have.attr', 'value', 'not a valid UK postcode')
         })
 
+        cy.get('.govuk-form-group--error > input#home-address-line-2').parent().should('exist').within(() => {
+          cy.get('.govuk-error-message').should('exist')
+          cy.get('input#home-address-line-2').should('have.attr', 'value', longText)
```
stephencdaly, line 119: Can we remove typing in the long text and checking for the validation  for address line 2 completely? It takes several seconds to type in and is very much an edge case so I don't think it's worth testing for the time taken.

PR #2180, opened 2020-08-13
PP-6828 Refactor service stubs

kbottla:
## WHAT
- Refactor service stubs and user stubs further

### DIFF
--- a/test/cypress/utils/gateway-account-stubs.js
+++ b/test/cypress/utils/gateway-account-stubs.js
@@ -48,8 +48,71 @@ const getDirectDebitGatewayAccountSuccess = function (opts) {
   }
 }
 
+const postCreateGatewayAccountSuccess = function (opts) {
+  return {
+    name: 'postCreateGatewayAccountSuccess',
+    opts: {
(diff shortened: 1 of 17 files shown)


test/cypress/utils/gateway-account-stubs.js
```diff
@@ -48,8 +48,71 @@ const getDirectDebitGatewayAccountSuccess = function (opts) {
...
+      repeat: 2
+    },
+    {
+      gateway_account_id: opts[1].gatewayAccountId,
+      payment_provider: opts[1].paymentProvider,
+      worldpay_3ds_flex: opts[1].worldpay3dsFlex,
+      credentials: opts[1].credentials
+    }]
+  }
+}
```
stephencdaly, line 98: This is a little bit of a weird stub, we could probably amend the Cypress test in some way to avoid having to do it like this. I might look into that quickly as part of the other story we've got in progress. Then maybe we can remove this.

kbottla: yeah, there are few which could be improved

PR #2185, opened 2020-08-13
PP-6834 Add government banking option to go live

sfount:
Allow users to select the government banking Worldpay contract using the
constant added in https://github.com/alphagov/pay-adminusers/pull/836.

At the time of writing, the flow for this option should follow the same
behaviour as the bring your own contract Worldpay selection, the
differentiation will be for the Support roles benefit.

Screenshot 2020-08-13 at 16 01 27

Screenshot 2020-08-13 at 16 01 32


sfount:
Content and flow to be run by @miriamalice before releasing, merging to verify behaviour in test environment.

PR #2192, opened 2020-08-14
PP-6864 Replace test for responsible person template logic

stephencdaly:
We had a cypress test that tested an edge case of template logic that both address line 1 and address line 2 display the same error summary in the summary list, and if there are errors for both fields, it is only displayed once.

To do this test, we had to type a very long string for address line 2 as this is the only error that can be simulated.

As this is template logic, it is suited to a template test. So replace the cypress test with some simple template tests for this.


stephencdaly:
Yeh it's kinda tested by cypress, rather than unit. There is still coverage for this in the cypress tests, with a small blindspot for the address line 2 error apart from what is covered here, but an error in address line 2 is a real edge case so I feel this template test is good enough to ensure we don't have a regression in the template.

There are unit tests for actual validation logic, rather than how the controller formats this list.

PR #2207, opened 2020-08-19
PP-6951 Improvements to refund controller

stephencdaly:
Based on comments that were made in the mobbing session.


app/services/transaction.service.js
```diff
@@ -73,28 +73,28 @@ const ledgerFindWithEvents = async function ledgerFindWithEvents (accountId, cha
...
+    'userExternalId': userExternalId
+  })
 
-    const params = {
-      gatewayAccountId: accountId,
-      chargeId: chargeId,
-      payload: payload,
-      correlationId: correlationId
-    }
+  const params = {
```
sfount, line 90: If we use the more specific `gatewayAccountId` at the parameter level the boilerplate could be reduced here
```suggestion
  const params = { gatewayAccountId, chargeId, payload, correlationId }
```

app/services/transaction.service.js
```diff
@@ -73,28 +73,28 @@ const ledgerFindWithEvents = async function ledgerFindWithEvents (accountId, cha
...
-      'userExternalId': userExternalId
-    })
+  const payload = {
+    amount: amount,
+    refund_amount_available: refundAmountAvailable,
+    user_external_id: userExternalId,
+    user_email: userEmail
+  }
+
+  logger.log('info', 'Submitting a refund for a charge', {
```
sfount, line 83: We've discussed how ideally we should be using standard logging keys here but we think there's a slightly bigger piece of work about making that easier in self service (potentially using a similar methodology to frontend or self service). 

That won't be part of this PR to keep it focussed.

sfount (approved):
Looking good!

app/services/transaction.service.js
```diff
@@ -72,29 +72,27 @@ const ledgerFindWithEvents = async function ledgerFindWithEvents (accountId, cha
...
+  }
+
+  const logContext = {
+    refund_amount_available: refundAmountAvailable,
+    amount: amount
+  }
+  logContext[keys.USER_EXTERNAL_ID] = userExternalId
+  logContext[keys.GATEWAY_ACCOUNT_ID] = gatewayAccountId
+  logContext[keys.PAYMENT_EXTERNAL_ID] = chargeId
+  logContext[keys.CORRELATION_ID] = correlationId
```
sfount, line 90: 🍬

sfount (approved):

PR #2213, opened 2020-08-20
PP-6951 Handle and show refund errors better

stephencdaly:
Make it clearer which errors can come from connector, and how this
relates to what we display to the user.

Use the `error_identifier` from the connector response rather than the
status code to determine what went wrong. To allow for this, switch to
using `base.client` rather than the old client to make the request to
connector.

Remove HTML from the refund success and error messages, instead pass
a plain string for flash to display.

Remove `transaction-details-refunds.ft.test.js` as these tests didn't
assert on anything meaningful and the behaviour they should cover is
covered by the Cypress tests for the transaction details page and the
controller integration test.

Fix `transaction-refund.controller.it.test.js` which didn't actually
assert on anything for the error cases, and would have failed if they
did 😬.


app/services/transaction.service.js
```diff
@@ -90,28 +90,36 @@ const refund = function refundTransaction (gatewayAccountId, chargeId, amount, r
...
-    connector.postChargeRefund(params, function () {
-      resolve()
-    }).on('connectorError', (err, response, body) => {
-      err = 'REFUND_FAILED'
-      if (response && response.statusCode === 400) {
-        if (body.reason) {
-          err = body.reason
-        }
+  try {
+    await connector.postChargeRefund(gatewayAccountId, chargeId, payload, correlationId)
```
sfount, line 101: 🎉  so much clearer, replacing use of the `oldBaseClient` makes upstream code 10x easier to read.

test/integration/transaction-details-refunds.ft.test.js
```diff
@@ -1,150 +0,0 @@
...
-      .set('Accept', 'application/json')
-      .send(viewFormData)
-      .expect(302)
-      .end((err, res) => {
-        expect(err).to.equal(null)
-        expect(res.header.location).to.equal('/transactions/12345')
-        done()
-      })
-  })
-})
```
sfount, line 150: 🤦 👍

sfount (approved):

PR #2215, opened 2020-08-20
PP-6951 Remove markup from success flash messages

stephencdaly:
We shouldn't be constructing HTML in controllers, templates should deal
with this.

There are a number of success flash messages where we use <h2>. All
success messages should be uniformly styled, so just remove the option
of having headers. If we need specific styling, individual templates
should handle custom flash message, but in all these cases it doesn't
seem necessary.


app/controllers/feedback/post-index.controller.js
```diff
@@ -28,7 +28,7 @@ ${req.body['feedback-suggestion']}`
   zendeskClient.createTicket(opts)
     .then(() => {
       lodash.set(req, 'session.pageData.feedback', {})
-      req.flash('generic', `<h2>Thanks for your feedback</h2>`)
+      req.flash('generic', `Thanks for your feedback`)
```
sfount, line 31: Very minor but these don't need to interpolate, favour `''`

sfount (approved):

PR #2216, opened 2020-08-21
PP-7002 Make error messages better everywhere

stephencdaly:
- Stop passing markup in flash messages.
- Make error messages displayed using `req.flash('genericError',` more consistent and more in line with design system advice.
- Stop rendering error views with the message "Internal server error", instead use the default error message for unexpected errors.
- Modify up some tests along the way to use `sinon.assert` to make the output more meaningful when they fail.

The next step will be to add the title "There is a problem" to all errors shown using `req.flash('genericError', ...)`
However there are still a few more errors which put more markup in the flash message (including links 😬 ). So those need to be addressed first in a separate PR.


app/browsered/field-validation-checks.js
```diff
@@ -10,12 +10,12 @@ const MAX_AMOUNT = 100000
 const validationErrors = {
   required: 'This field cannot be blank',
   mandatoryQuestion: 'You must answer this question',
-  currency: 'Choose an amount in pounds and pence using digits and a decimal point. For example “10.50”',
+  currency: 'Enter an amount in pounds and pence using digits and a decimal point. For example “10.50”',
```
sfount, line 13: 👍

app/controllers/api-keys/post-revoke.controller.js
```diff
@@ -20,8 +21,9 @@ module.exports = (req, res) => {
       req.flash('generic', 'The API key was successfully revoked')
       res.redirect(paths.apiKeys.index)
     })
-    .catch(err => {
-      req.flash('genericError', `<h2>Something went wrong</h2><p>${err}</p>`)
+    .catch(error => {
+      logger.error('Error revoking API key', { error })
```
sfount, line 25: As our log ingestion has a text length limit I wonder if it would be best to just log the message here. This shouldn't be an issue for now (especially if we know what kind of error we're catching) and will likely give other helpful info like a code but I think it would be good to be explicit about what the log should include to avoid overflowing the structured logging.

app/controllers/credentials.controller.js
```diff
@@ -93,11 +93,14 @@ module.exports = {
...
     if (!username) {
-      req.flash('genericError', `<h2>Please enter a valid username</h2>`)
+      req.flash('genericError', `Enter a username`)
     } else if (!password) {
-      req.flash('genericError', `<h2>Please enter a valid password</h2>`)
-    } else if (isPasswordLessThanTenChars(password)) {
-      req.flash('genericError', `<h2>Please enter a valid password</h2> ${isPasswordLessThanTenChars(password)}`)
+      req.flash('genericError', `Enter a password`)
+    } else {
+      const validationError = isPasswordLessThanTenChars(password)
```
sfount, line 100: From the context of the code it looks like this will return a 'reason' string if invalid and null if valid? As this isn't an actual `Error` object potentially naming this `failedValidationMessage` or something more specific would clear this up.

app/controllers/feedback/post-index.controller.js
```diff
@@ -28,15 +28,15 @@ ${req.body['feedback-suggestion']}`
   zendeskClient.createTicket(opts)
     .then(() => {
       lodash.set(req, 'session.pageData.feedback', {})
-      req.flash('generic', `Thanks for your feedback`)
+      req.flash('generic', 'Thanks for your feedback')
```
sfount, line 31: 👍

app/services/user.service.js
```diff
@@ -221,8 +221,11 @@ module.exports = {
...
    * @returns {Promise}
    */
-  updatePhoneNumber: (externalId, newPhoneNumber) => {
-    return getAdminUsersClient().updatePhoneNumberForUser(externalId, newPhoneNumber)
-      .catch(error => new Error(`There has been a problem updating the phone number: ${error}`))
+  updatePhoneNumber: async function (externalId, newPhoneNumber) {
+    try {
+      return await getAdminUsersClient().updatePhoneNumberForUser(externalId, newPhoneNumber)
+    } catch (err) {
+      throw new Error('There has been a problem updating the phone number. Please try again')
```
sfount, line 228: 👍  removing inlining the unknown thrown error

👍  reads clearer 

If this method doesn't actually do anything with the error it feels like this would be more appropriate either: 
* higher up the stack, responding to the request with this message (potentially context dependant) 
* lower down the stack, wrapping the lower level HTTP call to admin users with some kind of formatted error

sfount (approved):
Looks great! 

Cleaning up the redirect errors to remove low level error messages and be more consistent with the messaging is a huge improvement.

app/controllers/api-keys/post-revoke.controller.js
stephencdaly: good point, I'll change this

app/controllers/credentials.controller.js
stephencdaly: :+1:

stephencdaly:
Pushed a new commit with changes in response to comments

PR #2222, opened 2020-08-24
PP-6951 Use custom flash to show refund message

stephencdaly:
Use a custom flash rather than 'generic' and 'genericError' to pass success/failure message with the redirect after a refund request has been processed.

Then in the template for the transaction details page, handle these custom flash messages to display them how we want. This allows for having a title in the message that we were before creating by passing HTML markup in the flash message string.

Add an additional Cypress test for the refund success case.


kbottla (approved):
LGTM!

PR #2227, opened 2020-08-24
PP-7002 Make error messages better everywhere

stephencdaly:
Is a continuation https://github.com/alphagov/pay-selfservice/pull/2216 which was already reviewed, but renamed the branch to point to a different JIRA ticket.

- Stop passing markup in flash messages.
- Make error messages displayed using req.flash('genericError', more consistent and more in line with design system advice.
- Stop rendering error views with the message "Internal server error", instead use the default error message for unexpected errors.
- Modify up some tests along the way to use sinon.assert to make the output more meaningful when they fail.
- The next step will be to add the title "There is a problem" to all errors shown using req.flash('genericError', ...)

However there are still a few more errors which put more markup in the flash message (including links 😬 ). So those need to be addressed first in a separate PR.


sfount (approved):
Really good! 🎉

PR #2230, opened 2020-08-25
PP-6875 Change transaction pagination

kbottla:
## WHAT
- Limits the results count to 100K from ledger to improve performance of the transaction search page. Doing the full count of transactions seems to be the issue for slow response times for services with large volumes. So limiting the total could potentially improve response times. Limit is 100K because we allow CSV download of transactions up to 100K.
- Since we don't provide exact count, we need to change pagination as elements like last_page couldn't be derived.
  Also to make the navigation simpler, paginations links have been changed to just `Previous` & `Next`
  instead of numbered links.
- CSV_MAX_LIMIT env variable is used as the value to limit total as it is related


app/views/transactions/index.njk
```diff
@@ -28,7 +28,11 @@ Transactions - {{currentService.name}} {{currentGatewayAccount.full_type}} - GOV
   {% include "transactions/display-size.njk" %}
 
   <h3 class="govuk-heading-s govuk-!-font-weight-regular govuk-!-margin-top-3" id="total-results">
-    {{totalFormatted}} transactions
+    {% if totalOverLimit %}
+      Over {{csvMaxLimitFormatted}} transactions
+    {% else %}
+      {{totalFormatted}} transactions
+    {% endif %}
```
sfount, line 35: As the `transaction-view` utility has been updated I think this change will also need to be made in the `all-services-transactions/index.njk` as well. 

Unfortunately they use slightly different base templates because of some quirk at the time if I remember correctly; it would be good for us to try and remove that duplication in a future PR if possible.

app/utils/paginator.js
```diff
@@ -168,6 +169,30 @@ Paginator.prototype = {
     return namedRange
   },
 
+  buildNavigationLinks: function (_links, resultsLength) {
```
sfount, line 172: Very minor but this method could be named
```suggestion
  buildNavigationLinks: function buildNavigationLinks(_links, resultsLength) {
```

Also thinking about it there's no reason this needs to use an opaque name like `_links`, it could be more specific, I'll include a potential option but I don't think it's necessarily the right one:
```suggestion
  buildNavigationLinks: function buildNavigationLinks(resultsGeneratedPageLinks, resultsLength) {
```

app/utils/paginator.js
```diff
@@ -23,15 +23,16 @@ Paginator.MAX_PAGE_SIZE = LARGE_PAGE_SIZE
 
 /**
  * @param  {number} pageNumber
- * @param  {number} pageName
+ * @param  {string} pageName
```
sfount, line 26: Good catch 👍  If we're going to keep this up to date I suppose we should include the boolean disabled param

sfount (commented):
Looks great to me! A complex change that's actually simplified the code 👍 

Some minor comments but the only change that I think needs to happen is updating the all service transactions page as both controllers used the shared `transaction-view` utility.

app/utils/paginator.js
kbottla: 👍 Added `disabled` param

app/utils/paginator.js
kbottla: 👍 think I was in the zone of json. Updated!

app/views/transactions/index.njk
kbottla: Good point! updated.

kbottla:
@sfount Thanks for the review. Updates as per comments https://github.com/alphagov/pay-selfservice/compare/1307543586526e6a0becf405df48e402d0c6b3df..0472667161d296618bf3bc4c70abb08497d5c3bb

sfount (approved):

PR #2232, opened 2020-08-25
PP-7002 Display errors better for create payment link journey

stephencdaly:
Previously we were passing validation errors from the POST controller when we redirect to the GET controller in flash messages in a confusing way.
These also contained HTML markup constructed by the controller which is not advised.

Make this work in a more sensible way by putting the validation errors as well as the submitted form values into the session cookie in a "recovered" object. Then in the GET controller after redirect, retrieve from here and delete the object from the session.

Create a nunjucks macro "errorSummary" which takes a list of errors and displays them in the design system error summary component when required. This simplifies the repeated template logic. And can be used in other existing places where we show an error summary.

### DIFF
--- a/app/controllers/payment-links/get-amount.controller.js
+++ b/app/controllers/payment-links/get-amount.controller.js
@@ -7,10 +7,17 @@ const lodash = require('lodash')
 const { response } = require('../../utils/response.js')
 const paths = require('../../paths')
 
-module.exports = (req, res) => {
-  const pageData = lodash.get(req, 'session.pageData.createPaymentLink', {})
-  const paymentLinkAmount = req.body['payment-amount'] || pageData.paymentLinkAmount || ''
-  const paymentAmountType = req.body['amount-type-group'] || pageData.paymentAmountType || ''
+module.exports = function showAmountPage (req, res, next) {
+  const sessionData = lodash.get(req, 'session.pageData.createPaymentLink')
+  if (!sessionData) {
+    next(new Error('Payment link data not found in session cookie'))
+  }
+
+  const recovered = sessionData.amountPageRecovered || {}
+  delete sessionData.amountPageRecovered
+
+  const paymentLinkAmount = recovered.amount || sessionData.paymentLinkAmount || ''
+  const paymentAmountType = recovered.type || sessionData.paymentAmountType || ''
 
   return response(req, res, 'payment-links/amount', {
     paymentLinkAmount,
@@ -18,6 +25,7 @@ module.exports = (req, res) => {
     nextPage: paths.paymentLinks.amount,
     returnToStart: paths.paymentLinks.start,
     manage: paths.paymentLinks.manage,
-    isWelsh: pageData.isWelsh
+    isWelsh: sessionData.isWelsh,
+    errors: recovered.errors
   })
 }
--- a/app/controllers/payment-links/post-amount.controller.js
+++ b/app/controllers/payment-links/post-amount.controller.js
@@ -7,35 +7,38 @@ const lodash = require('lodash')
 const paths = require('../../paths')
 const { safeConvertPoundsStringToPence } = require('../../utils/currency-formatter')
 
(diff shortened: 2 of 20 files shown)


app/controllers/payment-links/get-amount.controller.js
```diff
@@ -7,17 +7,25 @@ const lodash = require('lodash')
 const { response } = require('../../utils/response.js')
 const paths = require('../../paths')
 
-module.exports = (req, res) => {
-  const pageData = lodash.get(req, 'session.pageData.createPaymentLink', {})
-  const paymentLinkAmount = req.body['payment-amount'] || pageData.paymentLinkAmount || ''
-  const paymentAmountType = req.body['amount-type-group'] || pageData.paymentAmountType || ''
+module.exports = function showAmountPage (req, res, next) {
```
sfount, line 10: Very minor but the guidance (you're potentially about to write 😆 ) might suggest doing 
```
function showAmountPage() {
}
module.exports = showAmountPage

File: app/controllers/payment-links/post-amount.controller.js [hunk 2]
```diff
@@ -7,35 +7,38 @@ const lodash = require('lodash')
...
+  const amount = req.body['payment-amount']
+
+  let formattedPaymentLinkAmount = ''
+  const errors = {}
+  if (!type) {
+    errors.type = 'Is the payment for a fixed amount?'
+  } else if (type === 'fixed') {
+    formattedPaymentLinkAmount = safeConvertPoundsStringToPence(amount)
+    if (amount === '' || formattedPaymentLinkAmount === null) {
+      errors.amount = 'Enter an amount in pounds and pence using digits and a decimal point. For example “10.50”'
```
sfount, line 26: So much clearer 👍

app/controllers/payment-links/post-web-address.controller.js
```diff
@@ -12,26 +12,36 @@ const makeNiceURL = string => {
...
-
-  productsClient.product.getByProductPath(updatedPageData.serviceNamePath, updatedPageData.productNamePath)
-    .then(product => {
-    // if product exists we need to alert the user they must use a different URL
-      req.flash('genericError', `<h2>There was a problem with the details you gave for:</h2><ul class="govuk-list govuk-error-summary__list"><li><a href="#payment-name-path">Website address</a></li></ul>`)
-      return res.redirect(paths.paymentLinks.webAddress)
-    })
-    .catch((err) => { // eslint-disable-line handle-callback-err
-    // if it errors then it means no product was found and that’s good
+  const path = req.body['payment-name-path']
```
sfount, line 21: Very minor but path in the context of a top level HTTP controller was ambitious here, something like `paymentLinkURLPath` or something may be more specific

app/views/macro/error-summary.njk
```diff
@@ -0,0 +1,18 @@
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
+
+{% macro errorSummary(params) %}
```
sfount, line 3: Really neat util, this kind of pattern could really clean up the many different ways we have of doing this

app/views/payment-links/amount.njk
```diff
@@ -11,22 +12,30 @@
...
-    {% if flash.errorType[0] === "paymentAmountType" %}
+    {% if errors.type %}
     {% set noSelectionError = {
         text: "Please choose an option"
     } %}
     {% endif %}
 
     {% set variableHTML %}
-      <div class="currency-input govuk-form-group{% if flash.errorType[0] === 'paymentAmountFormat' %} govuk-form-group--error{% endif %}">
+      <div class="currency-input govuk-form-group{% if errors.amount %} govuk-form-group--error{% endif %}">
```
sfount, line 34: So much less room for things to go wrong or not line up 🎉

sfount (approved):
Looks good to me! Some minor comments but not blocking and probably better grouped for a future PR

PR #2235, opened 2020-08-26
PP-6875 Use common transaction view

kbottla:
## WHAT
- Uses common transactions view page for both `all service transactions` and single service transactions routes
    - overrides `beforeContent` block from layout which is different for both service level transactions and all service transactions
    - Added test for multi service view and Refactored transaction lists ui tests setup
- Fixes total results shown.
    - we need to get limit + 1 transactions from ledger to be able to say services has over 100k transactions or over limit


stephencdaly (approved):
Good refactor 👌

PR #2238, opened 2020-08-27
PP-7002 Handle edit payment link conflicts better

stephencdaly:
It is currently only possible to edit one payment link at a time due to how we use the session for this journey. So editing 2 payment links in different browser tabs isn't possible and results in unexpected behaviour currently.

Handle this by:

- If on the summary page for editing the payment link and the ID in the URL doesn't match that in the session, clear the session and show the page for the payment link with id from the URL.
- If trying to visit a page to edit individual details of a payment link and the ID in the URL does not match the ID in the session, put the user back on the "Manage payment links" page with an error message.
- If attempting to make a POST request to update a payment link and the ID in the URL does not match the ID in the session, also put the user back on the "Manage payment links" page with an error message.

When they are put on the "Manage payment links" page, this clears any other session they have, so if there are multiple tabs open, editing on none of them will work. Until they start again on a single tab.

### DIFF
--- a/app/controllers/payment-links/post-edit.controller.js
+++ b/app/controllers/payment-links/post-edit.controller.js
@@ -1,28 +1,26 @@
 'use strict'
 
-// NPM dependencies
 const lodash = require('lodash')
 
-// Local dependencies
-const logger = require('../../utils/logger')(__filename)
 const paths = require('../../paths')
 const productsClient = require('../../services/clients/products.client.js')
-const auth = require('../../services/auth.service.js')
-const { renderErrorView } = require('../../utils/response.js')
 
-module.exports = (req, res) => {
-  const gatewayAccountId = auth.getCurrentGatewayAccountId(req)
+module.exports = async function updatePaymentLink (req, res, next) {
+  const { productExternalId } = req.params
+  const gatewayAccountId = req.account.gateway_account_id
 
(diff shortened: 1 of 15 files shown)


app/controllers/payment-links/post-edit.controller.js
```diff
@@ -1,28 +1,26 @@
...
+    return res.redirect(paths.paymentLinks.manage)
+  }
+
+  try {
+    await productsClient.product.update(gatewayAccountId, productExternalId, editPaymentLinkData)
+    lodash.unset(req, 'session.editPaymentLinkData')
+    req.flash('generic', `Your payment link has been updated`)
+    res.redirect(paths.paymentLinks.manage)
+  } catch (err) {
+    return next(new Error(`Update of payment link failed. Error: ${err.message}`))
```
kbottla, line 24: Are we logging `correlationId` with this ?

kbottla (commented):
Looks good, just a minor comment

app/controllers/payment-links/post-edit.controller.js
stephencdaly: good point, I'll make it so it always logs it

stephencdaly: I've updated `express-unhandled-error-handler.js` in a new commit so that it logs the correlation ID.

kbottla (approved):

PR #2239, opened 2020-08-27
PP-6875 Remove limit on results total from ledger

kbottla:
## WHAT
- Removes limit on total results from ledger as response times slightly increased due to postgres performing sequential scan on counts (for services with large volumnes) and queries cost is almost similar to using indexes.


app/services/clients/ledger.client.js
```diff
@@ -14,7 +14,8 @@ const defaultOptions = {
   baseUrl: process.env.LEDGER_URL,
   json: true,
   service: 'ledger',
-  limit_total_size: (process.env.CSV_MAX_LIMIT || 10000) + 1
+  limit_total_size: (process.env.CSV_MAX_LIMIT || 10000) + 1,
+  limit_total: process.env.LEDGER_ENABLE_LIMIT_TOTAL || false
```
sfount, line 18: Will the client expect a boolean value? Everything straight off of `process.env` will be a string.

This _might_ work depending on how the library interprets the input but it could be that JavaScript type coercion could turn the string "false" from the environment into something truthy. 

To avoid that it might be useful to be explicit here

```javascript
const shouldLimitTotal = process.env.LEDGER_ENABLE_TRANSACTION_COUNT_LIMIT === 'true'

const defaultOptions = {
  //...
  limit_total: shouldLimitTotal
}
```

kbottla: 👍  updated..

sfount (approved):

PR #2240, opened 2020-08-27
PP-7002 Always display generic flash messages as headers

stephencdaly:
When removing markup passed in flash messages, there was a content regression meaning all success messages were displayed with the govuk paragraph formatting, when before usually were either just a header or a header followed by a paragraph.

Make it so that all notifications shown using the 'generic' flash put the content in a h2.

For cases where we want a header and a paragraph, use a different flash key and handle this in the individual templates that show the the notifications.

Create a nunjucks macro for re-use of the notification component.


kbottla (approved):
LGTM!

PR #2241, opened 2020-08-28
PP-6875 Fix pagination navigation filters

kbottla:
## WHAT
- Added missing lastDigitsCardNumber and cardHolderName fields to the forms submitted during navigation and when changing the results size.


stephencdaly (approved):
It's crazy how this has been broken for so long

PR #2247, opened 2020-09-01
PP-7002 Don't pass markup in flash message for 2fa controller

stephencdaly:
Put errors in a 'recovered' object on the session and get them in the GET controller after redirect. Then pass errors into the template and use them to to display the error summary component.

Improve the validation message to indicate that the code may have expired. This new error message is taken from another place we validate 2fa auth messages.

Minor code improvements including using async/await.

### DIFF
--- a/app/controllers/two-factor-auth-controller/get-configure.controller.js
+++ b/app/controllers/two-factor-auth-controller/get-configure.controller.js
@@ -9,22 +9,26 @@ const logger = require('../../utils/logger')(__filename)
 const { response } = require('../../utils/response.js')
 const paths = require('../../paths')
 
-module.exports = (req, res) => {
-  const PAGE_PARAMS = {
-    method: lodash.get(req, 'session.pageData.twoFactorAuthMethod', 'APP')
-  }
+module.exports = async function showConfigureSeconfFactorMethod (req, res) {
+  const method = lodash.get(req, 'session.pageData.twoFactorAuthMethod', 'APP')
+  const recovered = lodash.get(req, 'session.pageData.configureTwoFactorAuthMethodRecovered', {})
+  lodash.unset(req, 'session.pageData.configureTwoFactorAuthMethodRecovered')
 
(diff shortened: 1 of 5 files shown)


SandorArpa (approved):
LOKTM

app/controllers/two-factor-auth-controller/get-configure.controller.js
```diff
@@ -9,22 +9,26 @@ const logger = require('../../utils/logger')(__filename)
 const { response } = require('../../utils/response.js')
 const paths = require('../../paths')
 
-module.exports = (req, res) => {
-  const PAGE_PARAMS = {
-    method: lodash.get(req, 'session.pageData.twoFactorAuthMethod', 'APP')
-  }
+module.exports = async function showConfigureSeconfFactorMethod (req, res) {
```
sfount, line 12: Minor typo here `showConfigureSecondFactorMethod`

app/controllers/two-factor-auth-controller/get-configure.controller.js
```diff
@@ -9,22 +9,26 @@ const logger = require('../../utils/logger')(__filename)
...
-      PAGE_PARAMS.qrCodeDataUrl = url
-      return response(req, res, 'two-factor-auth/configure', PAGE_PARAMS)
-    })
-    .catch(err => {
-      logger.error(`[requestId=${req.correlationId}] Failed to generate QR code - ${err.message}`)
-      req.flash('genericError', 'Something went wrong. Please try again or contact support.')
-      return res.redirect(paths.user.twoFactorAuth.index)
-    })
+  try {
+    const qrCodeDataUrl = await qrcode.toDataURL(otpUrl)
```
sfount, line 21: Much clearer 🎉

app/controllers/two-factor-auth-controller/get-configure.controller.js
stephencdaly: oops

PR #2250, opened 2020-09-02
BAU Remove unneccessary import comments

stephencdaly:
Global replace:
"// NPM dependencies\n" => ""
"// Local dependencies\n" => ""
"// Custom dependencies\n" => ""
"// Node.js core dependencies\n" => ""

Replaced some example API keys in some tests flagged up by the
gds-pre-commit hook with other strings that don't look like secrets.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json",
     "lines": null
   },
-  "generated_at": "2020-08-26T15:22:56Z",
+  "generated_at": "2020-09-02T14:26:57Z",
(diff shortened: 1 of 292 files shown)


sfount (dismissed):

sfount (approved):
There are many examples of where these comments have gotten out of sync with the actual organisation of the file and it seems to lead to more confusion that it helps resolve.

In general this should be very clear from the actual imports themselves, roughly following the layout:
* Node core library imports
* third party imports 
* local imports 
* global state declaration

PR #2258, opened 2020-09-07
PP-7018 Enable limit on total count from Ledger

kbottla:
## WHAT 
- Enables limiting the total count for search results and sets to 5K.
- As this is much lower limit, showing/hiding CSV download link is not possible due to total count (100K) needed for CSVs.So we will be allowing CSV downloads all the time
   - Only exception is services with total transactions above 5K need to apply at least one filter (other than page number and page size params) to be able to download CSV.


app/utils/transaction-view.js
```diff
@@ -193,3 +192,13 @@ function hasPageSizeLinks (connectorData) {
...
 }
+
+function showCsvDownload (connectorData, filters) {
+  if (connectorData.total <= LEDGER_TRANSACTION_COUNT_LIMIT) {
+    return true
+  }
+
+  return Object.keys(filters).filter(function (key) {
+    return key !== 'page' && key !== 'pageSize'
+  }).length !== 0
```
sfount, line 203: Very minor but this is probably sufficiently complex that naming it would help readability

```javascript
const hasUserFilteredResults = Object.keys(filters)
  .filter((key) => !['page', 'pageSize'].includes(key))
  .length

return hasUserFilteredResults
```

sfount (approved):
LGTM 👍 

I believe this release depends on the release of https://github.com/alphagov/pay-ledger/pull/912

app/utils/transaction-view.js
kbottla: Done this and IDE complained to make it inline 😄

kbottla: Updated in https://github.com/alphagov/pay-selfservice/pull/2259/files#diff-bc5732d3a5d7c7ae7ed3dfa0a9893fb2R201

PR #2259, opened 2020-09-07
PP-7018 Fix pagination filters

kbottla:
## WHAT
- Moves hidden search filters to common nunjucks file so pagination changes can be consistent
- Fixes pagination when transactions are filtered by single card brand


SandorArpa (approved):
LOKTM

PR #2260, opened 2020-09-07
BAU Fix displaying errors for forgotten password controller

stephencdaly:
Currently, if an email address is entered that no account is found for, we would attempt to redirect to the URL for resetting the password resulting in a confusing error message being displayed.

Instead, handle a 404 from adminusers by displaying a sensible error message and redirecting to the same page to enter an an email address to request a password reset.

Fix this and also fix the following:
- Display error messages properly using the error summary component and inline error messages.
- Make the controller easier to read by using async/await.
- Don't use anonymous functions for controllers.
- Move password validation out of the service and into the common server-side-form-validations module so we can distinguish between errors from adminusers and errors from selfservice. While doing this, make a few error messages better.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json",
     "lines": null
   },
-  "generated_at": "2020-09-02T14:26:57Z",
+  "generated_at": "2020-09-07T17:13:42Z",
   "plugins_used": [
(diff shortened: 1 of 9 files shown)


kbottla (approved):
LGTM! Fix looks good and nice refactoring 🎉 🎉

PR #2262, opened 2020-09-09
BAU Don't tell user account doesn't exist for password reset

stephencdaly:
Prevent email enumeration attempts by not telling users that an account
doesn't exist when an email address is entered for a non-existent
account.

Instead, show that page that says we have sent an email.

### DIFF
--- a/test/integration/forgotten-password.ft.test.js
+++ b/test/integration/forgotten-password.ft.test.js
@@ -47,7 +47,7 @@ describe('forgotten_password_controller', function () {
       .reply(200)
 
     await forgottenPasswordController.emailPost(req, res)
-    sinon.assert.called(res.redirect)
+    sinon.assert.calledWith(res.redirect, '/reset-password-requested')
   })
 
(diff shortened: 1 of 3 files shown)


test/integration/forgotten-password.ft.test.js
```diff
@@ -64,7 +64,7 @@ describe('forgotten_password_controller', function () {
     })
   })
 
-  it('send render the page with errors if adminusers returns a 404', async () => {
+  it('show the password reset email sent page if account is not found in adminusers', async () => {
```
alexbishop1, line 67: Should we maybe have a test if there’s something we do expect to show an error for (e.g. adminusers returns a 500)?

alexbishop1 (changes requested):
Looks good: just one request related to the tests.

test/integration/forgotten-password.ft.test.js
stephencdaly: sure, i'll add one

alexbishop1 (approved):

PR #2264, opened 2020-09-11
PP-7098 Add label to heading for Stripe Onboarding > VAT Numbe page > H1

iqbalgds:

### DIFF
--- a/app/views/stripe-setup/vat-number/index.njk
+++ b/app/views/stripe-setup/vat-number/index.njk
@@ -19,7 +19,7 @@
         errorList: errorList
       }) }}
     {% endif %}
-    <h1 class="govuk-heading-l govuk-!-margin-bottom-6">What is your organisation’s VAT number?</h1>
+    <label for="vat-number"><h1 class="govuk-heading-l govuk-!-margin-bottom-6">What is your organisation’s VAT number?</h1></label>
 
     <form id="vat-number-form" method="post"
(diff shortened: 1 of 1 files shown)


SandorArpa (approved):
LOKTM

app/views/stripe-setup/vat-number/index.njk
```diff
@@ -19,7 +19,7 @@
         errorList: errorList
       }) }}
     {% endif %}
-    <h1 class="govuk-heading-l govuk-!-margin-bottom-6">What is your organisation’s VAT number?</h1>
+    <label for="vat-number"><h1 class="govuk-heading-l govuk-!-margin-bottom-6">What is your organisation’s VAT number?</h1></label>
```
kbottla, line 22: not sure heading inside label will work well. This page suggests otherwise https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label. We could style label as heading using https://design-system.service.gov.uk/get-started/labels-legends-headings/

iqbalgds: Good spot - as this what DAC told us to do.
Your link from MDN contradicts this:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#Accessibility_concerns
I think we need a H1 for the page.
So maybe need to add a new label just for the field.

PR #2276, opened 2020-09-16
PP-7095 DAC Audit - Make tables in `DL` list

iqbalgds:
- Update table in demo payment
  - Convert to a DL list
- Update table in `My Profile` page
  - Convert into a DL list
  - Make the DL list have CSS `display: table` - as this allows us to us the design system table styling.
- Update `change` links to include hidden spans - so they are more descriptive for screen readers

### DIFF
--- a/app/assets/sass/application.scss
+++ b/app/assets/sass/application.scss
@@ -42,4 +42,5 @@ $govuk-page-width: 1200px;
 @import "components/reporting-columns";
 @import "components/account-status-panel";
 @import "components/psp-go-live-panel";
-
+@import "components/demo-payment";
+@import "components/my-profile";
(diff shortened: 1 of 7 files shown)


kbottla (approved):
Changes looks good to me. But https://cdpn.io/aardrian/debug/NzGaKP lists different interpretation by assistive technologies and not sure if definitive tags introduced behave differently or doesn't work as expected

iqbalgds:
> Changes looks good to me. But https://cdpn.io/aardrian/debug/NzGaKP lists different interpretation by assistive technologies and not sure if definitive tags introduced behave differently or doesn't work as expected

Useful link - thanks
Looking at the DAC report, hopefully this should be better then the existing tables

PR #2279, opened 2020-09-17
PP-7195 - Update Node to 12.18.4

nimalank7:
Description:
- Following a security patch, update Node version to 12.18.4
- Note that the SHA for DockerHub (sha256:f48e5c326a69822257cbafe720e2939b9ffc3803d73ea03c3ff0e9c9f9a7b74a) doesn't match docker pull (sha256:59fa78a2149e3470ba7346fb17938e2c48e17096006083003ee1673cc172d676)

**To review this PR:**

- [ ] Verify that the SHA hashes match what is available through dockerhub
- [ ] Everything is happy in CI/ test environments


docker/build_and_test.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node@sha256:2ab3d9a1bac67c9b4202b774664adaa94d2f1e426d8d28e07bf8979df61c8694
+FROM node:12.18.4@sha256:59fa78a2149e3470ba7346fb17938e2c48e17096006083003ee1673cc172d676
```
alexbishop1, line 1: This needs to be `FROM node:12.18.4-alpine3.12@sha256:59fa78a2149e3470ba7346fb17938e2c48e17096006083003ee1673cc172d676` (previous upgrade accidentally switched from Alpine to non-Alpine).

You’ll also need to change `python` to `python2`.

package.json
```diff
@@ -5,7 +5,7 @@
   "private": true,
   "license": "MIT",
   "engines": {
-    "node": "^12.16.3"
+    "node": "^12.18.4"
```
alexbishop1, line 8: This can only be 12.18.3 because that’s the most recent version the PaaS supports. From https://groups.google.com/a/digital.cabinet-office.gov.uk/g/gov-uk-paas-announce/c/aiF05_XAH8I:

```
## nodejs_buildpack: v1.7.26

* node:
    * Removed: 10.20.1, 12.18.0, 14.4.0, 14.5.0
    * Added: 12.18.3, 14.7.0, 14.8.0, 10.22.0
```

alexbishop1 (changes requested):

docker/build_and_test.Dockerfile
nimalank7: Done

package.json
nimalank7: Done

alexbishop1 (approved):
Apparently the different SHAs are caused by this fun issue: https://github.com/docker/hub-feedback/issues/1925

PR #2289, opened 2020-09-21
PP-7093 DAC Audit - Dashboard Date range - remove auto reload

iqbalgds:
- When users change the data range select on the dashboard - they now have to press a `Update` button.
- The page does not auto reload when changing a value in the date range select.

*Screenshot*
image


app/views/dashboard/_activity.njk
```diff
@@ -32,12 +36,10 @@
...
   }) }}
 
-  <noscript>
-    {{ govukButton({
-      classes: 'nojs-update-button',
-      text: "Update"
-    }) }}
-  </noscript>
+  {{ govukButton({
+    classes: 'dashboard-date-range__select',
```
sfount, line 40: Very minor but inconsistent use of `'` and `"` isn't ideal here

sfount (approved):
This looks like it resolves the issue it sets out to fix, looks good to me 👍 . Potentially the UX can be reviewed if this page is iterated on in the future

app/views/dashboard/_activity.njk
iqbalgds: Thanks - I will update that now

app/views/dashboard/_activity.njk
```diff
@@ -32,12 +36,10 @@
...
   }) }}
 
-  <noscript>
-    {{ govukButton({
-      classes: 'nojs-update-button',
-      text: "Update"
-    }) }}
-  </noscript>
+  {{ govukButton({
+    classes: "dashboard-date-range__select",
```
iqbalgds, line 40: Updated to double quotes

sfount (approved):

PR #2296, opened 2020-09-23
PP-7214 "Debounce" refund button submission

sfount:
Add the `preventDoubleClick` attribute to the refund payment button.
This "debounces" event propagation for clicks for 1 second after the
button is pressed.

This can only prevent subsequent clicks for 1 second.

This is a temporary fix which will help solve double clicking however as
this is a terminal POST request a further script should be added to
fully disable the button until the action is completed.


danworth (approved):
Seems like a sensible temporary solution 👍

PR #2307, opened 2020-09-29
PP-7288 Refactor create user account part of self registration journey

stephencdaly:
- Improve error messages to make them more better/more consistent with error messages we show in other places.
- Don't directly display errors originating from adminusers to the user, instead identify errors based on the status code and display messages defined in selfservice.
- Use the `server-side-form-validations.js` module for all validation, working towards retiring the `registration-validations.js` module.

### DIFF
--- a/app/controllers/register-service.controller.js
+++ b/app/controllers/register-service.controller.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const _ = require('lodash')
+const lodash = require('lodash')
 
 const logger = require('../utils/logger')(__filename)
 const paths = require('../paths')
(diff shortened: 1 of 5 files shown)


app/controllers/register-service.controller.js
```diff
@@ -35,60 +40,62 @@ module.exports = {
    * @param req
    * @param res
    */
-  submitRegistration: (req, res) => {
+  submitRegistration: async function submitRegistration (req, res) {
```
sfount, line 43: This method should be hung up on the walls as an example of taking something super complex and making it easy to understand. Really good change 👍

sfount (approved):
Huge improvement in legibility 👍 I haven't tested this personally but all of the tests are happy and coverage looks good

PR #2308, opened 2020-09-30
PP-7288 Refactor OTP part of self registration journey

stephencdaly:
- Improve error messages to make them more better/more consistent with error messages we show in other places.
- Use the `server-side-form-validations.js` module for all validation, working towards retiring the `registration-validations.js` module.
- Remove `otp-verify.js` middleware, as this logic makes more sense being in the controller.


app/controllers/register-service.controller.js
```diff
@@ -129,22 +135,51 @@ module.exports = {
...
+            verificationCode: 'The verification code you’ve used is incorrect or has expired'
+          }
+        }
+        res.redirect(303, paths.selfCreateService.otpVerify)
+      } else if (err.errorCode === 410) {
+        return renderErrorView(req, res, 'This invitation is no longer valid', 410)
+      } else {
+        return renderErrorView(req, res, 'Unable to process registration at this time', err.errorCode || 500)
+      }
+    }
```
stephencdaly, line 169: The above code for validating the OTP code was previously in the `otp-verify.js` middleware, which was very confusing.

app/middleware/otp-verify.js
```diff
@@ -1,59 +0,0 @@
...
-      logger.warn(`[requestId=${req.correlationId}] Invalid invite code attempted ${req.code}, error = ${err.errorCode}`)
-      switch (err.errorCode) {
-        case undefined:
-          handleInvalidOtp(err.message)
-          break
-        case 401:
-          handleInvalidOtp('Invalid verification code')
-          break
-        case 404:
-          errorResponse(req, res, 'Unable to process registration at this time', 404)
```
sfount, line 50: As discussed, adding assertions to the existing tests (where possible) on this branching logic would be good. Otherwise everything looks great

app/controllers/register-service.controller.js
```diff
@@ -129,22 +135,51 @@ module.exports = {
...
+
+    try {
+      const completeInviteResponse = await registrationService.createPopulatedService(req.register_invite.code, correlationId)
+      loginController.setupDirectLoginAfterRegister(req, res, completeInviteResponse.user_external_id)
+      return res.redirect(303, paths.selfCreateService.logUserIn)
+    } catch (err) {
+      if (err.errorCode === 409) {
+        const error = (err.message && err.message.errors) ? err.message.errors : 'Unable to process registration at this time'
```
sfount, line 177: Having `err` and `error` in the same namespace made me concerned for a minute, the logic all looks sound though! Maybe `errorMessage` best describes what this is and helps differentiate from the exception

app/utils/validation/server-side-form-validations.test.js
```diff
@@ -271,4 +271,24 @@ describe('Server side form validations', () => {
...
+
+  describe('otp code validation', () => {
+    it('should be valid for a valid OTP code', () => {
+      expect(validations.validateOtp('123').valid).to.be.true // eslint-disable-line
+    })
+
+    it('should not be valid for empty OTP code', () => {
+      expect(validations.validateOtp('')).to.deep.equal({
+        valid: false,
+        message: 'Enter your verification code'
```
sfount, line 283: 👍

sfount (approved):
Great refactors 👍

app/controllers/register-service.controller.js
```diff
@@ -175,35 +210,34 @@ module.exports = {
    * @param req
    * @param res
    */
-  submitOtpResend: function submitOtpResend (req, res) {
+  submitOtpResend: async function submitOtpResend (req, res) {
+    const sessionData = req.register_invite
     const correlationId = req.correlationId
-    const code = req.register_invite.code
+    const code = sessionData.code
```
sfount, line 216: Similar to the above on trusting the session is well formed for this to not NPE https://github.com/alphagov/pay-selfservice/pull/2308#discussion_r497490462

app/controllers/register-service.controller.js
stephencdaly: 👍 will rename in the next PR

app/controllers/register-service.controller.js
```diff
@@ -118,8 +118,14 @@ module.exports = {
    * @param req
    * @param res
    */
-  showOtpVerify: function showOtpVerify (req, res) {
-    res.render('self-create-service/verify-otp')
+  showOtpVerify: function showOtpVerify (req, res, next) {
+    const sessionData = req.register_invite
+    const recovered = sessionData.recovered || {}
```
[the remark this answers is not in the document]
stephencdaly: 👍 will move logic out of dodgy middleware to here in next PR

PR #2309, opened 2020-09-30
PP-7228 Remove middleware that checks registration cookie

stephencdaly:
We should check we have the data in the session where we access it in the controller rather than in this middleware. Move logic to controller.


SandorArpa (approved):
LOKTM

PR #2310, opened 2020-09-30
PP-7228 Refactor invite registration controller

stephencdaly:
- Improve error messages to make them better/more consistent with error messages we show in other places.
- Use the `server-side-form-validations.js` module for all validation, remove the `registration-validations.js` module.
- Rename `validate-invite.controller.ft.test.js` to `invite-validation.controller.ft.test.js` to match controller it tests. Remove duplicate tests that are already in `register-user.controller.ft.test.js`

## How to review

I wouldn't get too hung up on how the code used to be, as it's very convoluted and hard to understand. More that the new code makes sense.
I have manually validated that all paths of the journey work with local testing.


app/utils/registration-validations.js
```diff
@@ -1,89 +0,0 @@
...
-
-      if (!password || password.length < MIN_PASSWORD_LENGTH) {
-        reject(new Error('Password must be 10 characters or more'))
-      } else if (commonPassword(password)) {
-        reject(new Error('The password you tried to create contains a common phrase or combination of characters. Choose something that’s harder to guess.'))
-      } else {
-        resolve()
-      }
-    })
-  }
```
sfount, line 88: 🎉

sfount (approved):
LGTM 👍 I believe different paths replaced by removing the validator utility are covered by previous PRs/ existing tests. This could is a lot simpler and easier to understand

PR #2311, opened 2020-10-01
PP-7228 Clean up registration controllers

stephencdaly:
- Move exports to bottom of file
- Use async/await in remaining places

Tip: Set github diff to ignore whitespace changes. Not that much actually changes, just indentation.


alexbishop1 (approved):

rorymalcolm (approved):
👍

PR #2336, opened 2020-10-28
PP-7306 add is internal user

SandorArpa:
## WHAT
- add isInternalUser function to User object. This is to help when doing structured logging to filter out internal user from statistical dashboards

### DIFF
--- a/app/models/User.class.js
+++ b/app/models/User.class.js
@@ -47,6 +47,7 @@ class User {
     this.features = (userData.features || '').split(',').map(feature => feature.trim())
     this.secondFactor = userData.second_factor
     this.provisionalOtpKey = userData.provisional_otp_key || ''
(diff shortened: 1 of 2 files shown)


app/models/User.class.js
```diff
@@ -125,6 +126,14 @@ class User {
   getPermissionsForService (serviceExternalId) {
     return _.get(this.getRoleForService(serviceExternalId), 'permissions', []).map(permission => permission.name)
   }
+
+  /**
+   * @method isInternalUser
+   * @returns {boolean}
+   */
+  isInternalUser () {
```
stephencdaly, line 134: Can just access the `.internalUser` property from outside this class without the need fo a getter method. That's what we usually do

stephencdaly (approved):
👍

PR #2337, opened 2020-10-28
PP-7306 add number of live services to user object

SandorArpa:
## WHAT
- add numberOfLiveServices to User object. This is needed for structured logging for statistical purposes

### DIFF
--- a/test/unit/models/User.class.test.js
+++ b/test/unit/models/User.class.test.js
@@ -66,15 +66,50 @@ describe('Class: User', () => {
   describe('is internal user', () => {
     it('should return false', () => {
       process.env.GDS_INTERNAL_USER_EMAIL_DOMAIN = '@example.org'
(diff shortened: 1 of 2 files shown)


test/unit/models/User.class.test.js
```diff
@@ -66,15 +66,50 @@ describe('Class: User', () => {
...
+        }
+      }, {
+        service: {
+          name: 'System Generated',
+          external_id: 'externalServiceId4',
+          current_go_live_stage: 'LIVE'
+        }
+      }]
+    }
+    it('should ', function () {
```
stephencdaly, line 109: looks like you forgot to add a test name

stephencdaly (commented):
Looks good, just a minor thing

test/unit/models/User.class.test.js
SandorArpa: :face_palm

stephencdaly (approved):

PR #2343, opened 2020-11-02
PP-7306 Add logging for all live service transactions

kbottla:
## WHAT
- Added logging when all service transactions page is accessed and also flags to indicate if downloads are for all live service transactions and internal user

### DIFF
--- a/app/controllers/all-service-transactions/get.controller.js
+++ b/app/controllers/all-service-transactions/get.controller.js
@@ -11,6 +11,8 @@ const { getFilters, describeFilters } = require('../../utils/filters.js')
 const router = require('../../routes.js')
 const states = require('../../utils/states')
 const client = new ConnectorClient(process.env.CONNECTOR_URL)
+const logger = require('../../utils/logger')(__filename)
+const { keys } = require('@govuk-pay/pay-js-commons').logging
 
 const { CORRELATION_HEADER } = require('../../utils/correlation-header.js')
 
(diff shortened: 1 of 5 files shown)


app/controllers/payouts/payout-list.controller.js
```diff
@@ -18,9 +18,12 @@ const listAllServicesPayouts = async function listAllServicesPayouts (req, res)
       return
     }
     const payoutSearchResult = await payoutService.payouts(userPermittedAccountsSummary.gatewayAccountIds, req.user, page)
-    const logContext = { current_page: page }
+    const logContext = {
+      current_page: page,
+      internal_user: req.user.internalUser,
+      gateway_account_ids: userPermittedAccountsSummary.gatewayAccountIds
```
stephencdaly, line 24: could we also log the number of live services here?

app/controllers/all-service-transactions/get.controller.js
```diff
@@ -20,6 +22,15 @@ module.exports = async (req, res) => {
   try {
     const userPermittedAccountsSummary = await permissions.getLiveGatewayAccountsFor(req.user, 'transactions:read')
 
+    const logContext = {
+      gateway_account_ids: userPermittedAccountsSummary.gatewayAccountIds,
+      user_number_of_live_services: userPermittedAccountsSummary.gatewayAccountIds && userPermittedAccountsSummary.gatewayAccountIds.length,
```
stephencdaly, line 27: We already have this number on the User object. Wonder whether we should use that so that if we end up loging this in multiple places it can be uniform https://github.com/alphagov/pay-selfservice/blob/master/app/models/User.class.js#L53

app/services/transaction.service.js
```diff
@@ -52,6 +53,8 @@ const logCsvFileStreamComplete = function logCsvFileStreamComplete (timestampStr
     method: 'future',
     gateway_account_ids: gatewayAccountIds,
     multiple_accounts: gatewayAccountIds.length > 1,
+    internal_user: user.internalUser,
+    all_live_service_transactions: allLiveServiceTransactions,
```
stephencdaly, line 57: can we log the number of live services for the user here too?

app/controllers/all-service-transactions/get.controller.js
kbottla: Didn't realise we have this field on User object. Updated to use User property

app/services/transaction.service.js
kbottla: Updated

app/controllers/payouts/payout-list.controller.js
kbottla: Updated!

stephencdaly (approved):
👍

PR #2351, opened 2020-11-10
PP-5236 Cleanup token creation controllers

sfount:
Update controllers that deal with creating potentially external API
keys. Use more up to date Node guidance to make the code clearer.

* Use clearer variable naming
* Use async/ await
* Handle errors using `next` and middleware where appropriate
* Name any large anonymous methods
* Simplify object construction


stephencdaly (approved):
Nice cleanup

PR #2353, opened 2020-11-13
PP-6129 Specify token "source" for token creation

sfount:
Self service creates tokens for 4 main reasons:
* generating card API tokens for users
* creating new payment links
* demo payments feature
* test with your users feature

In each of these cases we should be passing a "type" parameter, this
translates to a public auth `TokenSource` which tells public auth if
this api key is intended to be used for an API client or internally as
one of our products.

Payment links and the two testing features we provide all work by
creating products and running test payments through those. As the public
auth token will be used by a "product" this type should be specified.

Update the generate API key route to no longer rely on the default
source (API).


SandorArpa (approved):
LGTM

PR #2354, opened 2020-11-13
PP-7207 DAC Audit - Transactions > Multi-select for Card Brand & Payment Status

iqbalgds:
- Add JS to toggle 'aria-expanded=true/false' on the button
- Fix ID of the button to match the label
- Update Cypress tests to use the new ID for:
  - Transaction > Card brand filter multi-select
  - Transaction > Status filter multi-select


kbottla (approved):
LGTM!

PR #2364, opened 2020-11-17
PP-7207 DAC Audit > Transactions > Multi-select bug updates

iqbalgds:
- Fix aria-expanded - so it is only set on the button.
- Add an extra 'fieldset' tag to wrap the container for the checkboxes.


SandorArpa (approved):

PR #2366, opened 2020-11-19
PP-7409 Add test vs. live content for payment link created message

sfount:
Payment links created in test and live environments have very different
use cases; the wording for the payment link creation references the
payment link being "live" which conflicts with our use of this status
for accounts.

Add a separate path for test and live accounts to provide users
additional information if they are just testing.

**Payment link created in a test account**

Screenshot 2020-11-19 at 12 06 30

**Payment link created in a live account**

Screenshot 2020-11-19 at 12 08 33

### DIFF
--- a/app/views/payment-links/manage.njk
+++ b/app/views/payment-links/manage.njk
@@ -12,10 +12,26 @@
 {% block mainContent %}
 <section class="govuk-grid-column-two-thirds">
 {% if flash.createPaymentLinkSuccess %}
-    {{ successNotification({
-      heading: "Your payment link is now live",
-      body: "Give this link to your users to collect payments for your service."
-    }) }}
+  {% if isTestGateway %}
+    {% set heading = "Your test payment link is ready to use" %}
+    {% set body %}
+      <span>
+        To collect payments from your users with this link you will need to
+        <a class="govuk-link" href="{{ paths.generateRoute(paths.requestToGoLive.index, { externalServiceId: externalServiceId }) }}">request to go live</a>
+        and create the payment link again.
+      </span>
+    {% endset %}
+  {% else %}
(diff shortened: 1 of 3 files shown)


kaydale:
Looks good to me

app/views/payment-links/manage.njk
```diff
@@ -12,10 +12,26 @@
...
 {% if flash.createPaymentLinkSuccess %}
-    {{ successNotification({
-      heading: "Your payment link is now live",
-      body: "Give this link to your users to collect payments for your service."
-    }) }}
+  {% if isTestGateway %}
+    {% set heading = "Your test payment link is ready to use" %}
+    {% set body %}
+      <span>
+        To collect payments from your users with this link you will need to
```
kbottla, line 19:
```suggestion
        To collect payments from your users you will need to  ?
```

sfount: As there's no line breaking content in this span everything in the block will be treated as the same line, they're on separate lines here for readability as the link in particular is long

kbottla: I was trying to convey something using suggestion but clearly not done well :) . Do we need `with this link` as it won't be same link after live ?

sfount: Ah! My bad, I thought the question mark was part of the suggestion 🤦

kbottla (approved):
LGTM!

app/views/payment-links/manage.njk
iqbalgds: Content has been updated.

kbottla (approved):
LGTM!

kbottla (approved):

PR #2367, opened 2020-11-19
PP-7091 DAC Audit - Multi-select - Close multi-selects when escape is…

iqbalgds:
… pressed.

- Add keypress handler to the document.body
- Detects if anyone presses `escape`
  - Closes any open `multi selects`
  - Sets focus to the multi select `open button`.


SandorArpa (approved):

PR #2374, opened 2020-11-24
PP-7416 Allow selfservice to set 3DS Secure integration version

alexbishop1:
Add support for sending a PATCH request to `/v1/api/accounts/{accountId}` on connector with a body like this:

```json
{
  "op": "replace",
  "path": "integration_version_3ds",
  "value": 2
}
```


nimalank7 (approved):
LGTM!

PR #2376, opened 2020-11-25
PP-7416 Add controller to enable/disable Worldpay 3DS Flex

alexbishop1:
Add controller to enable or disable Worldpay 3DS Flex. Not actually hooked up to a page yet.

### DIFF
--- a/app/controllers/your-psp/post-toggle-worldpay-3ds-flex.controller.js
+++ b/app/controllers/your-psp/post-toggle-worldpay-3ds-flex.controller.js
@@ -1,0 +1,26 @@
+'use strict'
+
+const paths = require('../../paths')
+const { renderErrorView } = require('../../utils/response')
+const { ConnectorClient } = require('../../services/clients/connector.client')
+const connector = new ConnectorClient(process.env.CONNECTOR_URL)
+
+module.exports = async function toggleWorldpay3dsFlex (req, res) {
+    const accountId = req.account.gateway_account_id
+    const toggleWorldpay3dsFlex = req.body['toggle-worldpay-3ds-flex']
+
+    if (req.body['toggle-worldpay-3ds-flex'] === 'on' || req.body['toggle-worldpay-3ds-flex'] === 'off') {
+        const enabling3dsFlex = toggleWorldpay3dsFlex === 'on'
+        const message = enabling3dsFlex ? '3DS Flex turned on' : '3DS Flex turned off'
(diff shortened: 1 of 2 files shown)


app/controllers/your-psp/post-toggle-worldpay-3ds-flex.controller.js
```diff
@@ -0,0 +1,26 @@
+'use strict'
+
+const paths = require('../../paths')
+const { renderErrorView } = require('../../utils/response')
+const { ConnectorClient } = require('../../services/clients/connector.client')
+
+module.exports = async (req, res) => {
```
oswaldquek, line 7: This function should have a name for clarity, e.g. ```module.exports = async toggleWorldpay3ds(req, res) { ... }```

app/controllers/your-psp/post-toggle-worldpay-3ds-flex.controller.js
```diff
@@ -0,0 +1,26 @@
...
+    const connector = new ConnectorClient(process.env.CONNECTOR_URL)
```
oswaldquek, line 8: Am not too familiar with the implications of Node but in Java I would move this line outside of the function.

oswaldquek (commented):
Looks good, just some code quality requests.

app/controllers/your-psp/post-toggle-worldpay-3ds-flex.controller.js
alexbishop1: I was just following the style of an existing file but I see we define it outside the function in other files, so I’ll do that here too.

app/controllers/your-psp/post-toggle-worldpay-3ds-flex.controller.js
alexbishop1: Thanks, I’ll do that.

nimalank7 (changes requested):
Our node JS style guide recommends that we don't export anonymous functions. Could we change `module.exports = async (req, res)` to `module.exports = async function newController(req, res)`

app/controllers/your-psp/post-toggle-worldpay-3ds-flex.controller.js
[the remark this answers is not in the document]
alexbishop1: Yes, done that.

nimalank7 (approved):
LGTM!

PR #2377, opened 2020-11-26
BAU Move client instantiation out of requests

sfount:
Large objects shouldn't be instantiated in code that is called frequently, where it can be helped (for example, during the lifecycle of a request).

Node modules and exports can be considered singletons by default, use this to allow clients to be instantiated once and called as many times as required.

Hopefully this can be raised and followed as general practice.


sfount:
It looks like anything relying on the old base client `on` event propagation relies on the client instance being unique; this will need to unpicked and re-factored to use promise based propagation. 

This should be removed from this PR where it applies with a comment added; it could be done as a follow up.

oswaldquek (approved):
LGTM

sfount:
This was done by @kbottla in a number of PRs made against `PP-7585`

https://github.com/alphagov/pay-selfservice/pull/2461

PR #2378, opened 2020-11-26
PP-7305 My Services > Fix page width issue.

iqbalgds:
- Update Nunjucks template to set page width to 'two thirds'
- Update Flash.njk so that it only add the container when there is a flash message.

### DIFF
--- a/app/views/includes/flash.njk
+++ b/app/views/includes/flash.njk
@@ -1,18 +1,21 @@
 {% from "../macro/success-notification.njk" import successNotification %}
 
 {% block flash %}
-  <div class="govuk-grid-column-two-thirds">
-    {% if flash.genericError %}
-    <div class="flash-container">
-      <div class="notification generic-error error-summary">
-        {{ flash.genericError | safe }}
+  {% if flash.genericError %}
+    <div class="govuk-grid-column-two-thirds">
+      <div class="flash-container">
+        <div class="notification generic-error error-summary">
+          {{ flash.genericError | safe }}
+        </div>
       </div>
     </div>
-    {% endif %}
-    {% if flash.generic %}
-    {{ successNotification({
-      heading: flash.generic | safe
-    }) }}
-    {% endif %}
-  </div>
+  {% endif %}
+
+  {% if flash.generic %}
+    <div class="govuk-grid-column-two-thirds">
+      {{ successNotification
+        ({heading: flash.generic | safe}) 
+      }}
+    </div>
+  {% endif %}
 {% endblock %}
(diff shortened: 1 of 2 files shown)


app/views/services/index.njk
```diff
@@ -5,61 +5,63 @@
...
+          View payments to your bank account
+        </a>
+      </p>
+      {% endif %}
+    </div>
+
+    <div class="govuk-grid-column-full">
+      {% if services.length > 7 %}
+        <div class="js-show flex-grid--row">
+          <div class="flex-grid--column-half tight">
```
stephencdaly, line 50: could we make this two-thirds for the search box and one third for "Clear filter" so the box is wider?

app/views/services/index.njk
```diff
@@ -5,61 +5,63 @@
...
-    {% endif %}
-    {% if has_account_with_payouts %}
-    <p class="govuk-body">
-      <a href="{{routes.payouts.list}}" class="govuk-link govuk-!-margin-right-3 govuk-link--no-visited-state">
-        View payments to your bank account
-      </a>
-    </p>
-    {% endif %}
-  </div>
+    <h1 class="govuk-heading-l govuk-grid-column-full govuk-!-margin-bottom-3 service-count">
```
stephencdaly, line 17: can we clean this up to remove all the `govuk-grid-column-full` from all the nested elements, as they're now inside a `govuk-grid-column-two-third` container?

app/views/includes/flash.njk
```diff
@@ -1,18 +1,18 @@
...
+      }) }}
     </div>
-    {% endif %}
-    {% if flash.generic %}
-    {{ successNotification({
-      heading: flash.generic | safe
-    }) }}
-    {% endif %}
-  </div>
+  {% endif %}
```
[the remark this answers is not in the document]
stephencdaly: Also, was the flash container causing issues with the display of the "My services" page? What happens if a notification is displayed to the user on the page?

stephencdaly (changes requested):

app/views/includes/flash.njk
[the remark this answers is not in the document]
iqbalgds: Good spot - I have fixed the `div` tags.

I noticed that the Flash container was always leaving an empty `<div class="govuk-grid-column-two-thirds">
  </div>` - even if there was not a Flash message.  I thought I would clean it up in case it cases any layout issues.

Now the `div` will only appear if there is a flash message.

app/views/services/index.njk
iqbalgds: Good spot - done.

app/views/services/index.njk
iqbalgds: Good spot - done.

stephencdaly (approved):
Looks good to me!

PR #2386, opened 2020-12-01
PP-7458 Refactor service-registration.service.js

stephencdaly:
- Move exports to bottom of file
- Name anonymous functions
- Remove function comment blocks that don't add anything
- Use async/await
- Remove unnecessary Promise creation
- Inline single line functions that are used internally in one place

## Note
Hiding whitespace changes will help review


SandorArpa (approved):
LOKTM

PR #2387, opened 2020-12-01
PP-7458 Log when new service is created

stephencdaly:
Log when new service is created:
- as part of user registration
- when an existing user adds a new service

Refactor surrounding code to conform to best practices.

For the create service response from adminusers, transform the response using the `Service` model, like we do for other responses.

### DIFF
--- a/app/controllers/create-service.controller.js
+++ b/app/controllers/create-service.controller.js
@@ -2,7 +2,7 @@
 
 const lodash = require('lodash')
 
-const { renderErrorView, response } = require('../utils/response')
+const { response } = require('../utils/response')
 const paths = require('../paths')
 const serviceService = require('../services/service.service')
 const userService = require('../services/user.service')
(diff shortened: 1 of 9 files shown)


app/controllers/create-service.controller.js
```diff
@@ -38,15 +38,19 @@ exports.post = (req, res) => {
...
-      })
+    return res.redirect(paths.serviceSwitcher.create)
+  }
+
+  try {
+    const service = await serviceService.createService(serviceName, serviceNameCy, req.user, correlationId)
+    await userService.assignServiceRole(req.user.externalId, service.externalId, 'admin', correlationId)
+    res.redirect(paths.serviceSwitcher.index)
+  } catch (err) {
+    renderErrorView(req, res, err)
```
sfount, line 49: I believe passing errors through to `next` will do the equivalent of a default render error view, that would be worth double checking though!
```suggestion
    next(err)
```

sfount (commented):
Refactor and additional log lines look good! Small question about error handling which might reduce dependencies

app/controllers/create-service.controller.js
stephencdaly: Yes it does, I almost did this but then was worried about having to update tests. I can do it

stephencdaly: Done

sfount (approved):
Great cleanup 🍬

PR #2388, opened 2020-12-01
PP-7416 Add button to enable or disable Worldpay 3DS Flex

alexbishop1:
Add a button to the ‘Your PSP’ page for Worldpay accounts to enable or disable 3DS Flex.

The enable button is shown only if it’s a Worldpay account, 3DS is enabled, 3DS Flex is disabled and there are 3DS Flex credentials.

image

The disable button is shown only if it’s a Worldpay account, 3DS is enabled and 3DS Flex is enabled.

image

No button is shown in other circumstances, meaning, for example, that a service cannot enable 3DS Flex if they have not entered 3DS Flex credentials.

image

The page that’s requested when the button is pressed requires the `toggle-3ds:update` permission whereas the page it’s displayed on requires the `gateway-credentials:read` permission. This is a less than ideal but only the `admin` role has either of these so the button won’t be shown to anyone who can’t press it.

Remove the ‘Remove credentials’ button from the 3DS Flex credentials page (which was the only way of turning off 3DS Flex before) because that’s no longer necessary (some code to support this function is still there; it will be removed in a future
commit).

### DIFF
--- a/app/controllers/your-psp/post-toggle-worldpay-3ds-flex.controller.js
+++ b/app/controllers/your-psp/post-toggle-worldpay-3ds-flex.controller.js
@@ -11,7 +11,7 @@ module.exports = async function toggleWorldpay3dsFlex (req, res) {
 
     if (req.body['toggle-worldpay-3ds-flex'] === 'on' || req.body['toggle-worldpay-3ds-flex'] === 'off') {
         const enabling3dsFlex = toggleWorldpay3dsFlex === 'on'
-        const message = enabling3dsFlex ? '3DS Flex turned on' : '3DS Flex turned off'
+        const message = enabling3dsFlex ? '3DS Flex has been turned on.' : '3DS Flex has been turned off. Your payments will now use 3DS only.'
         const integrationVersion3ds = enabling3dsFlex ? 2 : 1
         try {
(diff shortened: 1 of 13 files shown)


app/controllers/your-psp/post-toggle-worldpay-3ds-flex.controller.js
```diff
@@ -11,7 +11,7 @@ module.exports = async function toggleWorldpay3dsFlex (req, res) {
 
     if (req.body['toggle-worldpay-3ds-flex'] === 'on' || req.body['toggle-worldpay-3ds-flex'] === 'off') {
         const enabling3dsFlex = toggleWorldpay3dsFlex === 'on'
-        const message = enabling3dsFlex ? '3DS Flex turned on' : '3DS Flex turned off'
+        const message = enabling3dsFlex ? '3DS Flex has been turned on.' : '3DS Flex has been turned off. Your payments will now use 3DS only.'
```
oswaldquek, line 14: Where does this `message` get used in the UI?

alexbishop1: It appears in the green-bordered “flash” box at the top of the page when it loads again.

oswaldquek (approved):

PR #2389, opened 2020-12-02
PP-7416 Stop enabling 3DS when 3DS Flex credentials entered

alexbishop1:
Stop enabling 3DS when Worldpay 3DS Flex credentials are entered. We don’t enable 3DS Flex when credentials are entered any more (there’s now a button for that) so it doesn‘t make any sense to enable 3DS in general.

Also remove code that handles the logic for the ‘Remove credentials’ button because that button is no longer present now we have the button to enable or disable 3DS Flex.


nimalank7 (approved):
LGTM!

PR #2391, opened 2020-12-03
Pp 7360 create payment link add reporting columns

iqbalgds:
- Rename `edit-metatadata.njk` to `edit-reporting-columns.njk`
      - Update references in the `resource.controller.js`
- Take out `reporting columns` section from the 'edit-reporting-columns' nunjucks
      - Make it a separate include file.
    - Update `Add payment` flow review page
- Include new separate `edit-reporting-columns` nunjucks file - only show
      if the env variable PAYMENT_LINKS_INLINE_REPORTING_COLUMNS is present
- Update the Payment Link > Post Amount controller
      - Set the PAYMENT_LINKS_INLINE_REPORTING_COLUMNS flag in the session

image


app/controllers/payment-links/post-amount.controller.js
```diff
@@ -40,5 +40,9 @@ module.exports = function postAmount (req, res, next) {
     req.flash('generic', 'The details have been updated')
   }
 
+  if (process.env.PAYMENT_LINKS_INLINE_REPORTING_COLUMNS) {
```
kbottla, line 43: Can we use `process.env.PAYMENT_LINKS_INLINE_REPORTING_COLUMNS === 'true'`. Usually env value is treated as string and any value including 'false' will set `isPaymentLinksInlineReportingColumns` to true

kbottla (changes requested):

app/controllers/payment-links/post-amount.controller.js
iqbalgds: Good point - Just updated it.

In JS world, it would would ignore if it is 'false' - as its nots a truthy value. However, if someone put in `fals` then it would set it to true.  So better to make it explicit as per your recommendation.  Thanks

kbottla (approved):

app/controllers/payment-links/post-amount.controller.js
sfount: ➕ good catch and well updated, maybe we should put this in guidance.

Just FYI `process.env` doesn't parse any inputs from the system environment, everything is treated as a string. The string `"false"` is also a truthy value if coerced into a boolean. The only string that's evaluated as false is the empty string `""`.

PR #2397, opened 2020-12-08
PP-7491 Add error classes

stephencdaly:
Add custom error classes. These will be handled by a catch-all error handler at the end of the express stack which will render the appropriate error page.


sfount (approved):
🍬

PR #2398, opened 2020-12-08
Pp 7360 payment link reporting columns

iqbalgds:
PP-7360 Add Payment Links flow > Add Reporting Columns
    
    - Add new validation to `metadata-form.js`
      - Make sure metadata headers and values < 50
      - Make sure metadata headers are unique

    PP-7360 Update tunnel script - add products
    
    - Update script so that when you tunnel into test is also tunnels to 'products'
    - So that you can setup Payment Links in a local instance of Self service that tunnels to test.

### DIFF
--- a/app/controllers/payment-links/metadata/metadata-form.js
+++ b/app/controllers/payment-links/metadata/metadata-form.js
@@ -1,19 +1,43 @@
+const METADATA_MAX_HEADER_LENGTH = 30
+const METADATA_MAX_VALUE_LENGTH = 100
+const MAX_NO_OF_METADATA_COLUMNS = 10
+
 const fields = {
   metadataKey: {
     id: 'metadata-column-header',
     name: 'metadata-column-header',
-    validation: [ {
-      validator: isNotEmpty,
-      message: 'Enter a column header'
-    } ]
+    validation: [
+      {
+        validator: isNotEmpty,
+        message: 'Enter a column header'
+      },
+      {
+        validator: isValidLengthForColumnHeader,
+        message: `Column header must be ${METADATA_MAX_HEADER_LENGTH} characters or fewer`
+      },
+      {
+        validator: isNotDuplicate,
+        message: 'Column header must not already exist'
+      },
+      {
+        validator: isNotExceedingMaxNoOfReportingColumns,
+        message: `Number of reporting columns must be ${MAX_NO_OF_METADATA_COLUMNS} or fewer`
+      }
+    ]
   },
   metadataValue: {
     id: 'metadata-cell-value',
     name: 'metadata-cell-value',
-    validation: [ {
-      validator: isNotEmpty,
-      message: 'Enter cell content'
-    } ]
+    validation: [
+      {
+        validator: isNotEmpty,
+        message: 'Enter cell content'
+      },
+      {
+        validator: isValidLengthForCellContent,
+        message: `Cell content must be ${METADATA_MAX_VALUE_LENGTH} characters or fewer`
+      }
+    ]
   }
 }
 
@@ -31,11 +55,12 @@ const defaultFallbackError = {
 }
 
 class MetadataForm {
-  constructor (formData = {}) {
+  constructor (formData = {}, existingMetadata = {}) {
     this.values = {}
     this.fields = fields
     this.values[fields.metadataKey.id] = formData[fields.metadataKey.name]
     this.values[fields.metadataValue.id] = formData[fields.metadataValue.name]
+    this.existingMetadata = existingMetadata
   }
 
   // returns list of errors which may be on the form, both a map of ids to errors and a top level array
@@ -45,7 +70,7 @@ class MetadataForm {
     for (const fieldKey in fields) {
       const field = fields[fieldKey]
       field.validation.some((validationEntry) => {
-        const valid = validationEntry.validator(this.values[field.id])
+        const valid = validationEntry.validator(this.values[field.id], this.existingMetadata)
         if (!valid) {
           errors.push({
             href: `#${field.id}`,
@@ -78,6 +103,31 @@ class MetadataForm {
 
 function isNotEmpty (value) {
   return value && value.length !== 0
+}
+
+function isValidLengthForColumnHeader (value) {
+  return value && value.length !== 0 && value.length <= METADATA_MAX_HEADER_LENGTH
+}
+
+function isValidLengthForCellContent (value) {
+  return value && value.length !== 0 && value.length <= METADATA_MAX_VALUE_LENGTH
+}
+
+function isNotDuplicate (value, existingMetadata) {
+  if (existingMetadata) {
+    const found = Object.keys(existingMetadata).find(metadataKey => value === metadataKey)
+    return !found
+  }
+
+  return true
+}
+
+function isNotExceedingMaxNoOfReportingColumns (value, existingMetadata) {
+  if (Object.keys(existingMetadata) && Object.keys(existingMetadata).length >= MAX_NO_OF_METADATA_COLUMNS) {
+    return false
+  }
+
+  return true
 }
 
 module.exports = MetadataForm
--- a/test/unit/controller/payment-links/metadata/paymentlink-metadata-form.test.js
+++ b/test/unit/controller/payment-links/metadata/paymentlink-metadata-form.test.js
@@ -21,6 +21,64 @@ describe('Payment link metadata form model', () => {
     expect(tested.errorMaps['metadata-cell-value']).to.be.undefined // eslint-disable-line
   })
 
+  it('correctly validates given input that is too long', () => {
+    const body = { 
+      'metadata-column-header': 'aaaaaaaaaabbbbbbbbbbcccccccccc1',
+      'metadata-cell-value': 'aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeeeaaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeee2'
+    }
+
+    const form = new MetadataForm(body)
+    const tested = form.validate()
+
+    expect(tested.errors.length).to.eq(2)
+    expect(tested.errorMaps['metadata-column-header']).to.not.be.null // eslint-disable-line
+    expect(tested.errorMaps['metadata-cell-value']).to.not.be.null // eslint-disable-line
+  })
+
+  it('correctly validates a duplicate metadata column', () => {
+    const body = {
+      'metadata-column-header': 'test 2 header',
+      'metadata-cell-value': 'test 2 value'
+    }
+
+    const existingMetadata = {
+      'test 1 header': 'test 1 value',
+      'test 2 header': 'test 2 value'
+    }
+
+    const form = new MetadataForm(body, existingMetadata)
+    const tested = form.validate()
+    expect(tested.errors.length).to.eq(1)
+    expect(tested.errorMaps['metadata-column-header']).to.not.be.null // eslint-disable-line
+    expect(tested.errorMaps['metadata-cell-value']).to.be.undefined // eslint-disable-line
+  })
+
+  it('correctly validates when the number of metadata columns exceeds the max number of allowed metadata columns', () => {
+    const body = {
+      'metadata-column-header': 'test 11 header',
+      'metadata-cell-value': 'test 11 value'
+    }
+
+    const existingMetadata = {
+      'test 1 header': 'test 1 value',
+      'test 2 header': 'test 2 value',
+      'test 3 header': 'test 3 value',
+      'test 4 header': 'test 4 value',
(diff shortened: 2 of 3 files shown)


app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -45,7 +64,7 @@ class MetadataForm {
     for (const fieldKey in fields) {
       const field = fields[fieldKey]
       field.validation.some((validationEntry) => {
-        const valid = validationEntry.validator(this.values[field.id])
+        const valid = validationEntry.validator(this.values[field.id], this.existingMetadata)
```
iqbalgds, line 67: Could not get it working with out adding the extra param `existingMetadata `

app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -76,8 +95,17 @@ class MetadataForm {
   }
 }
 
-function isNotEmpty (value) {
+function isNotEmpty (value, existingMetadata) {
```
iqbalgds, line 98: Same as above

app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -36,6 +54,7 @@ class MetadataForm {
     this.fields = fields
     this.values[fields.metadataKey.id] = formData[fields.metadataKey.name]
     this.values[fields.metadataValue.id] = formData[fields.metadataValue.name]
+    this.existingMetadata = formData['existing-metadata']
```
sfount, line 57: I think I'd recommend passing this in as a separate (optional) parameter to the constructor. As existing metadata is not part of the actual form data this would keep the process clear

Something like
```js
class MetadataForm {
  constructor(formData = {}, existingMetadata = {}) {
    // ...
    this.existingMetadata = existingMetdata
  }
}
```

app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -76,8 +95,17 @@ class MetadataForm {
   }
 }
 
-function isNotEmpty (value) {
+function isNotEmpty (value, existingMetadata) {
   return value && value.length !== 0
 }
 
+function isValidLength (value, existingMetadata) {
```
sfount, line 102: If these methods don't actually require the `existingMetdata` attribute it's fine to leave them out

app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -2,21 +2,39 @@ const fields = {
...
-      message: 'Enter a column header'
-    } ]
+    validation: [
+      {
+        validator: isNotEmpty,
+        message: 'Enter a column header'
+      },
+      {
+        validator: isValidLength,
+        message: 'The column header is an invalid length'
```
sfount, line 12: These message should ideally follow the guidance in https://design-system.service.gov.uk/components/error-message/

Text input docs might help here https://design-system.service.gov.uk/components/text-input/

In this case we'd probably be going for 
>The column header must be 50 characters or less

app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -2,21 +2,39 @@ const fields = {
...
+      },
+      {
+        validator: isNotDuplicate,
+        message: 'The column header is an invalid length'
```
sfount, line 16: Looks like this message doesn't line up

app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -76,8 +95,17 @@ class MetadataForm {
...
+function isNotEmpty (value, existingMetadata) {
   return value && value.length !== 0
 }
 
+function isValidLength (value, existingMetadata) {
+  return value && value.length !== 0 && value.length <= META_DATA_MAX_LENGTH
+}
+
+function isNotDuplicate (value, existingMetadata) {
+  const found = existingMetadata.find(metadata => value === metadata['metadata-column-header'])
```
sfount, line 107: Ideal to use an variable here to avoid typing mistakes and make this resilient to the ID changing in the future 
```suggestion
  const found = existingMetadata.find(metadata => value === metadata[fields.metadataKey.id])
```

sfount (commented):
Looks like good progress! A few comments about the validation and messages

app/controllers/payment-links/metadata/metadata-form.js
iqbalgds: Thanks - I have updated it.

app/controllers/payment-links/metadata/metadata-form.js
iqbalgds: Done.

app/controllers/payment-links/metadata/metadata-form.js
iqbalgds: Done

app/controllers/payment-links/metadata/metadata-form.js
iqbalgds: Working now - removed the extra param.
Not sure why it was not working earlier.

app/controllers/payment-links/metadata/metadata-form.js
iqbalgds: Removed extra param, working now.

app/controllers/payment-links/metadata/metadata-form.js
iqbalgds: Done

app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -2,21 +2,39 @@ const fields = {
...
-      message: 'Enter a column header'
-    } ]
+    validation: [
+      {
+        validator: isNotEmpty,
+        message: 'Enter a column header'
+      },
+      {
+        validator: isValidLength,
+        message: 'Column header must be 50 characters or fewer'
```
sfount, line 12: It would be good for this to use the value in `META_DATA_MAX_LENGTH` to avoid duplication and make updating this in the future simpler

app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -80,4 +99,17 @@ function isNotEmpty (value) {
   return value && value.length !== 0
 }
 
+function isValidLength (value) {
+  return value && value.length !== 0 && value.length <= META_DATA_MAX_LENGTH
+}
+
+function isNotDuplicate (value, existingMetadata) {
+  if (existingMetadata && existingMetadata.length > 0) {
```
sfount, line 107: is `existingMetadata` expecting an array with a length? I think the format of metadata in general is an object of key value pairs, like
```js
{ key: value, anotherKey: value }
```

Is it possible to update this method to work with that? I don't know of anywhere where we have an array of objects with `key`, `value` params, potentially this could use something like
```js
const found = Object.keys(existingMetadata).find(metadataKey => value === metadataKey)
```

sfount (commented):
Comments well addressed! A few more minor things. 

I believe another constraint mentioned by the ticket is that there can be a maximum of 10 reporting columns for any given payment link, should we add that as a validation? It could make sure the `Object.keys(existingMetadata).length <= 10`

We'll probably remove the option to add a reporting column in the first place if the user already has 10 but it would be worth adding the validation as this is the stop gap before hitting the backend in a different part of the flow.

app/controllers/payment-links/metadata/metadata-form.js
iqbalgds: Good idea - done.

app/controllers/payment-links/metadata/metadata-form.js
iqbalgds: Done

test/unit/controller/payment-links/metadata/paymentlink-metadata-form.test.js
```diff
@@ -21,6 +21,64 @@ describe('Payment link metadata form model', () => {
     expect(tested.errorMaps['metadata-cell-value']).to.be.undefined // eslint-disable-line
   })
 
+  it('correctly validates given input that is an invalid length', () => {
```
iqbalgds, line 24: The tests before cover the case when there is no existing metadata.

app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -1,19 +1,42 @@
...
+      {
+        validator: isValidLength,
+        message: `Column header must be ${META_DATA_MAX_LENGTH} characters or fewer`
+      },
+      {
+        validator: isNotDuplicate,
+        message: 'Column header must not already exist'
+      },
+      {
+        validator: isNotExceedingMaxNoOfReportingColumns,
```
iqbalgds, line 22: As per feedback,  I have added an extra validator - to check that the number of reporting columns does not exceed 10.

app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -1,19 +1,42 @@
+const META_DATA_MAX_LENGTH = 50
+const MAX_NO_OF_METADATA_COLUMNS = 10
```
sfount, line 2: It would be good for these values to come from JS commons to have parity with the definitions in https://github.com/alphagov/pay-java-commons/blob/master/model/src/main/java/uk/gov/pay/commons/model/charge/ExternalMetadata.java. 

This could be done as a separate PR, I'd be happy to do this and it can be updated after, it shouldn't block this

iqbalgds:
Thanks for your feedback @sfount.
I have added some validation for checking if existing metadata exists 10 columns.

app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -1,19 +1,42 @@
+const META_DATA_MAX_LENGTH = 50
```
sfount, line 1: I've just double checked the ticket and the definitions for the backend and it looks like the key and the value have different specs 
```suggestion
const META_DATA_MAX_KEY_LENGTH = 30
const META_DATA_MAX_VALUE_LENGTH=100
```

These could then be passed into the validator when it's defined.

app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -1,19 +1,42 @@
...
-      message: 'Enter cell content'
-    } ]
+    validation: [
+      {
+        validator: isNotEmpty,
+        message: 'Enter cell content'
+      },
+      {
+        validator: isValidLength,
+        message: 'Cell content must be 50 characters or fewer'
```
sfount, line 37: This message should use the variable that's used for valid length

app/controllers/payment-links/metadata/metadata-form.js
iqbalgds: I have updated this PR with hardcoded values.

I will make an update to pay-js-commons to include these values and then update Self service.

app/controllers/payment-links/metadata/metadata-form.js
iqbalgds: Good spot - done.

app/controllers/payment-links/metadata/metadata-form.js
iqbalgds: Good spot - done.

app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -80,4 +105,29 @@ function isNotEmpty (value) {
   return value && value.length !== 0
 }
 
+function isValidLengthForColumnHeader (value) {
+  return value && value.length !== 0 && value.length <= METADATA_MAX_HEADER_LENGTH
+}
+
+function isValidLengthForCellContent (value) {
```
iqbalgds, line 112: Ended up splitting up the `isValidLength` method into:

`isValidLengthForColumnHeader ` & `isValidLengthForCellContent `

app/controllers/payment-links/metadata/metadata-form.js
```diff
@@ -80,4 +105,29 @@ function isNotEmpty (value) {
   return value && value.length !== 0
 }
 
+function isValidLengthForColumnHeader (value) {
+  return value && value.length !== 0 && value.length <= METADATA_MAX_HEADER_LENGTH
```
sfount, line 109: I don't think the length validators should be checking for empty in this case as the error message for empty should be different. If for some reason this validator ran first it would give the message 'Column header must be 50 characters or fewer' even though the thing it failed on was having length 0. 

In practice it looks like `isNotEmpty` will always run first so this shouldn't need to be blocking, we could update it as part of a future PR.

sfount (approved):
Looks great 👍  covers the scenarios documented in the ticket spec

app/controllers/payment-links/metadata/metadata-form.js
iqbalgds: Good point - I will create another PR for this.
Thanks

PR #2399, opened 2020-12-08
PP-7489 Add account and service middleware

kbottla:
## WHAT
- Added new middleware to get account and service based on URL path params

Middleware expects atleast one parameter (gatewayAccountExternalId or serviceExternalId) to be available on request to get gateway account/service. If both are available, additionally checks for gatewayAccount to be part of the service.

### DIFF
--- a/app/middleware/get-service-and-gateway-account.middleware.js
+++ b/app/middleware/get-service-and-gateway-account.middleware.js
@@ -1,0 +1,106 @@
+'use strict'
+
+const _ = require('lodash')
+
+const logger = require('../utils/logger')(__filename)
+const paths = require('../../app/paths')
+const { NotFoundError } = require('../../app/errors')
+const Connector = require('../services/clients/connector.client.js').ConnectorClient
+
+const { keys } = require('@govuk-pay/pay-js-commons').logging
+const connectorClient = new Connector(process.env.CONNECTOR_URL)
+
+async function getGatewayAccountByExternalId (gatewayAccountExternalId, correlationId) {
+  try {
+    const params = {
+      gatewayAccountExternalId: gatewayAccountExternalId,
+      correlationId: correlationId
+    }
+    // TODO: PP-7489 - add additional attributes derived by getAccount middleware
+    return await connectorClient.getAccountByExternalId(params)
+  } catch (err) {
+    const logContext = {}
+    logContext['error'] = err.message
+    logContext['GATEWAY_ACCOUNT_EXTERNAL_ID'] = gatewayAccountExternalId
+    logContext[keys.CORRELATION_ID] = correlationId
+
+    if (err.errorCode === 404) {
+      logger.info('Gateway account not found', logContext)
+      throw new NotFoundError('Gateway account not found')
+    } else {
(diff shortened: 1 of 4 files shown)


app/middleware/errors/not-found.error.js
```diff
@@ -0,0 +1,4 @@
+class NotFoundError extends Error {
```
kbottla, line 1: will rebase and replace with class from https://github.com/alphagov/pay-selfservice/pull/2397

app/middleware/get-service-and-gateway-account.middleware.js
```diff
@@ -0,0 +1,91 @@
...
+const logger = require('../utils/logger')(__filename)
+const paths = require('../../app/paths')
+const { NotFoundError } = require('../../app/errors')
+const Connector = require('../services/clients/connector.client.js').ConnectorClient
+
+const { keys } = require('@govuk-pay/pay-js-commons').logging
+const connectorClient = new Connector(process.env.CONNECTOR_URL)
+
+async function getGatewayAccountByExternalId (gatewayAccountExternalId, correlationId) {
+  if (!gatewayAccountExternalId) {
```
sfount, line 14: I believe it's an OK scenario if the gateway account hasn't been explicitly specified on the path, this might be the case if the route is at the service level (organisation settings for example).

I think this will throw a 500 internal server error if that's ever not defined on the path. As this middleware is just focussed on fetching ideally the only error it would throw is if it was asked to get something and that wasn't found (if there's no ID it could instead not try and get it).

sfount (commented):
Code looks great and follows recent guidance really well 👍  question about the control/ error flow through the middleware

app/middleware/get-service-and-gateway-account.middleware.js
kbottla: 👍 Agree, changed to not throw error when gateway account external ID is not available. Also made it mandatory for at least one external ID (gateway account or service) to be available on the request and should have `service` set on the request based on available external ID

app/middleware/get-service-and-gateway-account.middleware.js
```diff
@@ -0,0 +1,106 @@
...
+    return await connectorClient.getAccountByExternalId(params)
+  } catch (err) {
+    const logContext = {}
+    logContext['error'] = err.message
+    logContext['GATEWAY_ACCOUNT_EXTERNAL_ID'] = gatewayAccountExternalId
+    logContext[keys.CORRELATION_ID] = correlationId
+
+    if (err.errorCode === 404) {
+      logger.info('Gateway account not found', logContext)
+      throw new NotFoundError('Gateway account not found')
```
sfount, line 29: I think in some of these cases just logging at the error handling level would be sufficient. We would get the `error.message` logged for free and it also includes things like the correlation ID I believe. 

Streamlining logging here could be done as part of another PR.

sfount (approved):
Looks great 👍  much simpler to understand than some of the existing stuff and well covered. There may be one or two places we could simplify the flow through once we start using it but I think this is a great start

PR #2400, opened 2020-12-08
PP-7491 Handle custom errors in error handler middleware

stephencdaly:
- Rename `express-unhandled-error-handler.js` -> `error-handler.js` as it will now be used to deliberately handle errors thrown by Express stack.

- Add UserAccountDisabledError to handle the case thrown by the authentication middleware where the user is logged in, but their account has been disabled.

- Handle errors that will be thrown by newly added middleware in the same way as they are handled in existing middleware.

- In the case of the user account being disabled, we currently redirect to a `/login/noaccess` route. Instead just render the view as this route and controller don't do anything extra.


app/middleware/error-handler.js
```diff
@@ -0,0 +1,47 @@
...
+    return renderErrorView(req, res, 'You do not have the administrator rights to perform this operation.', 403)
+  }
+
+  if (err instanceof NotFoundError) {
+    res.status(404)
+    return res.render('404')
+  }
+
+  // trim the stack as long log messages are truncated by Splunk
+  logContext.stack = err.stack.substring(0, 300)
```
sfount, line 44: Is this still true? I know @heathd did some digging into why Splunk/ the logging pipeline was being truncated. This may no longer be the case

app/middleware/error-handler.js
```diff
@@ -0,0 +1,47 @@
...
+module.exports = function errorHandler (err, req, res, next) {
+  const logContext = {}
+  logContext[keys.CORRELATION_ID] = req.headers[CORRELATION_HEADER]
+
+  if (res.headersSent) {
+    return next(err)
+  }
+
+  if (err instanceof NotAuthenticatedError) {
+    req.session.last_url = req.originalUrl
```
sfount, line 24: Although 99.9/100 the session should be correctly configured for the request at the point of hitting this error handler it might be worth guarding against the possibility of it not, any code that could throw errors here would be handled unpredictably
```js
if (req.session) {
  req.session.last_url = req.originalUrl
}
```

sfount (approved):
Handler looks good 👍  nice simple test scenarios

app/middleware/error-handler.js
stephencdaly: good point 👍

app/middleware/error-handler.js
stephencdaly: Aha https://payments-platform.atlassian.net/browse/PP-7003
Nice, I'll remove this

sfount (approved):

PR #2401, opened 2020-12-08
PP-7489 Add adminusers and connector clients methods

kbottla:
## WHAT
Added client methods to get gateway accounts and service by external ID. These will be used by new getGatewayAccountAndService middleware (https://github.com/alphagov/pay-selfservice/pull/2399)
- Added `getAccountByExternalId` to connector client to retrieve gateway account by external ID
- Added `getServiceByExternalId` & `getServiceForGatewayAccount` to adminusers client to get service by external ID or for gateway account

### DIFF
--- a/app/services/clients/connector.client.js
+++ b/app/services/clients/connector.client.js
@@ -15,6 +15,7 @@ const StripeAccount = require('../../models/StripeAccount.class')
 const SERVICE_NAME = 'connector'
 const ACCOUNTS_API_PATH = '/v1/api/accounts'
 const ACCOUNT_API_PATH = ACCOUNTS_API_PATH + '/{accountId}'
+const ACCOUNT_API_BY_EXTERNAL_ID_PATH = ACCOUNTS_API_PATH + '/external-id/{externalId}'
 const CHARGES_API_PATH = ACCOUNT_API_PATH + '/charges'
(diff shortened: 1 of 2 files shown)


app/services/clients/connector.client.js
```diff
@@ -167,6 +173,34 @@ ConnectorClient.prototype = {
...
+        startTime: startTime,
+        correlationId: params.correlationId,
+        method: 'GET',
+        description: 'get an account by external id',
+        service: SERVICE_NAME
+      }
+
+      let callbackToPromiseConverter = createCallbackToPromiseConverter(context)
+
+      oldBaseClient.get(url, params, callbackToPromiseConverter, null)
```
sfount, line 200: If possible it would be ideal if we didn't add any new instances of using the old base client, I think Connector already has examples of methods that use the new base client? 

This would
* avoid adding new promise construction code
* allow upstream code to use async/ await patterns with no additional effort
* not require a connector client instance per request (for the pub/ sub error handling)

sfount (commented):
Looks good, comment about trying to use newer patterns

app/services/clients/connector.client.js
kbottla: There are few methods in connector client using new base client. Updated!

sfount (approved):

PR #2402, opened 2020-12-08
PP-7630 Create payment link reporting columns in-line flow

sfount:
Infrastructural changes required to bring reporting columns into the payment link flow.

This PR
* adds controllers to locally (with a cookie session) add, remove and update metadata against the current payment link being created
* introduces simple helpers to manage metadata locally
* allows editing of the reporting column key if you're using the new flow, this is now possible as the results will not yet have been committed against a database

This code will not yet be accessible as it's behind the flag introduced by the previous PR against this ticket.


iqbalgds:
Really good work 👍

iqbalgds (approved):

PR #2403, opened 2020-12-10
PP-7488 Ensure authentication middleware is included on all routes that require authorisation

stephencdaly:
## PP-7488 Include MOTO masking settings page in authenticated paths

Not all authentication middleware was run on these routes. This means we weren't checking that the user has a valid session and we weren't applying CSRF checks to the POST route.

##     PP-7488 Ensure all "My profile" routes run authorisation middleware

Group "/my-profile/" paths in object within paths object.

Previously, "/my-profile/phone-number" wasn't added to the authenticated paths array, so not all authorisation middleware was included on this route. This means we weren't checking that the user has a valid session and we weren't applying CSRF checks to the POST route.

Add `paths.profile` to the authenticated routes array so all paths under it have the authorisation middleware and CSRF middleware applied.

### DIFF
--- a/app/controllers/service-users.controller.js
+++ b/app/controllers/service-users.controller.js
@@ -22,7 +22,7 @@ const mapByRoles = function (users, externalServiceId, currentUser) {
       }
       if (currentUser.externalId === user.externalId) {
         mappedUser.is_current = true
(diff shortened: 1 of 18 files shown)


kbottla (approved):
LGTM!

PR #2404, opened 2020-12-10
PP-7488 Add unit test to ensure authorisation middleware used

stephencdaly:
Add test to ensure that authorisation middleware is called for all routes that require authentication.

Achieve this by specifying in the test which paths do not require authentication. Then check all route setup calls to ensure that the authorisation middleware is used for routes with paths that are not in this list.

This will ensure that the middleware is used for any new routes that are registered, unless they are explicitly added to the exclude list.

When the new authentication middleware has been added, will modify to check if the new middleware is run instead. And can be updated to remove checks for the old middleware when we have fully replaced it.

### DIFF
--- a/test/unit/routes.test.js
+++ b/test/unit/routes.test.js
@@ -1,0 +1,85 @@
+'use strict'
+
+const sinon = require('sinon')
+
+const paths = require('../../app/paths')
+const routes = require('../../app/routes')
+const oldAuthorisationMiddleware = require('../../app/services/auth.service').enforceUserAuthenticated
+
+const pathsNotRequiringAuthentication = [
+  '/style-guide',
+  paths.user.logIn,
+  paths.user.otpLogIn,
+  paths.user.otpSendAgain,
+  paths.user.logOut,
+  paths.user.noAccess,
+  paths.user.forgottenPassword,
+  paths.user.passwordRequested,
+  paths.user.forgottenPasswordReset,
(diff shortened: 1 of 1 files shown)


test/unit/routes.test.js
```diff
@@ -0,0 +1,85 @@
...
+    // We currently call `app.use` to specify a list of paths to use certain middleware for, 
+    // including the authorisation middleware. We need to check that paths are either included in
+    // the array for this call, or the middleware is added to the stack individually for the route
+    const authenticatedPathsArg = app.use.getCalls()
+      .find(call => {
+        return Array.isArray(call.args[0]) &&
+          call.args.includes(oldAuthorisationMiddleware)
+      })
+      .args[0]
+    const authenticatedPaths = flattenPaths(authenticatedPathsArg)
```
sfount, line 64: It took me a minute to fully get what was going on with this step before getting to the routes! I think this edge case coverage could also be removed if we're encouraging route based authorisation with the new auth middleware

sfount (approved):
Clear, concise and covers a weak point in app authorisation that has already caught paths that were misconfigured.

This is an interesting and well executed idea which I think would improve the confidence in security in all of the public facing apps 👍

test/unit/routes.test.js
stephencdaly: Yep, it can be removed down the line, maybe we should put a reminder in to do it on another story?

PR #2406, opened 2020-12-10
PP-7489 Extend gateway account object

kbottla:
## WHAT
- Extends gateway account object (get account and service middleware) with additional attributes similar to get-gateway-account.js middleware


sfount (approved):
LGTM 🍬

PR #2408, opened 2020-12-11
PP-7488 Add authorisation middleware

stephencdaly:
Add middleware that will replace existing authorisation middleware and be used alongside the new middleware to resolve the service and gateway account from the URL path params.

This middleware optionally takes a permission string and checks:
- That the user is logged in and has a valid session
- The the middleware to resolve the service from the URL path param has successfully resolved a service if there is a service/gateway account id path param
- That the user has a service role for the resolved service
- If a permission has been specified, that the user has the permission for the resolved service

## Questions

- Existing auth middleware calls `csrf.ensureSessionHasCsrfSecret(req, res, next)`. I haven't added anything to do this on the assumption that we'll change how CSRF is implemented as part of our change, is that correct?
- If there is a User on the session, but they have not completed second factor authentication, currently we redirect them to the OTP entry page. This PR instead directed them to the login page by passing a NotAuthenticatedError to `next`. Should we maintain existing behaviour and add another error type to allow for this, or does this seem ok?

### DIFF
--- a/app/middleware/user-is-authorised.js
+++ b/app/middleware/user-is-authorised.js
@@ -1,0 +1,40 @@
+'use strict'
+
+const { NotAuthorisedError, NotAuthenticatedError, UserAccountDisabledError } = require('../errors')
+const { SERVICE_EXTERNAL_ID, GATEWAY_ACCOUNT_EXTERNAL_ID } = require('../paths').keys
+
+module.exports = function getUserIsAuthorisedMiddleware (permission) {
+  return function userIsAuthorised (req, res, next) {
+    const { user, session, params, service } = req
+
+    if (!user) {
+      return next(new NotAuthenticatedError('User not found on request'))
+    }
+    if (!session) {
+      return next(new NotAuthenticatedError('Session not found on request'))
+    }
+    if (user.sessionVersion !== session.version) {
+      return next(new NotAuthenticatedError(`Invalid session version - session version is ${session.version}, user session version is ${user.sessionVersion}`))
+    }
+    if (user.disabled) {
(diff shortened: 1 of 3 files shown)


sfount (commented):
Haven't looked through all of the code yet but it looks sensible to me, having this all in one place is night and day. 

Thinking through the questions: 

1.  Correct no CSRF here, CSRF will hopefully be taken care of by `PP-7621`
2.  I believe the decision about where to redirect on auth failure is made based on the users cookie session. As that's available to the app error handler at the end of the stack what do you feel about continuing to throw a uniform `NotAuthenticatedError` but than have the thing that will redirect do so based on the progress they've made through auth so far (I think it's something like if there exists the flag for 2fa or not?)

app/middleware/user-is-authorised.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+    if (user.disabled) {
+      return next(new UserAccountDisabledError('User account is disabled'))
+    }
+    if (!(session.secondFactor && session.secondFactor === 'totp')) {
+      return next(new NotAuthenticatedError('Not completed second factor authentication'))
+    }
+
+    if (params[GATEWAY_ACCOUNT_EXTERNAL_ID] || params[SERVICE_EXTERNAL_ID]) {
+      if (!service) {
```
sfount, line 27: I wonder if it would be good to add a sense check here that if there's a gateway account ID it belongs to the service that has been placed on the session. 

This currently relies on the request data middleware all behaving correctly, I can't think of a scenario where a session could be added that the gateway account doesn't belong to but it would be a fairly simple check and potentially help firm up edge cases.

sfount (approved):

PR #2409, opened 2020-12-11
PP-7482 Payment links rename manage payment link index

sfount:
As the root page for managing payment links is now nested in the
`paymentLinks.manage` attribute, rename the key to `index`.


SandorArpa (approved):
LOKTM

PR #2415, opened 2020-12-14
PP-7422: Add a function to call to connector to check Worldpay 3DS Flex credentials

adityapahuja:
This change

- adds a new function `postCheckWorldpay3dsFlexCredentials` which makes a call to a connector application to check Worldpay 3DS Flex credentials.
- adds a PACT test to test the function to ensure that it can handle a happy scenario correctly.
- fixes authenticate PACT tests failures.

### DIFF
--- a/app/services/clients/connector.client.js
+++ b/app/services/clients/connector.client.js
@@ -30,6 +30,7 @@ const ACCEPTED_CARD_TYPES_FRONTEND_PATH = ACCOUNT_FRONTEND_PATH + '/card-types'
 const ACCOUNT_NOTIFICATION_CREDENTIALS_PATH = '/v1/api/accounts' + '/{accountId}' + '/notification-credentials'
 const ACCOUNT_CREDENTIALS_PATH = ACCOUNT_FRONTEND_PATH + '/credentials'
 const EMAIL_NOTIFICATION__PATH = '/v1/api/accounts/{accountId}/email-notification'
+const CHECK_WORLDPAY_3DS_FLEX_CREDENTIALS_PATH = '/v1/api/accounts/{accountId}/worldpay/check-3ds-flex-config'
 const FLEX_CREDENTIALS_PATH = '/v1/api/accounts/{accountId}/3ds-flex-credentials'
 const TOGGLE_3DS_PATH = ACCOUNTS_FRONTEND_PATH + '/{accountId}/3ds-toggle'
 
@@ -301,11 +302,31 @@ ConnectorClient.prototype = {
   },
 
   /**
- *
- * @param {Object} params
- * @param {Function} successCallback
- * @returns {Promise}
- */
+   * Checks Worldpay 3DS Flex credentials
+   *
+   * @param {Object} params
+   * @returns {Promise<Object>}
+   */
+  postCheckWorldpay3dsFlexCredentials: function (params) {
+    return baseClient.post(
+      {
+        baseUrl: this.connectorUrl,
+        url: CHECK_WORLDPAY_3DS_FLEX_CREDENTIALS_PATH.replace('{accountId}', params.gatewayAccountId),
+        json: true,
+        body: params.payload,
+        correlationId: params.correlationId,
+        description: 'Check Worldpay 3DS Flex credentials',
+        service: SERVICE_NAME
+      }
+    )
+  },
+
+  /**
+   *
+   * @param {Object} params
+   * @param {Function} successCallback
+   * @returns {Promise}
+   */
   post3dsFlexAccountCredentials: function (params) {
     return new Promise((resolve, reject) => {
       const url = _get3dsFlexCredentialsUrlFor(params.gatewayAccountId, this.connectorUrl)
@@ -695,7 +716,7 @@ ConnectorClient.prototype = {
     })
   },
 
-    /**
+  /**
    * @param gatewayAccountId
    * @param integrationVersion3ds (number)
    * @param correlationId
--- a/test/fixtures/worldpay-3ds-flex-credentials.fixtures.js
+++ b/test/fixtures/worldpay-3ds-flex-credentials.fixtures.js
@@ -1,0 +1,48 @@
+'use strict'
+
+const path = require('path')
+const _ = require('lodash')
+
+const pactBase = require(path.join(__dirname, '/pact-base'))
+const pactRegister = pactBase()
+
+const checkValidWorldpay3dsFlexCredentialsRequest = function checkValidWorldpay3dsFlexCredentialsRequest (opts = {}) {
+  const data = {
+    correlationId: opts.correlationId || 'a1',
+    gatewayAccountId: opts.gatewayAccountId || 333,
+    payload: {
+      organisational_unit_id: opts.organisationalUnitId || '5bd9b55e4444761ac0af1c80',
+      issuer: opts.issuer || '5bd9e0e4444dce153428c940',
+      jwt_mac_key: opts.jwtMacKey || 'fa2daee2-1fbb-45ff-4444-52805d5cd9e0'
+    }
+  }
+
+  return {
+    getPactified: () => {
+      return pactRegister.pactify(data)
+    },
+    getPlain: () => {
+      return _.clone(data)
+    }
+  }
+}
+
+const checkValidWorldpay3dsFlexCredentialsResponse = function checkValidWorldpay3dsFlexCredentialsResponse (opts = {}) {
+  const data = {
+    result: opts.result || 'valid'
+  }
+
+  return {
+    getPactified: () => {
+      return pactRegister.pactify(data)
+    },
+    getPlain: () => {
+      return _.clone(data)
+    }
+  }
+}
+
+module.exports = {
+  checkValidWorldpay3dsFlexCredentialsRequest,
+  checkValidWorldpay3dsFlexCredentialsResponse
+}
--- a/test/unit/clients/connector-client/connector-post-worldpay-check-3ds-flex-config.pact.test.js
+++ b/test/unit/clients/connector-client/connector-post-worldpay-check-3ds-flex-config.pact.test.js
@@ -1,0 +1,67 @@
+'use strict'
+
+const path = require('path')
+const chai = require('chai')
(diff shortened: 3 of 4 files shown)


app/services/clients/connector.client.js
```diff
@@ -301,11 +307,40 @@ ConnectorClient.prototype = {
...
+        url: url,
+        defer: { resolve: resolve, reject: reject },
+        startTime: startTime,
+        correlationId: params.correlationId,
+        body: params.payload
+      }
+      requestLogger.logRequestStart(context)
+
+      const callbackToPromiseConverter = createCallbackToPromiseConverter(context)
+      oldBaseClient.post(url, params, callbackToPromiseConverter)
```
sfount, line 333: 👋  this review is just for JavaScript rather than functionality, it would be ideal if this could use the new "baseClient" as we're trying to incrementally phase out this older pattern. We're doing that because: 
* creating promises ourselves higher in the stack and passing `resolve`, `reject` around can lead to tight coupling and hard to follow code
* the pub/ sub method of async used for error handling requires a new instance of the client on every request which has caused issues in the past

The interface for the new base client is very similar! There are plenty of examples in the `connector.client.js` file, it might look something like: 
```js 
postCheckWorldpay3dsFlexCredentials: function postCheckWorldpay3dsFlexCredentials(params) {
  return baseClient.post({
    baseUrl: this.connectorUrl,
    url: CHECK_WORLDPAY_3DS_FLEX_CREDENTIALS_PATH.replace('{accountId}', params.gatewayAccountId),
    // ...
  })
}
```

The current client is already promise based underneath so just returning that will accomplish the same thing. 

Please let me know if this isn't clear or if I can help out making this change!

adityapahuja: Thanks @sfount. I will make the changes.

test/unit/clients/connector-client/connector-post-worldpay-check-3ds-flex-config.pact.test.js
```diff
@@ -0,0 +1,127 @@
...
+    describe('and the credentials are valid and pass the existing format validation', () => {
+      const checkValidWorldpay3dsFlexCredentialsRequest = worldpay3dsFlexCredentialsFixtures.checkValidWorldpay3dsFlexCredentialsRequest()
+      const checkValidWorldpay3dsFlexCredentialsResponse = worldpay3dsFlexCredentialsFixtures.checkValidWorldpay3dsFlexCredentialsResponse()
+      before(done => {
+        const pactifiedRequest = checkValidWorldpay3dsFlexCredentialsRequest.getPactified()
+        const pactifiedResponse = checkValidWorldpay3dsFlexCredentialsResponse.getPactified()
+
+        provider.addInteraction(
+          new PactInteractionBuilder(`${ACCOUNTS_RESOURCE}/${EXISTING_GATEWAY_ACCOUNT_ID}/${CHECK_WORLDPAY_3DS_FLEX_CREDENTIALS}`)
+            .withState('has the credentials')
```
oswaldquek, line 45: This state should be more specific. I suggest `a gateway account with Worldpay 3DS Flex credentials exists`. Then, in pay-connector's [ContractTest](https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/pact/ContractTest.java#L72-L73), there should be a method like 
```
    @State("a gateway account with Worldpay 3DS Flex credentials exists")
    public void aGatewayAccountWithWorldpay3dsFlexCredentialsExists(Map<String, String> params) {
        // add gatweay account with worldpay 3ds flex creds to the database
    }
```
Then, a pull request should be raised against pay-connector first before this selfservice PR can pass.

test/unit/clients/connector-client/connector-post-worldpay-check-3ds-flex-config.pact.test.js
```diff
@@ -0,0 +1,127 @@
...
+        const valid3dsFlexCredentials = checkValidWorldpay3dsFlexCredentialsRequest.getPlain()
+        const validResult = checkValidWorldpay3dsFlexCredentialsResponse.getPlain()
+        connectorClient.postCheckWorldpay3dsFlexCredentials(valid3dsFlexCredentials)
+          .should.be.fulfilled.then((response) => {
+            expect(response).to.deep.equal(validResult)
+          }).should.notify(done)
+      })
+    })
+
+    describe('and the credentials are invalid and pass the existing format validation', () => {
```
oswaldquek, line 68: Generally speaking, pact tests are expensive and we don't want to unnecessarily cater for every scenario using a pact test. In the context of this story, the purpose of the pact test is to ensure that `/v1/api/accounts/{accountId}/worldpay/check-3ds-flex-config` does return 
```
{
  "result": "valid"
}
```
and we want to ensure this contract will never be broken.

Tests like this (which check for valid credentials but don't pass the format validation) should be unit/integration tests in pay-connector itself, so this test case can be removed.

test/unit/clients/connector-client/connector-post-worldpay-check-3ds-flex-config.pact.test.js
```diff
@@ -0,0 +1,127 @@
...
+        const invalid3dsFlexCredentials = checkInvalidWorldpay3dsFlexCredentialsRequest.getPlain()
+        const invalidResult = checkInvalidWorldpay3dsFlexCredentialsResponse.getPlain()
+        connectorClient.postCheckWorldpay3dsFlexCredentials(invalid3dsFlexCredentials)
+          .should.be.fulfilled.then((response) => {
+            expect(response).to.deep.equal(invalidResult)
+          }).should.notify(done)
+      })
+    })
+
+    describe('and the credentials are valid and pass the existing format validation but credentials check fails', () => {
```
oswaldquek, line 100: As per the reason above, this test should be removed.

oswaldquek (changes requested):
Good effort as it's your first pact test! Some tests can be removed however.

test/unit/clients/connector-client/connector-post-worldpay-check-3ds-flex-config.pact.test.js
adityapahuja: Thanks @oswaldquek. I will make the changes.

test/unit/clients/connector-client/connector-post-worldpay-check-3ds-flex-config.pact.test.js
adityapahuja: Done.

test/unit/clients/connector-client/connector-post-worldpay-check-3ds-flex-config.pact.test.js
adityapahuja: Thanks for the explanation. I have removed unhappy scenarios.

test/unit/clients/connector-client/connector-post-worldpay-check-3ds-flex-config.pact.test.js
adityapahuja: Done. Please see the [pull request](https://github.com/alphagov/pay-connector/pull/2754) on pay-connector.

app/services/clients/connector.client.js
adityapahuja: I have updated to use `baseClient.post()`.

test/fixtures/worldpay-3ds-flex-credentials.fixtures.js
```diff
@@ -0,0 +1,48 @@
...
+
+const pactBase = require(path.join(__dirname, '/pact-base'))
+const pactRegister = pactBase()
+
+const checkValidWorldpay3dsFlexCredentialsRequest = function checkValidWorldpay3dsFlexCredentialsRequest (opts = {}) {
+  const data = {
+    correlationId: opts.correlationId || 'a1',
+    gatewayAccountId: opts.gatewayAccountId || 333,
+    payload: {
+      organisationalUnitId: opts.organisationalUnitId || '5bd9b55e4444761ac0af1c80',
```
oswaldquek, line 14: This possibly needs to be `organisation_unit_id`

test/fixtures/worldpay-3ds-flex-credentials.fixtures.js
```diff
@@ -0,0 +1,48 @@
...
+      issuer: opts.issuer || '5bd9e0e4444dce153428c940',
+      jwtMacKey: opts.jwtMacKey || 'fa2daee2-1fbb-45ff-4444-52805d5cd9e0'
```
oswaldquek, line 16: This possibly needs to be `jwt_mac_key`

test/fixtures/worldpay-3ds-flex-credentials.fixtures.js
adityapahuja: Yes, PACT failed due to the following error. Will fix it.

```
BodyComparisonResult(mismatches={$=[BodyMismatch(expected={"result":"valid"}, actual={"error_identifier":"GENERIC","message":["Unrecognized field \"organisationalUnitId\" (class uk.gov.pay.connector.gatewayaccount.model.Worldpay3dsFlexCredentialsRequest), not marked as ignorable (3 known properties: \"issuer\", \"organisational_unit_id\", \"jwt_mac_key\"])\n
```

adityapahuja: Thanks. I have replaced `organisationalUnitId` with `organisation_unit_id`.

test/fixtures/worldpay-3ds-flex-credentials.fixtures.js
adityapahuja: Thanks. I have replaced `jwtMacKey` with `jwt_mac_key`.

adityapahuja:
I have replaced the following fields in a payload used to sent to Connector application.

1. correlationId replaced with correlation_id.
2. gatewayAccountId replaced with gateway_account_id.

adityapahuja:
I am sticking with `correlationId` and `gatewayAccountId`. They are not used in a payload.

oswaldquek (approved):
Well done!

PR #2417, opened 2020-12-14
PP-7489 Get request data middleware normalise gateway account as string

sfount:
Connector returns `gateway_account_id` as a `Number` value, AdminUsers
service `gateway_account_ids` as `String`s.

Update the test to realistically reflect this and fail. Update the
middleware to cast number to string so they can be compared with
`Array.includes`.


kbottla (approved):
LGTM!

PR #2419, opened 2020-12-15
PP-7509 Pass session metadata through to products calls

sfount:
The create and manage flow both now curate and validate metadata in the
payment link session, allowing the user to safely create, update and
delete metadata.

The backend now supports receiving a complete metadata object, pass this
through to the API calls.


kbottla (approved):
LGTM!

PR #2421, opened 2020-12-16
PP-7509 Log salient reporting column actions

sfount:
Updating reporting columns should catch all errors that could happen
during controller execution.

Add logs with salient information requested in the ticket for tracking
volumes of in-session changes.

Fix duplicate check to make it in-line with the products and public api
backend.


SandorArpa (approved):
LOKTM

kbottla (approved):
LGTM!

PR #2422, opened 2020-12-16
PP-7488 Remove permission check from new authorisation middleware

stephencdaly:
We've decided that our new pattern for setting up routes will be to create various routers for the different route paths. Something like the following:
  - Service router - sets up all routes for pages related to a specific service
  - Account router - sets up all routes for pages related to a specific account
  - Some other router(s) for other logged in routes such as user profile settings and all services views

We can then initialise the authorisation middleware (user-is-authorised.js) on the router level rather than on each individual route as this middleware must be run on every logged in route.

The permissions middleware requires initialisation with a specific permission string which will depend on the route, so must be included on individual routes and therefore needs to remain separate.

Remove permissions checks from the new authorisation middleware and update existing permissions middleware to throw a new `PermissionDeniedError` which is handled by the error handler middleware to follow our new pattern. This will not change any external behaviour.

### DIFF
--- a/app/errors.js
+++ b/app/errors.js
@@ -26,6 +26,12 @@ class UserAccountDisabledError extends DomainError {
 class NotAuthorisedError extends DomainError {
 }
 
+class PermissionDeniedError extends DomainError {
+  constructor (permission) {
+    super(`User does not have permission ${permission} for service`)
+  }
+}
+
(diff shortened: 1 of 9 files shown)


kbottla (approved):
LGTM!

PR #2424, opened 2020-12-17
PP-7489 Remove logging from gatewayAccountAndService middleware

kbottla:
## WHAT
- Removed most of the logging as error handler middleware is logging the details now for errors thrown and adds additional context.
- getGatewayaccountAndService middleware doesn't throw exceptions anymore if service or gateway account are not found. Authorisation middleware will check for service/gateway account and user access and throws relevant error
- Refactored tests following story review comments to make sure assertions are always reached

### DIFF
--- a/app/middleware/get-service-and-gateway-account.middleware.js
+++ b/app/middleware/get-service-and-gateway-account.middleware.js
@@ -3,8 +3,7 @@
 const _ = require('lodash')
 
 const logger = require('../utils/logger')(__filename)
-const paths = require('../../app/paths')
-const { NotFoundError } = require('../../app/errors')
+const { SERVICE_EXTERNAL_ID, GATEWAY_ACCOUNT_EXTERNAL_ID } = require('../paths').keys
 const Connector = require('../services/clients/connector.client.js').ConnectorClient
 
 const { keys } = require('@govuk-pay/pay-js-commons').logging
@@ -34,15 +33,14 @@ async function getGatewayAccountByExternalId (gatewayAccountExternalId, correlat
   } catch (err) {
     const logContext = {}
     logContext['error'] = err.message
+    logContext['error_code'] = err.errorCode
     logContext['GATEWAY_ACCOUNT_EXTERNAL_ID'] = gatewayAccountExternalId
     logContext[keys.CORRELATION_ID] = correlationId
 
     if (err.errorCode === 404) {
       logger.info('Gateway account not found', logContext)
-      throw new NotFoundError('Gateway account not found')
     } else {
-      logger.error('Error when attempting to retrieve gateway account', logContext)
-      throw new Error('Error retrieving Gateway account')
+      logger.error('Error retrieving gateway account', logContext)
     }
   }
 }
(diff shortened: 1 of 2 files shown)


iqbalgds (changes requested):

app/middleware/get-service-and-gateway-account.middleware.js
```diff
@@ -71,47 +65,35 @@ function getService (user, serviceExternalId, gatewayAccount, correlationId) {
     // Since it is only card currently supported, value is always 'true'
     service.hasCardGatewayAccount = true
     return service
```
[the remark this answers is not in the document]
kbottla: It is logged by authorised middleware if service is not set https://github.com/alphagov/pay-selfservice/blob/9630ff79bad4fa45f1aee2cd82c3b6550beca0fa/app/middleware/user-is-authorised.js#L28

app/middleware/get-service-and-gateway-account.middleware.js
```diff
@@ -34,16 +33,11 @@ async function getGatewayAccountByExternalId (gatewayAccountExternalId, correlat
   } catch (err) {
     const logContext = {}
     logContext['error'] = err.message
+    logContext['error_code'] = err.errorCode
     logContext['GATEWAY_ACCOUNT_EXTERNAL_ID'] = gatewayAccountExternalId
     logContext[keys.CORRELATION_ID] = correlationId
 
-    if (err.errorCode === 404) {
```
stephencdaly, line 40: If it's a 404, do we not want to return an error? We probably want to continue and show the not authorised page if we're not explicitly returning a not found error anymore. So calling `next` without arguments is probably needed for the case of a 404.

app/middleware/get-service-and-gateway-account.middleware.js
```diff
@@ -34,16 +33,11 @@ async function getGatewayAccountByExternalId (gatewayAccountExternalId, correlat
   } catch (err) {
     const logContext = {}
     logContext['error'] = err.message
+    logContext['error_code'] = err.errorCode
     logContext['GATEWAY_ACCOUNT_EXTERNAL_ID'] = gatewayAccountExternalId
     logContext[keys.CORRELATION_ID] = correlationId
```
stephencdaly, line 38: Can probably skip all this logging for an unexpected error. The error handler will log enough and we can trace together exactly what went wrong by following through and looking at the connector logs

app/middleware/get-service-and-gateway-account.middleware.js
stephencdaly: I misread, we aren't throwing an error in this case. Although we probably don't want to log at error level for a 404.

app/middleware/get-service-and-gateway-account.middleware.js
stephencdaly: Can ignore, it would make sense to log something here probably. Just need to make sure it's only logging at error level if it's a problem with the system that we need to address

kbottla: Updated logging. Logs at error level only if there is a problem and info level when not found.

app/middleware/get-service-and-gateway-account.middleware.js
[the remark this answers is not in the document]
iqbalgds: Thanks - sounds good

app/middleware/get-service-and-gateway-account.middleware.js
```diff
@@ -71,47 +69,35 @@ function getService (user, serviceExternalId, gatewayAccount, correlationId) {
...
-    const serviceExternalId = req.params[paths.keys.SERVICE_EXTERNAL_ID]
-    const gatewayAccountExternalId = req.params[paths.keys.GATEWAY_ACCOUNT_EXTERNAL_ID]
-
-    if (!serviceExternalId && !gatewayAccountExternalId) {
-      const logContext = {}
-      logContext[keys.CORRELATION_ID] = correlationId
-      logger.info('Could not resolve gateway account external ID or service external ID from request params')
-      throw new Error('Could not resolve gateway account external ID or service external ID from request params')
-    }
+    if (req.user) {
```
stephencdaly, line 77: could consider reversing this condition to reduce indentation but that's a style choice

stephencdaly (approved):
Looks good to me

iqbalgds (approved):

PR #2425, opened 2020-12-17
PP-7488 Skip resolve service middleware if service already resolved

stephencdaly:
Skip running this middleware if the service has already been resolved. This is because the `permissions.js` middleware also returns this middleware. We are changing to a different middleware to resolve the service as part of https://payments-platform.atlassian.net/browse/PP-7520 so this is to enable that switchover while it is in progress. We can remove the resolve service middleware when this is done.

Bundled in a commit to run the linter. It's only `resolve-service.js` that has code changes.


SandorArpa (approved):
LOKTM

PR #2426, opened 2020-12-17
PP-7520 Establish account url structure

sfount:
Update the dashboard route to use the new account URL structure. 

This proposes a method of adding an `account` property to the paths, all paths in this property would be added to an account sub router which would make sure the URL structure is consistent. 

This would also allow us to "upgrade" requests from the previous structure during the transition by checking for paths that are in the `account` property and redirecting to the new URL if possible.


sfount:
Superseded by https://github.com/alphagov/pay-selfservice/pull/2429

PR #2428, opened 2020-12-18
PP-7422: Add tests on checking Worldpay 3DS Flex credentials

adityapahuja:
## WHAT

This change adds two tests to Cypress to test the following scenarios.

1. Connector returns result = invalid after checking Worldpay 3DS Flex credentials.
2. Connector fails while checking Worldpay 3DS Flex credentials.

## HOW 

I ran both mocha tests and cypress tests. They passed. I verified responses from stubs and they were valid.

### DIFF
--- a/test/cypress/integration/settings/your-psp.cy.test.js
+++ b/test/cypress/integration/settings/your-psp.cy.test.js
@@ -22,6 +22,16 @@ describe('Your PSP settings page', () => {
     issuer: '5bd9e0e4444dce153428c940',
     jwt_mac_key: 'fa2daee2-1fbb-45ff-4444-52805d5cd9e0'
   }
+  const testInvalidFlexCredentials = {
+    organisational_unit_id: '5bd9b55e4444761ac0af1c81',
+    issuer: '5bd9e0e4444dce153428c941',
+    jwt_mac_key: 'ffffffff-aaaa-1111-1111-52805d5cd9e1'
+  }
+  const testFailureFlexCredentials = {
+    organisational_unit_id: '5bd9b55e4444761ac0af1c82',
+    issuer: '5bd9e0e4444dce153428c942',
+    jwt_mac_key: 'ffffffff-ffff-ffff-ffff-ffffffffffff'
+  }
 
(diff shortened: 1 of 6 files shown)


test/cypress/integration/settings/your-psp.cy.test.js
```diff
@@ -133,6 +159,46 @@ describe('Your PSP settings page', () => {
         expect(location.pathname).to.eq(`/your-psp`)
       })
     })
+
+    it('should not allow invalid 3DS Flex credentials to be saved', () => {
```
oswaldquek, line 163: This looks like it's re-testing https://github.com/alphagov/pay-selfservice/blob/master/test/cypress/integration/settings/your-psp.cy.test.js#L118-L127 but in a more thorough way. I wonder if we can delete https://github.com/alphagov/pay-selfservice/blob/master/test/cypress/integration/settings/your-psp.cy.test.js#L118-L127?

oswaldquek: So re-write https://github.com/alphagov/pay-selfservice/blob/master/test/cypress/integration/settings/your-psp.cy.test.js#L117 to be just a positive test.

adityapahuja: L118-L127 shows a user is entering the following details and submitted them and tests them to ensure that these values fail during validation process in self-service application before going to Connector for checking Worldpay 3DS Flex credentials and displays the error message.

```
Organisational unit ID: Invalid organisational unit ID
Issuer: Invalid issuer
JWT MAC key: Invalid JWT MAC key
```

If we remove these lines, we are going to skip testing `validate3dsFlexCredentials()` function and leading and trailing space removals. Are we okay to remove these lines to turn this test case into a positive test?

oswaldquek: Oh ok. Both tests test for invalid credentials but one test fails input format validation but this new one has valid formats but they are invalid in the sense they are not recognised by Worldpay.

oswaldquek (approved):

PR #2436, opened 2020-12-21
PP-7585 Publicauth client: use new baseclient

kbottla:
## WHAT
- Publicauth client: replaces usage of old base client (old-base.client.js) with new client (base.client.js) which provides better interface with sensible defaults


stephencdaly (approved):
Looks good

PR #2437, opened 2020-12-21
PP-7588 Demonstrate how pact dependency will be removed from fixtures

stephencdaly:
Want to remove the pact dependency so that fixtures can be used directly by Cypress test code, which cannot use Node dependencies.

The pact dependency is currently used to construct pacts for responses which include the appropriate matchers.

To do this:
- Remove `.getPactified()` methods returned by fixtures. Instead, "pactify" the response in the pact test itself.
- Add utility to return "pactifiers" to pactify certain responses.
- Remove `getPlain()` and `getObject()` methods returned by fixtures and just return the raw response object. We can construct the models returned by `getObject()` wherever we need them easily enough.

### DIFF
--- a/test/fixtures/user.fixtures.js
+++ b/test/fixtures/user.fixtures.js
@@ -322,23 +322,8 @@ module.exports = {
     }
   },
 
-  /**
-   * @param request Params override response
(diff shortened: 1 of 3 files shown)


kbottla (commented):
LGTM!

kbottla (approved):

PR #2439, opened 2020-12-21
PP-7588 Remove pact dependency from service fixtures

stephencdaly:
Remove dependency on pact-base which depends on the npm module @pact-foundation/pact/

Use the pactifier utility to pactify any responses in pact test constructed using the fixtures.

Any request objects used in pact tests don't need to be pactified, so just use the raw fixture response here.

Replace usages of getAsObject() with using the constructor for the Service model.

Calls to getPlain() for the fixtures are removed as the fixtures now return a plain object.

Also removes some fixtures that weren't used.


kbottla (approved):
LGTM!

PR #2440, opened 2020-12-21
PP-7588 Remove pact dependency from account and card type fixtures

stephencdaly:
Remove dependency on pact-base which depends on the npm module @pact-foundation/pact/ from `card.fixtures.js` and `gateway-account.fixtures,js`.

Use the pactifier utility to pactify any responses in pact test constructed using the fixtures.

Calls to getPlain() for the fixtures are removed as the fixtures now return a plain object

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json",
     "lines": null
   },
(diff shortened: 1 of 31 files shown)


kbottla (approved):
LGTM!

PR #2443, opened 2020-12-22
PP-7585 Connector client: use new baseclient

kbottla:
## WHAT
- Connector client : updated number of client methods to use new baseclient (base.client.js) which provides better interface.
- Also updated credentials controller to use async/await as it relies on old baseclient way of working (passing callback function to old baseclient)

### DIFF
--- a/app/controllers/credentials.controller.js
+++ b/app/controllers/credentials.controller.js
@@ -87,7 +87,7 @@ module.exports = {
     loadIndex(req, res, EDIT_NOTIFICATION_CREDENTIALS_MODE)
   },
 
-  updateNotificationCredentials: function (req, res) {
+  updateNotificationCredentials: async function (req, res) {
     const accountId = auth.getCurrentGatewayAccountId((req))
     const connectorUrl = CONNECTOR_URL + '/v1/api/accounts/{accountId}/notification-credentials'
     const { username, password } = _.get(req, 'body')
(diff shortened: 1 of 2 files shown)


app/controllers/credentials.controller.js
```diff
@@ -118,23 +118,25 @@ module.exports = {
...
-    }, function (connectorData, connectorResponse) {
-      var duration = new Date() - startTime
+    try {
+      await connectorClient().postAccountNotificationCredentials({
+        payload: { username, password },
+        correlationId: correlationId,
+        gatewayAccountId: accountId
+      })
+
+      let duration = new Date() - startTime
```
stephencdaly, line 128: this can be `const`

app/controllers/credentials.controller.js
```diff
@@ -118,23 +118,25 @@ module.exports = {
...
       logger.info(`[${correlationId}] - POST to ${url} ended - elapsed time: ${duration} ms`)
-      res.redirect(303, router.paths.yourPsp.index)
-    }).on('connectorError', function (err, connectorResponse) {
-      var duration = new Date() - startTime
+      return res.redirect(303, router.paths.yourPsp.index)
+    } catch (err) {
+      let duration = new Date() - startTime
```
stephencdaly, line 132: also `const`

app/services/clients/connector.client.js
```diff
@@ -155,23 +155,11 @@ ConnectorClient.prototype = {
...
-        method: 'GET',
-        description: 'get an account',
-        service: SERVICE_NAME
-      }
-
-      let callbackToPromiseConverter = createCallbackToPromiseConverter(context)
-
-      oldBaseClient.get(url, params, callbackToPromiseConverter, null)
-        .on('error', callbackToPromiseConverter)
+    let url = _accountUrlFor(params.gatewayAccountId, this.connectorUrl)
```
stephencdaly, line 158: again, `const` if it's not going to be overwritten

stephencdaly (approved):
Looks good.

A few instances where `let` should be `const` but fine to merge this and fix that later.

app/controllers/credentials.controller.js
```diff
@@ -118,23 +118,25 @@ module.exports = {
...
     var correlationId = req.headers[CORRELATION_HEADER] || ''
 
-    connectorClient().postAccountNotificationCredentials({
-      payload: { username, password },
-      correlationId: correlationId,
-      gatewayAccountId: accountId
-    }, function (connectorData, connectorResponse) {
-      var duration = new Date() - startTime
+    try {
+      await connectorClient().postAccountNotificationCredentials({
```
sfount, line 122: Now that these client methods are moved over from the old base clients pub/ sub method of error propagation it would no be possible to instantiate the connector client once and then use it in the methods, this would avoid creating a more expensive object once per request. 

See https://github.com/alphagov/pay-selfservice/pull/2377/files to confirm where and when this was happening with different clients, there are some comments on the areas it was impossible to make this change before these great changes you've proposed here. 

That pattern would be something like
```js
const conncerClient = new connectorClient()

function serviceMethod() {
  return connnectorClient.postAccountNotificationCredentials({})
}
```

sfount (commented):
Now that these client methods

app/controllers/credentials.controller.js
kbottla: Good point! I will look into instantiating client as above once connector client is fully migrated to use new baseclient.

app/controllers/credentials.controller.js
kbottla: Updated above and the rest in this PR https://github.com/alphagov/pay-selfservice/pull/2445

app/controllers/credentials.controller.js
kbottla: Changed connector client instantiation out of requests here https://github.com/alphagov/pay-selfservice/pull/2461

PR #2449, opened 2020-12-23
PP-7588 Remove pact dependency from remaining fixtures

stephencdaly:
Follows previous commits to remove the pact dependency from fixtures.
This finishes removing the dependency from all fixtures.


kbottla (approved):
LGTM!

PR #2450, opened 2020-12-23
PP-7585 Remove old baseclient

kbottla:
## WHAT
- Connector client: updated remaining methods to use new baseclient
- Removed old baseclient as it is no longer needed


app/services/email.service.js
```diff
@@ -15,117 +15,103 @@ const connectorClient = function () {
...
-          emailCollectionMode: data.email_collection_mode,
-          refundEmailEnabled: data.email_notifications.REFUND_ISSUED && data.email_notifications.REFUND_ISSUED.enabled
-        })
-      })
-      .catch(err => {
-        logger.info(`[${correlationId}] - GET account %s ended - elapsed time: %s ms`, accountID, new Date() - startTime)
-        clientFailure(err, reject, 'GET', correlationId, false)
-      })
-  })
+    logger.info(`[${correlationId}] - GET account %s ended - elapsed time: %s ms`, accountID, new Date() - startTime)
```
stephencdaly, line 25: I imagine we can remove all the request logging from this service. The client does its own logging for requests. Might want to be sure it's not used by any dashboards I guess, but I doubt these ones are.

test/unit/services/email.service.test.js
```diff
@@ -12,58 +11,69 @@ function getEmailService (connectorClientStub) {
...
+  ConnectorClient: function () {
+    return {
+      updateConfirmationEmail: function () {
+        return Promise.resolve()
+      },
+      updateConfirmationEmailEnabled: function () {
+        return Promise.resolve()
+      },
+      getAccount: function () {
+        return Promise.resolve(
```
stephencdaly, line 24: Would be good to use a fixture to construct this response while we're here.

stephencdaly (approved):
Nice work! Some small comments but shouldn't stop from merging this.

PR #2456, opened 2020-12-24
PP-7595 Refactor Edit Service Name controller to use Async / Await

iqbalgds:
Refactor code to use aync/await
- Update 'edit-service-name' controller
- Update 'service.service' calls to connector when editing a service name
- Update tests

### DIFF
--- a/app/controllers/edit-service-name.controller.js
+++ b/app/controllers/edit-service-name.controller.js
@@ -8,7 +8,7 @@ const formatPath = require('../utils/replace-params-in-path')
 const serviceService = require('../services/service.service')
 const { validateServiceName } = require('../utils/service-name-validation')
 
-exports.get = (req, res) => {
+function getServiceName (req, res) {
   let pageData = lodash.get(req, 'session.pageData.editServiceName')
 
   if (pageData) {
@@ -24,7 +24,7 @@ exports.get = (req, res) => {
   return responses.response(req, res, 'services/edit-service-name', pageData)
 }
 
-exports.post = (req, res) => {
+async function postEditServiceName (req, res, next) {
   const correlationId = lodash.get(req, 'correlationId')
   const serviceExternalId = lodash.get(req, 'service.externalId')
   const serviceName = lodash.get(req, 'body.service-name')
@@ -40,12 +40,16 @@ exports.post = (req, res) => {
     })
     res.redirect(formatPath(paths.editServiceName.index, req.service.externalId))
   } else {
-    return serviceService.updateServiceName(serviceExternalId, serviceName, serviceNameCy, correlationId)
-      .then(() => {
-        res.redirect(paths.serviceSwitcher.index)
-      })
-      .catch(err => {
-        responses.renderErrorView(req, res, err)
-      })
+    try {
+      await serviceService.updateServiceName(serviceExternalId, serviceName, serviceNameCy, correlationId)
+      res.redirect(paths.serviceSwitcher.index)
+    } catch (err) {
+      next(err)
+    }
   }
+}
+
+module.exports = {
+  get: getServiceName,
+  post: postEditServiceName
 }
--- a/app/services/service.service.js
+++ b/app/services/service.service.js
@@ -25,25 +25,25 @@ async function getGatewayAccounts (gatewayAccountIds, correlationId) {
     .map(gatewayAccount => new CardGatewayAccount(gatewayAccount).toMinimalJson())
 }
 
-function updateServiceName (serviceExternalId, serviceName, serviceNameCy, correlationId) {
+async function updateServiceName (serviceExternalId, serviceName, serviceNameCy, correlationId) {
   if (!serviceExternalId) {
     return Promise.reject(new Error(`argument: 'serviceExternalId' cannot be undefined`))
   }
-  return adminUsersClient.updateServiceName(serviceExternalId, serviceName, serviceNameCy, correlationId)
-    .then(result => {
-      // Update gateway account service names in connector
-      const gatewayAccountIds = lodash.get(result, 'gateway_account_ids', [])
-      return Promise.all(gatewayAccountIds.map(gatewayAccountId => {
-        if (gatewayAccountId && !isADirectDebitAccount(gatewayAccountId)) {
-          return connectorClient.patchServiceName(gatewayAccountId, serviceName, correlationId)
-        }
-      })
-      )
-        .then(() => result)
+
+  const result = await adminUsersClient.updateServiceName(serviceExternalId, serviceName, serviceNameCy, correlationId)
+
+  const gatewayAccountIds = lodash.get(result, 'gateway_account_ids', [])
+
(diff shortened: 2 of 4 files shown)


app/controllers/edit-service-name.controller.js
```diff
@@ -8,7 +8,7 @@ const formatPath = require('../utils/replace-params-in-path')
 const serviceService = require('../services/service.service')
 const { validateServiceName } = require('../utils/service-name-validation')
 
-exports.get = (req, res) => {
+exports.get = function getServiceName (req, res) {
```
stephencdaly, line 11: can we move the exports to the bottom of the file to to meet with our convention? https://pay-team-manual.cloudapps.digital/manual/coding-standards/node-js-style-guide.html#module-exports

app/controllers/edit-service-name.controller.js
```diff
@@ -40,12 +40,11 @@ exports.post = (req, res) => {
...
-        res.redirect(paths.serviceSwitcher.index)
-      })
-      .catch(err => {
-        responses.renderErrorView(req, res, err)
-      })
+    try {
+      await serviceService.updateServiceName(serviceExternalId, serviceName, serviceNameCy, correlationId)
+      res.redirect(paths.serviceSwitcher.index)
+    } catch (err) {
+      responses.renderErrorView(req, res, err)
```
stephencdaly, line 47: I think passing `err` to `renderErrorView` will try to use it as a message to display to the user. But in this case it might end up as something like `[Object object]` or something shown to the user 😬. Even if not, we shouldn't be displaying error messages like this to the user.

Can we change this so we call `next` and pass the error to it. The new error handling middleware code at the end of the stack will then handle it and call the error.

I know this isn't something you changed, but we should fix it.

app/services/service.service.js
```diff
@@ -24,25 +24,29 @@ async function getGatewayAccounts (gatewayAccountIds, correlationId) {
...
-        .then(() => result)
-    })
-    .then(result => {
-      return new Service(result)
-    })
+    )
+
+    return new Service(result)
+  } catch (err) {
+    logger.error(`[${correlationId}] Calling connector to update service name failed`)
```
stephencdaly, line 48: we should throw the error here to be handled by the controller

test/unit/controller/edit-service-name/post.controller.test.js
```diff
@@ -56,11 +56,11 @@ describe('Controller: editServiceName, Method: get', () => {
...
-      const result = editServiceNameCtrl.post(req, res)
-      if (result) {
-        result.then(() => done()).catch(done)
-      } else {
-        done(new Error('Didn\'t return a promise'))
+
+      try {
+        await editServiceNameCtrl.post(req, res) 
+      } catch (err) {
+        throw new Error('Didn\'t return a promise')
```
stephencdaly, line 63: don't need the try/catch here I think. If there's an error thrown it will fail the test anyway.

test/unit/controller/edit-service-name/post.controller.test.js
```diff
@@ -88,11 +88,11 @@ describe('Controller: editServiceName, Method: get', () => {
...
-      const result = editServiceNameCtrl.post(req, res)
-      if (result) {
-        done(new Error('Returned a promise'))
-      } else {
-        done()
+
+      try {
+        editServiceNameCtrl.post(req, res)
+      } catch (err) {
+        throw new Error('Returned a promise')
```
stephencdaly, line 95: I don't think this does the same thing as before. We should expect it to return a promise now as it's an async function, so just remove the try/catch

test/unit/services/service.service.test.js
```diff
@@ -213,13 +210,12 @@ describe('service service', function () {
...
-          setTimeout(() => {
-            expect(patchServiceName.callCount).to.equal(2)
-          }, 250)
-          expect(JSON.stringify(service)).to.deep.equal('{"gatewayAccountIds":[10,9,"DIRECT_DEBIT:adashdkjlq3434lk"]}')
-        }).should.notify(done)
+      const service = await serviceService.updateServiceName(externalServiceId, newServiceName, correlationId)
+
+      setTimeout(() => {
+        expect(patchServiceName.callCount).to.equal(2)
+      }, 250)
```
stephencdaly, line 217: I don't think we need the timeout now we're awaiting the result?

stephencdaly (changes requested):
A few things, some of which were existing issues which I think we should fix while you're looking at this code.

app/services/service.service.js
stephencdaly: I reckon we don't need the logging either - the client code will log an appropriate error. So just remove the try/catch from this function maybe and let any errors propagate through to the controller.

app/controllers/edit-service-name.controller.js
iqbalgds: Done

app/controllers/edit-service-name.controller.js
iqbalgds: Done

app/services/service.service.js
iqbalgds: Done - removed the try/catch.

test/unit/controller/edit-service-name/post.controller.test.js
iqbalgds: Removed the try/catch block.

test/unit/services/service.service.test.js
iqbalgds: Done

test/unit/controller/edit-service-name/post.controller.test.js
```diff
@@ -56,25 +52,20 @@ describe('Controller: editServiceName, Method: get', () => {
...
 
-    it(`should call 'responses.renderErrorView' with req, res and the error received from the client`, () => {
-      expect(mockResponses.renderErrorView.called).to.equal(true)
-      expect(mockResponses.renderErrorView.args[0]).to.include(req)
-      expect(mockResponses.renderErrorView.args[0]).to.include(res)
-      expect(mockResponses.renderErrorView.args[0][2] instanceof Error).to.equal(true)
-      expect(mockResponses.renderErrorView.args[0][2].message).to.equal('somet went wrong')
+    it(`should call 'next' with the error`, () => {
+      expect(next.called).to.equal(true)
+      expect(next.args[0].toString()).to.equal('Error: something went wrong')
```
stephencdaly, line 63: using the following will give nicer error messages should this fail:

```
const expectedError = sinon.match.instanceOf(Error)
     .and(sinon.match.has('message', 'something went wrong'))
sinon.assert.calledWith(next, expectedError)
```

stephencdaly (approved):
Nice work

test/unit/controller/edit-service-name/post.controller.test.js
stephencdaly: although what's here is fine, just for future reference

PR #2457, opened 2020-12-24
PP-7295 Send service created date in go-live zendesk ticket

stephencdaly:
We now store the date a service was greated on. Include it in the request to go live Zendesk ticket so we have more information about the time taken between creating the test service and requesting a live account when we look at these tickets.


alexbishop1 (commented):
Should the commit message reference PP-7295?

stephencdaly:
Oh yeh didn't realise there was a story, thanks

stephencdaly:
I read about the requirement from a story, but then forgot there was one.

alexbishop1 (approved):
I assume if there’s no creation date for a service, we’ll just use the empty string, which is fine (we do a similar thing for the IP address, which we only include if it’s a Stripe service).

stephencdaly:
Updated to say that we don't have the date if it isn't available.

alexbishop1 (approved):

PR #2461, opened 2020-12-24
PP-7585 Instantiate connector client as singleton and cleanup logging

kbottla:
## WHAT
- Moves connector client instantiation out of request methods so connector client object is instantiated once and not per request
- Removed request logging from email.service & credentials.controller as both of these now uses new baseclient which logs the same


stephencdaly (approved):
Nice cleanup

PR #2465, opened 2020-12-29
PP-7585 Fix client method postAcceptedCardsForAccount

kbottla:
## WHAT
- Fixes connector client method `postAcceptedCardsForAccount`: Incorrect data is being sent to connector. It should just be `payload` from controller.
 Caused by https://github.com/alphagov/pay-selfservice/commit/7b793b5d944652fbf7ef17790f25950d5e846794#diff-0a8d6b0be878b39296cfff51f4a3c04eef6f05c5a39740158c1f2a4e86001190L390


sfount (approved):
It concerns me that this wasn't flagged up by any of the app test coverage.

We need this to go through to talk correctly with the backend but it would be good to make sure there's some kind of integration test that expects `POST`ing exactly as the backend would.

PR #2467, opened 2020-12-30
PP-7585 Refactor adminUsers client to remove instance correlationId

kbottla:
## WHAT
- Refactored adminUsers client methods to move `correlationId` to each individual call rather than being set at the instance level.  This allows code that calls this adminUsers client to instantiate once using singleton pattern.
- Updated code to instantiate adminUsers client once rather than per call


sfount (approved):
Looks great 👍

PR #2468, opened 2020-12-31
PP-7595 Switch payment links controllers to use asnyc/await

kbottla:
## WHAT
- Replaces promise code to use asnyc/await syntax which makes the code easier to follow.

### DIFF
--- a/app/controllers/payment-links/get-delete.controller.js
+++ b/app/controllers/payment-links/get-delete.controller.js
@@ -6,30 +6,27 @@ const productsClient = require('../../services/clients/products.client.js')
 const publicAuthClient = require('../../services/clients/public-auth.client.js')
 const auth = require('../../services/auth.service.js')
 
-module.exports = (req, res) => {
+module.exports = async (req, res) => {
   const gatewayAccountId = auth.getCurrentGatewayAccountId(req)
-  productsClient.product.getByProductExternalId(gatewayAccountId, req.params.productExternalId)
-    .then(product => {
-      const deleteToken = publicAuthClient.deleteTokenForAccount({
-        accountId: gatewayAccountId,
-        correlationId: req.correlationId,
-        payload: {
-          token: product.apiToken
-        }
-      })
-      const deleteProduct = productsClient.product.delete(gatewayAccountId, req.params.productExternalId)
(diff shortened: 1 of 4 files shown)


oswaldquek (approved):
LGTM

app/controllers/payment-links/get-delete.controller.js
```diff
@@ -6,30 +6,27 @@ const productsClient = require('../../services/clients/products.client.js')
...
+
+    const deleteToken = publicAuthClient.deleteTokenForAccount({
+      accountId: gatewayAccountId,
+      correlationId: req.correlationId,
+      payload: {
+        token: product.apiToken
+      }
     })
+    const deleteProduct = productsClient.product.delete(gatewayAccountId, req.params.productExternalId)
+    await Promise.all([deleteToken, deleteProduct])
```
oswaldquek, line 23: Didn't know we can do this. Nice.

app/controllers/payment-links/get-manage.controller.js
```diff
@@ -10,26 +10,25 @@ const { renderErrorView } = require('../../utils/response')
...
+      productsLength: paymentLinks.length,
+      englishPaymentLinks,
+      welshPaymentLinks,
+      paths,
+      externalServiceId
+    }
+    return response(req, res, 'payment-links/manage', pageData)
+  } catch (err) {
+    logger.error(`[requestId=${req.correlationId}] Get ADHOC product by gateway account id failed - ${err.message}`)
+    return renderErrorView(req, res)
```
sfount, line 32: This PR is great! So much clearer. 

Quick comment that now that we've set up error handling appropriately for the middleware architecture generic error handling can be done downstream of the controller (for custom error handling like redirecting the controller can take of those). 

The following would have the equivalent functionality but would be one less place the developer has to think about responding. 

```javascript
next(err)
```

kbottla: 👍🏽 Updated in https://github.com/alphagov/pay-selfservice/pull/2471/files#diff-e3c9d7742c1ef0d275b4a361576dda5dd01c427955b6c94663e10f19dc739cd6R29

PR #2471, opened 2021-01-04
PP-7595 Switch more promises to use asnyc/await

kbottla:
## WHAT
- Replaces promise code to use asnyc/await syntax which makes the code easier to follow.


SandorArpa (approved):
LOKTM

PR #2472, opened 2021-01-04
PP-7422: Handle bad result returned by connector

adityapahuja:
This change updates post-flex controller to save valid Worldpay 3DS Flex credentials only and only if the result is valid. This change is necessary because the result can contain the value other than "valid" and "invalid". It adds 1 more test to Cypress to test the following scenario.

1. Connector returns result = bad data after checking Worldpay 3DS Flex credentials.

Please note that Self Service application has no way to check that response's body is null or empty before processing it. If the response contains no body or other data,  post-flex controller will attempt to retrieve result from the response's body and this will cause Self Service application to display a generic problem page.

### DIFF
--- a/test/cypress/integration/settings/your-psp.cy.test.js
+++ b/test/cypress/integration/settings/your-psp.cy.test.js
@@ -32,6 +32,11 @@ describe('Your PSP settings page', () => {
     issuer: '5bd9e0e4444dce153428c942', // pragma: allowlist secret
     jwt_mac_key: 'ffffffff-ffff-ffff-ffff-ffffffffffff'
   }
+  const testBadResultFlexCredentials = {
+    organisational_unit_id: '5bd9b55e4444761ac0af1c83',
+    issuer: '5bd9e0e4444dce153428c943', // pragma: allowlist secret
+    jwt_mac_key: 'fa2daee2-1fbb-45ff-4444-52805d5cd9e3'
+  }
 
   function setupYourPspStubs (opts = {}) {
     let user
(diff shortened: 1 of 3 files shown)


test/cypress/integration/settings/your-psp.cy.test.js
```diff
@@ -78,13 +83,20 @@ describe('Your PSP settings page', () => {
       issuer: testFailureFlexCredentials.issuer,
       jwt_mac_key: testFailureFlexCredentials.jwt_mac_key
     })
+    const postCheckWorldpay3dsFlexCredentialsReturnsBadResult = gatewayAccountStubs.postCheckWorldpay3dsFlexCredentialsWithBadResult({
+      gatewayAccountId: gatewayAccountId,
+      organisational_unit_id: testBadResultFlexCredentials.organisational_unit_id,
```
danworth, line 88: If you wanted to you could use the `...` operator on the `testBadResultFlexCredentials` object rather than having to reference each of its attributes individually, like
```
> { gatewayAccountId: "someid", ...testBadResultFlexCredentials }
{
  gatewayAccountId: 'someid',
  organisational_unit_id: '5bd9b55e4444761ac0af1c83',
  issuer: '5bd9e0e4444dce153428c943',
  jwt_mac_key: 'fa2daee2-1fbb-45ff-4444-52805d5cd9e3'
}
```

Just thought I'd mention it in passing, since you posted the PR in the Kaizen channel I'll let someone from your team review the PR, looks ok to me though 👍

adityapahuja: Thanks @danworth. I have made the changes.

oswaldquek (approved):
LGTM. Can we have another test where connector doesn't return a 200?

adityapahuja:
> LGTM. Can we have another test where connector doesn't return a 200?

It is already covered by the previous PR (https://github.com/alphagov/pay-selfservice/pull/2428). It tests `should display generic problem page when checking 3DS Flex credentials fails`. `checking 3DS Flex credentials fails` is where the connector will return 500.

PR #2474, opened 2021-01-08
BAU Use design system notification banner component

stephencdaly:
Use the new design system notification banner component https://design-system.service.gov.uk/components/notification-banner/ when we display success messages back to the user.

Previously, looked like this:

Screenshot 2021-01-07 at 18 26 45

Now looks like this:

Screenshot 2021-01-07 at 18 26 36


iqbalgds:
Nice work 👍

iqbalgds (approved):

PR #2479, opened 2021-01-12
PP-7583 Url Structure > Add CSRF middleware to account router

iqbalgds:
- Add the CSRF middleware function 'validateAndRefreshCsrf' to the account router.

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -102,7 +102,7 @@ module.exports.paths = paths
 
 module.exports.bind = function (app) {
   const account = new Router({ mergeParams: true })
-  account.use(getServiceAndAccount, userIsAuthorised)
+  account.use(getServiceAndAccount, userIsAuthorised, ensureSessionHasCsrfSecret, validateAndRefreshCsrf)
 
   app.get('/style-guide', (req, res) => response(req, res, 'style_guide'))
 
--- a/test/integration/payment-types/payment-types.it.test.js
+++ b/test/integration/payment-types/payment-types.it.test.js
@@ -3,6 +3,7 @@
 const request = require('supertest')
 const nock = require('nock')
 const chai = require('chai')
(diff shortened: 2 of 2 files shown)


app/routes.js
```diff
@@ -102,7 +102,7 @@ module.exports.paths = paths
 
 module.exports.bind = function (app) {
   const account = new Router({ mergeParams: true })
-  account.use(getServiceAndAccount, userIsAuthorised)
+  account.use(getServiceAndAccount, userIsAuthorised, validateAndRefreshCsrf)
```
kbottla, line 105: Looks like we need `ensureSessionHasCsrfSecret` where creates new csrfSecret if one doesn't exist on session.

```suggestion
  account.use(getServiceAndAccount, userIsAuthorised, ensureSessionHasCsrfSecret, validateAndRefreshCsrf)
```

kbottla (changes requested):

app/routes.js
iqbalgds: Just updated it - thanks.

kbottla (approved):
LGTM!

app/routes.js
```diff
@@ -102,7 +102,7 @@ module.exports.paths = paths
 
 module.exports.bind = function (app) {
   const account = new Router({ mergeParams: true })
-  account.use(getServiceAndAccount, userIsAuthorised)
+  account.use(getServiceAndAccount, userIsAuthorised, ensureSessionHasCsrfSecret, validateAndRefreshCsrf)
```
sfount, line 105: Apologies I missed this before it was reviewed, I don't think you need `ensureSessionHasCsrfSecret` here, I believe that's done once when the session is initially authorised and created and should then be subsequently available -- I think the issue with the tests passing was with the payment types integration test

PR #2481, opened 2021-01-12
PP-7583 Cleanup mock session test helper

stephencdaly:
Functional tests use helper functions in `mock-session.js` to create
the express app with the desired session.

Clean this up by making it more clear how and when the current
gateway account cookie is set by the helper methods.


kbottla (approved):
LGTM!

PR #2482, opened 2021-01-12
PP-7583 Remove usages of getCurrentGatewayAccountId from controllers

stephencdaly:
This method contains various logic to look at the cookie and the
available services for the user to determine which gateway account id
should be retrieved.

It is also called by the current middleware that retrieves the gateway
account and puts it on the `req` object as `req.account`. This means
that when it is called in controllers, it will always get the same
gateway account id as `req.account.gateway_account_id`.

Therefore can replace usage in controllers with getting
`req.account.gateway_account_id`.

Fix all the places in function tests where we either weren't nocking the
request to connector to get the gateway account or it wasn't returning
a response with all the expected fields.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json",
     "lines": null
   },
-  "generated_at": "2021-01-12T14:15:13Z",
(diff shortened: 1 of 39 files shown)


SandorArpa (approved):
LOKTM

PR #2485, opened 2021-01-13
PP-7583 Move 3ds routes to use account URL structure

stephencdaly:
Use the account router for these routes so that the URL contains the
account external ID and we use our new middleware for resolving the
service/account and checking the user is authorised.

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -88,11 +88,15 @@ const stripeSetupDashboardRedirectController = require('./controllers/stripe-set
 const {
   healthcheck, registerUser, user, dashboard, selfCreateService, transactions, credentials,
   apiKeys, serviceSwitcher, teamMembers, staticPaths, inviteValidation, editServiceName, merchantDetails,
-  notificationCredentials: nc, paymentTypes: pt, emailNotifications: en, toggle3ds: t3ds, toggleMotoMaskCardNumberAndSecurityCode, prototyping, paymentLinks,
+  notificationCredentials, emailNotifications: emailNotifications, toggleMotoMaskCardNumberAndSecurityCode, prototyping, paymentLinks,
   requestToGoLive, policyPages, stripeSetup, stripe, digitalWallet,
   settings, yourPsp, allServiceTransactions, payouts
 } = paths
-const { toggleBillingAddress: billingAddress } = paths.account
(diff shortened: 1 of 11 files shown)


app/routes.js
```diff
@@ -88,7 +88,7 @@ const stripeSetupDashboardRedirectController = require('./controllers/stripe-set
 const {
   healthcheck, registerUser, user, dashboard, selfCreateService, transactions, credentials,
   apiKeys, serviceSwitcher, teamMembers, staticPaths, inviteValidation, editServiceName, merchantDetails,
-  notificationCredentials: nc, paymentTypes: pt, emailNotifications: en, toggle3ds: t3ds, toggleMotoMaskCardNumberAndSecurityCode, prototyping, paymentLinks,
+  notificationCredentials: nc, paymentTypes: pt, emailNotifications: en, toggle3ds, toggleMotoMaskCardNumberAndSecurityCode, prototyping, paymentLinks,
```
sfount, line 91: nice cleanup 👍 . It would be good if this could also move under `paths.account` to make it clear it's now using the new structure and allow any code that uses that group to include this route

stephencdaly: done

app/utils/nav-builder.js
```diff
@@ -94,13 +94,17 @@ const adminNavigationItems = (currentPath, permissions, type, paymentProvider, a
...
-      url: formatAccountPathsFor(paths.paymentTypes.index, account.external_id),
-      current: pathLookup(currentPath, paths.paymentTypes.index),
+      url: formatAccountPathsFor(paths.account.paymentTypes.index, account.external_id),
+      current: pathLookup(currentPath, paths.account.paymentTypes.index),
       permissions: permissions.payment_types_read && type === 'card'
     }
   ]
 }
 
+// const pathMatches(url, paths) {
```
sfount, line 104: Can be removed I think

sfount (approved):

app/utils/nav-builder.js
stephencdaly: oops yeh thanks accidental WIP

stephencdaly: will remove in next PR just to save build time

PR #2486, opened 2021-01-13
PP-7583 Fix nav builder

stephencdaly:
Fix the nav builder so that highlighting the current page in both the top navigation bar and the side settings navigation is highlighted:
- when the URL contains the `/account/ACCOUNT_EXTERNAL_ID/` prefix.
- for pages it didn't previously work for because they weren't included in the list of paths to match on.


sfount (approved):
Clears up the 'current' highlighting functionality well 👍

PR #2488, opened 2021-01-14
PP-7583 Move email notifications routes

stephencdaly:
Change email notifications routes to use the new account URL structure.

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -88,12 +88,13 @@ const stripeSetupDashboardRedirectController = require('./controllers/stripe-set
 const {
   healthcheck, registerUser, user, dashboard, selfCreateService, transactions, credentials,
   apiKeys, serviceSwitcher, teamMembers, staticPaths, inviteValidation, editServiceName, merchantDetails,
-  notificationCredentials, emailNotifications: emailNotifications, toggleMotoMaskCardNumberAndSecurityCode, prototyping, paymentLinks,
+  notificationCredentials, toggleMotoMaskCardNumberAndSecurityCode, prototyping, paymentLinks,
   requestToGoLive, policyPages, stripeSetup, stripe,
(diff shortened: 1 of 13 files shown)


app/routes.js
```diff
@@ -253,19 +253,19 @@ module.exports.bind = function (app) {
...
+  account.post(emailNotifications.update, permission('email-notification-paragraph:update'), getEmailNotification, paymentMethodIsCard, emailNotificationsController.update)
+  account.get(emailNotifications.collection, permission('email-notification-template:read'), getEmailNotification, paymentMethodIsCard, emailNotificationsController.collectionEmailIndex)
+  account.post(emailNotifications.collection, permission('email-notification-toggle:update'), getEmailNotification, paymentMethodIsCard, emailNotificationsController.collectionEmailUpdate)
+  account.get(emailNotifications.confirmation, permission('email-notification-template:read'), getEmailNotification, paymentMethodIsCard, emailNotificationsController.confirmationEmailIndex)
+  account.post(emailNotifications.confirmation, permission('email-notification-toggle:update'), getEmailNotification, paymentMethodIsCard, emailNotificationsController.confirmationEmailUpdate)
+  account.post(emailNotifications.off, permission('email-notification-toggle:update'), getEmailNotification, paymentMethodIsCard, emailNotificationsController.confirmationEmailOff)
```
sfount, line 265: As discussed do we want to remove `paymentMethodIsCard`? 

I would also be in favour of moving the code from `getEmailNotification` from the middleware 
* it uses old error rendering directly in response in middleware 
* it's a service "fetch data" model and would be a lot clearer happening at the begining of the controller

If we want to keep this PR clear and do a follow up that works for me, what do you think?

sfount (approved):

PR #2492, opened 2021-01-14
PP-7445 Update accounts URL utility

sfount:
Simple utility to upgrade or redirect account specific URLs from the
current method to the proposed URL structure.

This will:
* redirect users who are following older links
* allow a seamless change for users actively using the admin tool during a
release
* log upgrades so that we can see how often this is happening
and when legacy code can be removed.

### DIFF
--- a/app/utils/gateway-account-urls.js
+++ b/app/utils/gateway-account-urls.js
@@ -1,0 +1,47 @@
+'use strict'
+// check if a missed URL (404) is a URL that has been upgraded during the
+// account URL structure change. When this utility is reporting few or no
+// upgrades it can be removed
+const urlJoin = require('url-join')
+const paths = require('../paths')
+const formattedPathFor = require('./replace-params-in-path')
+const flattenNestedValues = require('./flatten-nested-values')
+
+// only flatten paths once given the singleton module export patten, these
+// should never change after the server spins up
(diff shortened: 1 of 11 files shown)


app/utils/gateway-account-urls.js
```diff
@@ -0,0 +1,47 @@
...
+const allAccountPaths = flattenNestedValues(paths.account)
+const templatedAccountPaths = allAccountPaths.filter((path) => path.includes(':'))
+
+const removeEmptyValues = (value) => !!value
+
+function isLegacyAccountsUrl (url) {
+  if (allAccountPaths.includes(url)) {
+    return true
+  } else {
+    // the path isn't directly in the list, check to see if it's a templated value
```
stephencdaly, line 21: I guess we haven't moved over any templated URLs yet so we can't write a test for this part? It would probably be good to have a test when we can add one, so maybe we should come back to this.

stephencdaly (approved):
Looks good

app/utils/gateway-account-urls.js
sfount: Agreed, ideally that should be moved out to be tested independently but we'll update this as soon as a templated path is moved over, great spot.

PR #2498, opened 2021-01-15
PP-7683 Tidy up routes file

stephencdaly:
Group routes by the level they are at, and do some general tidying.

Remove unnecessary middleware:

`getAccount` and `permission` from all services transactions routes

- these should not be called as this route is accessed when the user is not within a service.

`getAccount` from feedback route

- we use `resolveService` on this route to send the service id in the feedback, but don't need `getAccount`.


sfount (approved):
Much easier to navigate 👍

PR #2499, opened 2021-01-15
PP-7445 Set account external ID on login

sfount:
As we transition URL structure for accounts only new middleware routes
and the service switcher will be setting the external id required to
redirect to the correct URLs.

Ensure the external ID is set when the user logs in, this should then be
correct until the ID is changed by a known redirect controller.


SandorArpa (approved):
LOKTM

stephencdaly (approved):

PR #2500, opened 2021-01-15
PP-7583 Add new account routes for routes we will move

stephencdaly:
Add the new routes which contain the account external ID for routes we
are going to switch over but haven't yet.

This is so that these routes will exist before we come to redirecting to
them from other pages.

Without doing this first, users may have issues while we are deploying
the app, where some instances of selfservice have the new route
available but others don't. This could cause users to see 404s on
redirects.


kbottla (approved):
LGTM!

PR #2507, opened 2021-01-18
Revert "PP-7583 Move your psp routes"

stephencdaly:
This reverts commit f86dee8341b20553280abea18df1c99beac6fbc9.

Reverting as it breaks end to end tests due to the credentials not being available on the new get account by external ID connector endpoint.


sfount (approved):

SandorArpa (approved):
LOKTM

PR #2509, opened 2021-01-18
PP-7583 Payment links account url structure

sfount:
Update all payment links routes and references to point to the new
account URL structure.

Removes the functional tests for every payment link page, they are brittle
and now largely covered by Cypress.


SandorArpa (approved):
LGTM

PR #2510, opened 2021-01-18
BAU Stop `--fix` lint from running in CI

sfount:
Running `standard --fix` will automatically update files that standard
can fix by itself, when standard does this it will exit with a 0 code as
it has successfully fixed files.

Linting should exit non 0 if there are any issues flagged, running the
lint command resolves this.

This hasn't been too big of an issue because major issue or issues that
standard can't automatically resolve still exit non-0 even with the
`--fix` flag provided.

* cleanup some manual calling of `node_modules/`


SandorArpa (approved):
LOKTM

sfount:
This should be done as part of a large Node linting/ devexp ticket. Cautious of changing CI behaviour without thinking it through.

PR #2511, opened 2021-01-18
BAU Remove unused dependencies

sfount:
Some dependencies are flagged as not being used, these continue to
require maintenance and increase the apps risk surface.

Some are used indirectly by gulp and have been omitted here.


SandorArpa (approved):
LOKTM

PR #2515, opened 2021-01-19
PP-7668 URL Restructure - Transactions

iqbalgds:
With @SandorArpa

- Update to use the new '/account/<external-id>' format for transaction pages.
  - Update controllers.
  - Update templates.
  - Update functional tests.
  - Update UI template tests

### DIFF
--- a/app/controllers/transactions/transaction-detail-redirect.controller.js
+++ b/app/controllers/transactions/transaction-detail-redirect.controller.js
@@ -1,23 +1,31 @@
 'use strict'
 
-const paths = require('../../paths')
 const { userServicesContainsGatewayAccount } = require('../../utils/permissions')
 const Ledger = require('../../services/clients/ledger.client')
 const { renderErrorView } = require('../../utils/response.js')
 const router = require('../../routes')
+const formatAccountPathsFor = require('../../utils/format-account-paths-for')
+const { ConnectorClient } = require('../../services/clients/connector.client')
+const connector = new ConnectorClient(process.env.CONNECTOR_URL)
 
 const defaultMsg = 'Error processing transaction view'
 const notFound = 'Charge not found'
 
-module.exports = async (req, res) => {
+module.exports = async function redirectToTransactionDetail (req, res) {
   const chargeId = req.params.chargeId
+
   try {
-    let charge = await Ledger.transactionWithAccountOverride(chargeId)
+    const charge = await Ledger.transactionWithAccountOverride(chargeId)
     if (userServicesContainsGatewayAccount(charge.gateway_account_id, req.user)) {
       req.gateway_account.currentGatewayAccountId = charge.gateway_account_id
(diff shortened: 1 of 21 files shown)


app/controllers/transactions/transaction-detail-redirect.controller.js
```diff
@@ -1,23 +1,23 @@
...
+module.exports = async function redirectToTransactionDetail (req, res) {
   const chargeId = req.params.chargeId
   try {
     let charge = await Ledger.transactionWithAccountOverride(chargeId)
     if (userServicesContainsGatewayAccount(charge.gateway_account_id, req.user)) {
       req.gateway_account.currentGatewayAccountId = charge.gateway_account_id
       req.session = { ...req.session, backLink: req.header('Referer') }
       charge = null
-      res.redirect(302, router.generateRoute(paths.transactions.detail, { chargeId }))
+      res.redirect(302, router.generateRoute(formatAccountPathsFor(router.paths.account.transactions.detail, req.account.external_id), { chargeId }))
```
sfount, line 20: No need for `router.generateRoute` here, the format account paths works generically similar to the original formatter, it should be possible to:

```suggestion
      res.redirect(302, formatAccountPathsFor(router.paths.account.transactions.detail, req.account.external_id, chargeId))
```

A little bit less to understand and potentially then able to remove the dependency on router. I think this is true of all of the formatters in this PR.

app/views/transaction-detail/_refund.njk
```diff
@@ -1,5 +1,4 @@
-
-<form id="refundForm" action="/transactions/{{charge_id}}/refund" method="post" class="target-to-show {% if flash.genericError %}active{% endif %}">
+<form id="refundForm" action="{{refundUrl}}" method="post" class="target-to-show {% if flash.genericError %}active{% endif %}">
```
sfount, line 1: `formatAccountPathsFor` is available to the template, potentially using it here would reduce the number of things needed to be passed in to the template 

```suggestion
<form id="refundForm" action="{{ formatAccountPathsFor(routes.account.transactions.refund, currentGatewayAccount.external_id, charge_id) }}" method="post" class="target-to-show {% if flash.genericError %}active{% endif %}">
```

sfount (commented):
Code looks good, a few quick URL formatting things I noticed; depending on others opinions I don't think they'd need to be blocking

app/controllers/transactions/transaction-detail-redirect.controller.js
iqbalgds: Good spot - looks cleaner.
Done.

app/views/transaction-detail/_refund.njk
iqbalgds: Good spot - done.

iqbalgds:
Good spots - ready for re-review.

PR #2519, opened 2021-01-21
PP-7629 Account URL upgrader redirect to login without session

sfount:
If the user is not logged in but have directly visited a legacy account
specific URL we may not have all the information we need to direct
them to the correct account.

In this case previously middleware on those URLs would have redirected
the user to login.

Update the URL upgrader to factor in this scenario.

Prove this by removing the API keys route from the legacy middleware
auth stack, this page is currently accessed by ZAP tests before having ever
logged in.


kbottla (approved):
LGTM!

PR #2521, opened 2021-01-22
BAU: Fix all service transactions download

kbottla:
## WHAT
- All live service transactions download link doesnot appear for users as we check for transaction.read permission,
  but this route is not tied to specific service and we can show the link directly (other checks about number of transactions before transactions download link appears is still in place).
  All live services transactions controllers (get, download) has checks in place to render error page if user doesn't have access to live services.


SandorArpa (approved):
LOKTM

PR #2522, opened 2021-01-22
PP-7677 Add account dashboard route

stephencdaly:
Add new route for the account dashboard. Adding this before merging changes to redirect to this route so that there is no risk of getting 404s on redirects when we deploy changes to redirect to the new route.


kbottla (approved):
LGTM!

PR #2530, opened 2021-01-25
PP-7521 Move csrf to server level middleware

kbottla:
## WHAT

- Removes csrf middleware (our own) and replaces by express csurf (https://github.com/expressjs/csurf) as server level middleware for CSRF token creation and validation.
- When initialising csrf middleware, 'value' passed is a function that extracts `csrfToken` (field currently used by forms for hidden csrf token) from request body. And the middleware also sets 'csrf' token on response local variable so the token is available for all routes and views. This removes the need for additional functionality on routes middleware.
- Healthcheck route has been moved out of routes to server level, as this route need to be defined before csrf is introduced to the stack (see https://github.com/expressjs/csurf#ignoring-routes). Otherwise healthcheck requests will be rejected with `misconfigured csrf` error.

- 'csrf' library is moved as 'dev' dependency as we still use it to generate csrf token for functional tests ('csurf' library uses 'csrf' for underlying token mechanism)


server.js
```diff
@@ -39,6 +41,19 @@ function warnIfAnalyticsNotSet () {
   }
 }
 
+function addCsrfMiddleware (app) {
+  app.use(csrf({
+    value: function (req) {
+      return req.body && req.body.csrfToken
```
sfount, line 47: I suppose it's worth explicitly saying that the server will only be supporting CSRF validation through POST requests and will ignore headers/ query strings for now. That's in-line with the existing behaviour so looks like a good move.

sfount (approved):
So much simpler to understand! Reading through the server entry point configuration is enough to know routes will be validated against CSRF and how to add the token to templates. Looks great 👍

server.js
kbottla: Yeah good call. Will add a comment in next PR to make it clear

PR #2533, opened 2021-01-26
PP-7641 Add getAccount middleware to feedback routes

kbottla:
## WHAT
- When submitting feedback, users are shown account specific menu (Dashboard , transactions ...) which is currently broken as we don't get account (in the middleware) anymore and also account info is also passed to zendesk.
- Adding getAccount middleware back to feedback routes to fix broken links while we decide on moving feedback to service level or  keeping it at account level (in which case we need to update feedback pages to new account URL structures)


SandorArpa (approved):
LOKTM

PR #2536, opened 2021-01-27
PP-7641 Remove paymentmethodIsCard middleware

kbottla:
## WHAT
- Removed `paymentMethodIsCard` middleware as we are currently supporting card payments only
- Also refactored few middleware to pass errors to next() directly which will be handled by error-handler middleware
  - Restrict to sandbox middleware now throws 404 error instead of 403. As we have switched to new URL structure, users shouldn't be relying on cookie and incorrectly access sandbox specific pages.


stephencdaly (approved):
LGTM

PR #2537, opened 2021-01-28
PP-7641 Remove getStripeAccount middleware

kbottla:
## WHAT
- Removed getStripeAccount middleware and inlined the call to connector (to get Stripe account) in controllers.
  Only post controllers (for stripe setup) need to get Stripe account from connector.


stephencdaly (approved):
Looks good

PR #2538, opened 2021-01-29
PP-7641 Remove check-bank-details-not-submitted middleware

kbottla:
## WHAT
- Removed check bank details not submitted middleware and inlined the same checks in get & post controllers


app/controllers/stripe-setup/bank-details/get.controller.js
```diff
@@ -1,7 +1,19 @@
...
+const paths = require('../../../paths')
+const formatAccountPathsFor = require('../../../utils/format-account-paths-for')
 
-module.exports = (req, res) => {
+module.exports = (req, res, next) => {
+  const stripeAccountSetup = req.account.connectorGatewayAccountStripeProgress
+
+  if (!stripeAccountSetup) {
+    return next(new Error('Stripe setup progress is not available on request'))
+  } else {
```
stephencdaly, line 12: if you wanted to save an extra few lines and some indentation, you could remove the `else` as we return if the `stripeAccountSetup` isn't present.

app/controllers/stripe-setup/bank-details/get.controller.js
```diff
@@ -1,7 +1,19 @@
...
 
 const { response } = require('../../../utils/response')
+const paths = require('../../../paths')
+const formatAccountPathsFor = require('../../../utils/format-account-paths-for')
 
-module.exports = (req, res) => {
+module.exports = (req, res, next) => {
+  const stripeAccountSetup = req.account.connectorGatewayAccountStripeProgress
+
+  if (!stripeAccountSetup) {
```
sfount, line 10: With the current middleware setup is there ever a reason this wouldn't be available?

I believe these routes are already restricted to live Stripe accounts, if the call to Connector to get the Stripe account setup failed we would have likely handled that error and not

If it's a safe assumption we can make it would help avoid the additional checks on each route
```
const hasSetupBankAccount = req.account.connectorGatewayAccountStripeProgress.bankAccount

if (hasSetupBankAccount) {
  // ...
}
response(req, res, 'stripe-setup/bank-details/index')
```

Being concise probably isn't important here so no need to make the change for legibility.  Just wondering if we'll ever actually encounter this code.

stephencdaly (approved):
Looks good

app/controllers/stripe-setup/bank-details/get.controller.js
kbottla: Good spot! will update in next PR.

app/controllers/stripe-setup/bank-details/get.controller.js
kbottla: In the get-service-and-gateway-account middleware `connectorGatewayAccountStripeProgress` is set conditionally
https://github.com/alphagov/pay-selfservice/blob/a5df5d98e944924e17255108bf7cc02ced4b7cf6/app/middleware/get-service-and-gateway-account.middleware.js#L27, so there is a possibility of it not being available sometimes (although I doubt this happening)

PR #2546, opened 2021-02-02
PP-7625 Sort services by live/test then alphabetically

stephencdaly:
If the `ENABLE_MY_SERVICES_AS_DEFAULT_VIEW` environment variable is set to true, sort services on the "My services" page with services having a live account sorted to the top and secondary sorting by name.

Continue to sort by ID if the feature flag is not set to true.

Remove code supporting direct debit accounts from the "My services" controller.

## Note

Recommended to hide whitespace differences for review as some indentation has changed.


SandorArpa (approved):
LGTM

PR #2547, opened 2021-02-02
PP-7445 URL Restructure > Feedback form

iqbalgds:
- Update URL structure for `feedback` routes so it no longer uses `service`
  or `gateway account` in the URL.
- Update routes.js
  - feedback routes - remove middleware for getting the gateway account or service.
- Update the Feedback NJK file
  - Add a `back to my services` link
- Update functional tests
  - Rename files to make it clear they are functional tests.
  - Remove the stubs for `service` and `gateway account` details.
- Add Cypress test for the feedback form.

image


app/controllers/feedback/post-index.controller.js
```diff
@@ -1,24 +1,20 @@
...
+  const message = `Feedback rating: ${req.body['feedback-rating']}
 ----
 ${req.body['feedback-suggestion']}`
 
   const opts = {
     email: req.body.email,
     name: '(no name supplied)',
     type: 'question',
-    subject: `Feedback from: ${req.body['service-name']}`,
+    subject: `General feedback`,
```
kbottla, line 17: Shall we say we `Feedback from service` to make it clear for support?

app/views/feedback/index.njk
```diff
@@ -4,6 +4,16 @@
 Give feedback — GOV.UK Pay
 {% endblock %}
 
+{% block beforeContent %}
+  {{ super() }}
+  {{
+    govukBackLink({
+      text: "Back to my services",
```
kbottla, line 11: Can we change this to "My services" as users can visit page from any page

kbottla (commented):
Changes looks good to me. Few minor suggestions on naming

app/views/feedback/index.njk
iqbalgds: Good point

kbottla (approved):

app/controllers/feedback/post-index.controller.js
iqbalgds: Done

kbottla (approved):

PR #2548, opened 2021-02-02
PP-7694 Move request to go live pages to service level

stephencdaly:
The URLs for these pages were at the service level, but they were displayed with the account navigation.

As we are going to remove the cookie for determining the current gateway account, these pages will no longer know which account the user has come from.

Remove the `getAccount` middleware from these routes and stop showing the account navigation. Continue displaying the service name header and include a "My services" back link in the top left corner.  Remove some top padding on the index page as this space is now filled by the back link.

Screenshot 2021-02-02 at 16 05 33

Replace the link on the confirmation page with a "Go to My services" link.

Screenshot 2021-02-02 at 16 12 14


kbottla (approved):
LGTM!

PR #2549, opened 2021-02-03
PP-7445 Feedback form - refactor Zendesk call

iqbalgds:
with @SandorArpa

- Refactor the Zendesk call to create a ticket.
  - Use Aysnc/await when calling the 'zendesk.client.js' util.
- Refactor 'zendesk.client.js'
  - Remove promise wrapper and just 'return' the output of the
    'zendeskClient.tickets.create'

### DIFF
--- a/app/controllers/feedback/post-index.controller.js
+++ b/app/controllers/feedback/post-index.controller.js
@@ -5,7 +5,7 @@ const logger = require('../../utils/logger')(__filename)
 const paths = require('../../paths')
 const zendeskClient = require('../../services/clients/zendesk.client')
 
-module.exports = (req, res) => {
+module.exports = async function postZendeskFeedback (req, res) {
   const message = `Feedback rating: ${req.body['feedback-rating']}
 ----
 ${req.body['feedback-suggestion']}`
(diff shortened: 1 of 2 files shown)


app/services/clients/zendesk.client.js
```diff
@@ -11,33 +10,26 @@ const zendeskClient = zendesk.createClient({
   proxy: process.env.http_proxy
 })
 
-module.exports = {
+function createTicket (opts) {
+  zendeskClient.tickets.create({
```
sfount, line 14: I believe this promise will need to be returned to the caller of the function for the async/ await error handling above to correctly propagate 
```suggestion
  return zendeskClient.tickets.create({
```

app/controllers/feedback/post-index.controller.js
```diff
@@ -14,23 +14,20 @@ ${req.body['feedback-suggestion']}`
     email: req.body.email,
     name: '(no name supplied)',
     type: 'question',
-    subject: `Feedback from service`,
+    subject: `Feedback from: ${req.body['service-name']}`,
```
kbottla, line 17: Do we ever have `service-name` form field available as we don't use service/account middleware anymore ?

app/services/clients/zendesk.client.js
iqbalgds: Adding `return` did not work.
So we reverted the `zendesk.client.js` back to how it was.

app/controllers/feedback/post-index.controller.js
iqbalgds: Thanks Krishna - made a mistake when updating the code - this is now `Feedback from service`

iqbalgds:
`zendesk.client.js` - could not remove the `promise` wrapper. The `return zendeskClient.tickets.create` does not work.
app/controllers/feedback/post-index.controller.js - Update to use Async/Await.

kbottla (approved):
LGTM!

app/services/clients/zendesk.client.js
iqbalgds: Actually it does work.  I have updated the code.
Thanks

app/controllers/feedback/post-index.controller.js
```diff
@@ -19,18 +19,15 @@ ${req.body['feedback-suggestion']}`
...
-      return res.redirect(paths.feedback)
-    })
+  try {
+    await zendeskClient.createTicket(opts)
+
+    lodash.set(req, 'session.pageData.feedback', {})
+    req.flash('generic', 'Thanks for your feedback')
+    return res.redirect(paths.feedback)
+  } catch (err) {
+    logger.error(`Error posting request to Zendesk - ${err}`)
```
sfount, line 29: This leaves unpacking the meaningful bit of the error to JavaScript which can mean getting some `[object Object]` fun in the logs. Recommend logging the errors message or the full error as context to the log
```suggestion
    logger.error(`Error posting request to Zendesk - ${err && err.message}`)
```
```suggestion
    logger.error('Error posting request to Zendesk', err)
```

sfount (approved):
Looks good, code reads very clearly 👍  comment about the error logging, it probably doesn't need to hold this pr up

app/controllers/feedback/post-index.controller.js
iqbalgds: Thanks - I will make another PR for this change, I like the second option so will do that.

PR #2550, opened 2021-02-03
BAU Remove 3DS flex warning

sfount:
Now that credentials are verified before being saved in the backend
there is no need to warn users about this implication up front.

Also @stevenjmesser told me to.


stevenjmesser (approved):
Looks great to me!

alexbishop1 (approved):

PR #2554, opened 2021-02-04
PP-7445 Feedback routes - Update Zendesk error handling.

iqbalgds:
- Update log message for Zendesk errors to log the full error context

## WHAT

## HOW 
_Steps to test or reproduce:_


app/controllers/feedback/post-index.controller.js
```diff
@@ -26,7 +26,7 @@ ${req.body['feedback-suggestion']}`
     req.flash('generic', 'Thanks for your feedback')
     return res.redirect(paths.feedback)
   } catch (err) {
-    logger.error(`Error posting request to Zendesk - ${err}`)
+    logger.error('Error posting request to Zendesk', err)
```
stephencdaly, line 29: I think this will put the object in the structured arguments of the log message, with the fields of the error object as top-level fields on the log line. 

I think to make this clearer, it should be inside an `error` field on the log line. So
```logger.error('Error posting request to Zendesk', { error: err })```

stephencdaly (approved):

PR #2555, opened 2021-02-04
BAU Attempt to fix pact port in use errors

stephencdaly:
Selfservice tests fail regularly with the error:
```Error: Port 'XXXXX' is already in use by another process```

The pact-node provider will by default use an available random port if no port is specified
https://www.npmjs.com/package/@pact-foundation/pact-node

Remove our code to get a random port without checking availability and use the assigned port in the URL used by the client in the pact test.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json",
     "lines": null
   },
-  "generated_at": "2021-02-04T12:30:01Z",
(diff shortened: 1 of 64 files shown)


alexbishop1 (commented):
Is `productsApiKey` not necessary?

stephencdaly:
> Is `productsApiKey` not necessary?

It wasn't used by the function that accepted it as a parameter

alexbishop1 (approved):

PR #2557, opened 2021-02-05
PP-7531 Add root controller

kbottla:
## WHAT
- Added new root (for path `/`) controller which redirects users to account dashboard when `ENABLE_MY_SERVICES_AS_DEFAULT_VIEW` flag is not enabled and to My service page when the flag is enabled

### DIFF
--- a/app/controllers/root/index.controller.js
+++ b/app/controllers/root/index.controller.js
@@ -1,0 +1,16 @@
+'use strict'
+
+const paths = require('../../paths')
+const formatAccountPathsFor = require('../../utils/format-account-paths-for')
+
+function get (req, res) {
+  if (process.env.ENABLE_MY_SERVICES_AS_DEFAULT_VIEW === 'true') {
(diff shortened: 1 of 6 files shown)


app/controllers/root/index.controller.js
```diff
@@ -0,0 +1,16 @@
+'use strict'
+
+const paths = require('../../paths')
+const formatAccountPathsFor = require('../../utils/format-account-paths-for')
+
+const get = function index (req, res) {
```
stephencdaly, line 6: The const name is different to the function name

I think we decided
```
function get (req, res) {
...
}
```
without assigning to a constant was best to avoid this.

stephencdaly (commented):
Looks good just a minor thing about the function declaration

app/controllers/root/index.controller.js
kbottla: Updated!

stephencdaly (approved):

PR #2564, opened 2021-02-10
PP-7701 Show notification on my services page

stephencdaly:
- Show a notification banner when users land on the My services page to tell them about the behavioural change to have this as the default page.

- When a "Hide message" button is clicked, set a cookie and hide the banner.
- On subsequent page loads, check the cookie to decide whether to show the notification or not.

- This works for both javascript enabled and not enabled. For javascript enabled the banner is hidden and the cookie is set by client-side javascript. If the javascript is not enabled the form inside the banner is submitted to a route that sets the cookie then redirects to re-display the my services page.

- Have not added environment variable check on date to show banner until yet.


test/cypress/integration/my-services/my-services-notification-banner.cy.test.js
```diff
@@ -0,0 +1,29 @@
+'use strict'
+
+const gatewayAccountStubs = require('../../stubs/gateway-account-stubs')
+const userStubs = require('../../stubs/user-stubs')
+
+const userExternalId = 'authenticated-user-id'
+
+describe('Service has a live account that supports payouts', () => {
```
kbottla, line 8:
```suggestion
describe('My services notification banner or something ?', () => {
```

stephencdaly: good spot - copy and paste we didn't change

stephencdaly: updated

kbottla (approved):
Looks good!! 🎉

PR #2567, opened 2021-02-12
PP-7701 My Services Default page > Notificatons banner - fix analytics bug

iqbalgds:
- Update the analytics call when the `hide message` button is called
  - Update to use `ga` rather then the older `_gaq`


SandorArpa (approved):
LOKTM

PR #2568, opened 2021-02-12
PP-7592 add onward journey link to email error

SandorArpa:
## WHAT
- when a user is on the sign up page, and enters a valid email address, but it is not on the approved list, they are shown the link to contact support
- replaced 'govukInput' with custom code. govukInput.errorMessage.text does not render html markups

### DIFF
--- a/app/views/self-create-service/register.njk
+++ b/app/views/self-create-service/register.njk
@@ -18,23 +18,22 @@
   <h1 class="govuk-heading-l page-title">Create an account</h1>
   <form action="{{routes.selfCreateService.register}}" method="post" id="submit-service-creation" class="form submit-registration">
     <input id="csrf" name="csrfToken" type="hidden" value="{{csrf}}"/>
-    {{
-      govukInput({
-        id: "email",
-        name: "email",
-        type: "email",
-        classes: "govuk-input--width-20",
-        label: {
-            text: "Email",
-            classes: "govuk-label--s"
-        },
-        value: email,
-        hint: {
-          text: "Must be from a government organisation"
-        },
-        errorMessage: { text: errors.email } if errors.email else false
-      })
-    }}
+    <div class="govuk-form-group {% if errors['email'] %} govuk-form-group--error {% endif %}">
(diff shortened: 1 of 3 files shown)


app/views/self-create-service/register.njk
```diff
@@ -18,23 +18,22 @@
...
+    <div class="govuk-form-group {% if errors['email'] %} govuk-form-group--error {% endif %}">
+        <label class="govuk-label govuk-label--s" for="email">
+            Email
+        </label>
+        <div id="email-hint" class="govuk-hint">
+            <p class="govuk-hint">Must be from a government organisation</p>
+        </div>
+        {% if errors['email'] %}
+            <span id="email-error" class="govuk-error-message">
+                <span class="govuk-visually-hidden">Error:</span> {{errors["email"]}}. If you cannot create an account using your public sector email address, please <a href="">contact support</a>.
```
iqbalgds, line 30: The href is empty.
I think we need something like this in there: mailto:support

test/ui/self-create-service.ui.test.js
```diff
@@ -96,4 +96,26 @@ describe('Self-create service view', () => {
...
+      errors: { email: 'Enter a public sector email address' }
+    }
+
+    const body = renderTemplate('self-create-service/register', templateData)
+
+    body.should.containSelector('h1').withExactText('Create an account')
+
+    body.should.containSelector('form#submit-service-creation').withAttribute('action', paths.selfCreateService.register)
+    body.should.containInputField('email', 'email')
+    body.should.containSelector('#email-error').withText('Error: Enter a public sector email address. If you cannot create an account using your public sector email address, please contact support.')
```
iqbalgds, line 115: If possible, can we update the test to check for the email?

app/views/self-create-service/register.njk
SandorArpa: 🤦

sfount: There's also this page https://www.payments.service.gov.uk/support/

I don't know if the story specified

iqbalgds (approved):

PR #2569, opened 2021-02-12
PP-7708 List team members with take telephone payments roles

alexbishop1:
On the page for managing team members, list users who have the `view-and-initiate-moto` or `view-refund-and-initiate-moto` roles.

It’s not currently possible to assign these roles to users so there should not be any yet.

This contains a temporary change to the `getRoleByExtId` function to hide the new roles from the code for managing team members and inviting new users so someone cannot sneakily assign these new roles to users before we’re ready.


app/controllers/service-users.controller.js
```diff
@@ -60,25 +60,14 @@ module.exports = {
 
     const onSuccess = function ([members, invitedMembers]) {
       const teamMembers = mapByRoles(members, externalServiceId, req.user)
-      const numberOfAdminMembers = teamMembers.admin.length
-      const numberOfViewOnlyMembers = teamMembers[roles['view-only'].name].length
-      const numberOfViewAndRefundMembers = teamMembers[roles['view-and-refund'].name].length
-      const numberActiveMembers = numberOfAdminMembers + numberOfViewOnlyMembers + numberOfViewAndRefundMembers
```
alexbishop1, line 66: These were, like, never used in the template.

app/views/team-members/team-members.njk
```diff
@@ -118,6 +118,62 @@ Team members - GOV.UK Pay
...
+                  <a class="govuk-link govuk-link--no-visited-state govuk-!-display-block " id="view-and-initiate-moto-member-{{ loop.index }}-link" href="{{ team_member.link }}">
+                {% else %}
+                  <span>
+                {% endif %}
+                {{ team_member.username }} {% if team_member.is_current %}(you){% endif %}
+                {% if permissions.users_service_read %}
+                  </a>
+                {% else %}
+                  </span>
+                {% endif %}
```
sfount, line 141: Conditional closing tags frightening me a bit! 

Could the condition be lifted for readability?

(suggestion not tested!)

```suggestion
                {% set permissionText %}
                  {{ team_member.username }} {% if team_member.is_current %}(you){% endif %}
                {% endset %}
                {% if permissions.users_service_read %}
                  <a href="{{ team_member.link }}">{{ permissionText }}</a>
                {% else %}
                  <span>{{ permissionText }}</span>
                {% endif %}
```

sfount (commented):
I haven't looked at the context of the PR or functionality as a whole but a quick comment on a potentially fragile point in the templating. It probably shouldn't block the PR

app/views/team-members/team-members.njk
alexbishop1: That entire block is just a lightly-modified block copied and pasted from another part of the page but you’re right the logic is hideous so I’ll have a look at fixing it.

alexbishop1: OK, done that.

app/utils/roles.js
```diff
@@ -13,7 +15,7 @@ module.exports = {
   getRoleByExtId: roleExtId => {
     let found
     _.toArray(roles).forEach(role => {
-      if (role.extId === roleExtId) {
+      if (role.extId === roleExtId && role.extId !== 500 && role.extId !== 600) {
```
sfount, line 18: I don't think this is a good precedent to set, it's not clear from the code what 500 or 600 are or why they'd be excluded. 

If this pattern had to exist I would recommend naming this and separating the inline logic so that it's explicit, something along the lines of

```js 
const nonPermissableRoleExternalIds = [ 500, 600 ]
if (nonPermissableRoleExternalIds.includes(role.extId)) {
  return
}
```

This whole function and patten is confusing to be honest

sfount (approved):
Really good template cleanup 👍  I haven't verified the functionality works as expected but it follows and improves on the pattern

app/utils/roles.js
alexbishop1: This will be gone soon, hopefully in a few hours.

PR #2576, opened 2021-02-16
PP-7708 Allow assigning existing users to telephone payments roles

alexbishop1:
Allow admin users of a service to change an existing user’s role to `view-and-initiate-moto` or `view-refund-and-initiate-moto`.

For now, these two new roles are only shown if the service has the `agent_initiated_moto_enabled` flag set, which we’re using to limit which services have the agent-initiated MOTO payments feature.

If the flag is false, a determined admin could still assign one of the new roles to a user. However, on its own that will not allow
the user to to do anything more, so it’s not really worth trying to prevent this.

### DIFF
--- a/app/views/team-members/team-member-permissions.njk
+++ b/app/views/team-members/team-member-permissions.njk
@@ -30,62 +30,154 @@ Edit team member permissions - GOV.UK Pay
   </dl>
   <form id="role-update-form" method="post" action="{{editPermissionsLink}}">
     <input id="csrf" name="csrfToken" type="hidden" value="{{csrf}}"/>
-    {{
-      govukRadios({
-        idPrefix: 'role-input',
-        name: 'role-input',
-        fieldset: {
-          legend: {
-            text: 'Permission level',
-            classes: 'govuk-fieldset__legend--s'
-          }
-        },
-        items: [
-          {
-              value: admin.id,
-              id: "role-admin-input",
-              text: 'Administrator',
-              classes: 'govuk-label--s',
-              checked: admin.checked,
-              hint: {
-                html: '<ul class="govuk-list pay-text-grey">
-                        <li>View transactions</li>
-                        <li>Refund payments</li>
-                        <li>Manage settings</li>
-                      </ul>'
-              }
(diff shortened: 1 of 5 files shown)


SandorArpa (approved):
LGTM

app/views/team-members/team-member-permissions.njk
```diff
@@ -30,62 +30,154 @@ Edit team member permissions - GOV.UK Pay
...
-              classes: 'govuk-label--s',
-              checked: admin.checked,
-              hint: {
-                html: '<ul class="govuk-list pay-text-grey">
-                        <li>View transactions</li>
-                        <li>Refund payments</li>
-                        <li>Manage settings</li>
-                      </ul>'
-              }
+    {% if serviceHasAgentInitiatedMotoEnabled %}
```
sfount, line 33: I think the duplication here can be avoided, as the entries seem to have exactly the same content it feels prime for getting out of sync with each other.

It might work something like

```njk
{% set roleOptionItems = [{
    value: admin.id,	
    id: "role-admin-input"
  }, {
    // ...
  }] 
%}
 
{% if serviceHasAgentInitiatedMotoEnabled %}

{% set roleOptionItems = roleOptionItems.push({
      value: viewAndInitiateMoto.id,
      id: "role-view-and-initiate-moto-input"
    }, {
       value: viewRefundAndInitiateMoto.id,
       id: "role-view-refund-and-initiate-moto-input"
    }) 
%}

{% endif %}

{% govukRadios({
  items: roleOptionItems
}) %}
```

This solution hasn't actually been tried, the syntax may be whack.

alexbishop1: They’re not quite the same: they have slightly different hints. But I’ll see what I can do.

sfount: Ah I didn't notice the extra hint implicit with the flag being set. This doesn't need to block the pr

alexbishop1: I think I will leave it for now. While adding extra roles to the array is fairly straightforward, also going back and modifying all the hints would be a bit gnarly!

PR #2577, opened 2021-02-17
PP-7740 Add request Stripe test account page

kbottla:
## WHAT
- Added new controller (GET) to request Stripe test account and relevant tests

with @iqbalgds

### DIFF
--- a/app/controllers/request-psp-test-account/get.controller.js
+++ b/app/controllers/request-psp-test-account/get.controller.js
@@ -1,0 +1,21 @@
+'use strict'
+
+const { response } = require('../../utils/response')
+const { NOT_STARTED, CREATED, REQUEST_SUBMITTED } = require('../../models/psp-test-account-stage')
+const goLiveStage = require('../../models/go-live-stage')
+
+module.exports = function getRequestPspTestAccount (req, res, next) {
+  const service = req.service
+  try {
+    const pageData = {
+      requestForPspTestAccountNotStarted: (service.currentPspTestAccountStage === NOT_STARTED || !service.currentPspTestAccountStage) && service.currentGoLiveStage !== goLiveStage.LIVE,
+      requestForPspTestAccountSubmitted: service.currentPspTestAccountStage === REQUEST_SUBMITTED,
+      pspTestAccountCreated: service.currentPspTestAccountStage === CREATED,
+      isServiceLive: service.currentGoLiveStage === goLiveStage.LIVE
+    }
+
+    return response(req, res, 'request-psp-test-account/index', pageData)
+  } catch (error) {
+    return next(error)
+  }
+}
--- a/app/views/request-psp-test-account/index.njk
+++ b/app/views/request-psp-test-account/index.njk
@@ -1,0 +1,73 @@
+{% extends "../layout.njk" %}
+
+{% block pageTitle %}
+  Request Stripe test account - {{ currentService.name }} - GOV.UK Pay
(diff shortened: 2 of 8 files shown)


app/controllers/request-psp-test-account/get.controller.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
+
+const { response, renderErrorView } = require('../../utils/response')
+const { NOT_STARTED, CREATED, REQUEST_SUBMITTED } = require('../../models/psp-test-account-stage')
+const goLiveStage = require('../../models/go-live-stage')
+
+module.exports = async (req, res) => {
+  const service = req.service
+  service.currentPspTestAccountStage = CREATED
```
sfount, line 9: Does this intentionally set the `currentPspTestAccountStage` and then do logical comparisons on it later? Assuming this value can change should this only be setting it if there isn't already an existing value?

app/controllers/request-psp-test-account/get.controller.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
+
+const { response, renderErrorView } = require('../../utils/response')
+const { NOT_STARTED, CREATED, REQUEST_SUBMITTED } = require('../../models/psp-test-account-stage')
+const goLiveStage = require('../../models/go-live-stage')
+
+module.exports = async (req, res) => {
```
sfount, line 7: Named function would be useful here

app/controllers/request-psp-test-account/get.controller.js
```diff
@@ -0,0 +1,22 @@
...
+    const pageData = {
+      requestForPspTestAccountNotStarted: (service.currentPspTestAccountStage === NOT_STARTED && service.currentGoLiveStage !== goLiveStage.LIVE),
+      requestForPspTestAccountSubmitted: service.currentPspTestAccountStage === REQUEST_SUBMITTED,
+      pspTestAccountCreated: service.currentPspTestAccountStage === CREATED,
+      isServiceLive: service.currentGoLiveStage === goLiveStage.LIVE
+    }
+
+    return response(req, res, 'request-psp-test-account/index', pageData)
+  } catch (error) {
+    return renderErrorView(req, res, false, error.errorCode)
```
sfount, line 20: As this error handler doesn't do anything custom it should favour `next(error)`, the error will be handled in the same way but with consistent logging standards

(also wondering what error could actually happen here! I don't think anything above should throw, but not going to get in the way of more cautious rather than less)

app/views/request-psp-test-account/index.njk
```diff
@@ -0,0 +1,73 @@
...
+
+{% block mainContent %}
+  <div class="govuk-grid-column-two-thirds">
+
+  {% if isServiceLive %}
+    <h1 class="govuk-heading-l">
+      Test account cannot be requested
+    <h1>
+    <p class="govuk-body">
+      Stripe test account cannot be requested for live service. You can create another service and request Stripe test account
```
sfount, line 25: I assume the content here is not finalised? A number of grammatical issues I think 
```suggestion
      Stripe test accounts cannot be requested for a live service. You can create another service and request a Stripe test account
```

app/controllers/request-psp-test-account/get.controller.js
kbottla: 🤦🏽 This should haven't been in.

sfount (approved):

PR #2581, opened 2021-02-17
PP-7708 Allow inviting new users to telephone payments roles

alexbishop1:
Allow admin users of a service to invite a user granting them the `view-and-initiate-moto` or `view-refund-and-initiate-moto` role.

For now, these two new roles are only shown if the service has the `agent_initiated_moto_enabled` flag set, which we’re using to limit which services have the agent-initiated MOTO payments feature.

If the flag is false, a determined admin could still invite a user and grant them one of the new roles. However, on its own that will not allow the user to to do anything more, so it’s not really worth trying to prevent this.

### DIFF
--- a/app/views/team-members/team-member-invite.njk
+++ b/app/views/team-members/team-member-invite.njk
@@ -25,60 +25,144 @@ Invite a new team member - GOV.UK Pay
   <form id="invite-member-form" method="post" action="{{teamMemberInviteSubmitLink}}" class="govuk-grid-column-two-thirds">
     <input id="csrf" name="csrfToken" type="hidden" value="{{csrf}}"/>
     {{ govukInput({
-        label: {
-          text: "Email address"
-        },
-        id: "invitee-email",
-        name: "invitee-email",
-        classes: "govuk-!-width-two-thirds",
-        type: "email"
-      })
-    }}
-
-    {{ govukRadios({
-      idPrefix: "role-input",
-      name: "role-input",
-      fieldset: {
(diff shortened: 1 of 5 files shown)


oswaldquek:
Is it possible to put a screenshot of team-member-invite.njk here?

alexbishop1:
Yeah, sorry about that! The diff makes the changes look way bigger than they actually are.

Here’s a screenshot:

image

If the service doesn’t have the flag set, the bottom two options aren’t there and none of the hints mention telephone payments (so what it looks like currently).

app/views/team-members/team-member-invite.njk
```diff
@@ -24,61 +24,158 @@ Invite a new team member - GOV.UK Pay
...
+              Cannot manage settings"
+            }
           }
-        },
-        {
-          value: view.id,
-          text: "View only",
+        ]
+      }) }}
+    {% else %}
```
oswaldquek, line 117: Is it possible to extract everything under this `else` block, seeing as its common to both cases (where service has agent initiated moto enabled and where it hasn't)?

alexbishop1: Not everything because the hints are different for the different roles depending on whether the service has the flag or not. But there’s no need for the email field code to be duplicated so I’ll fix that.

alexbishop1: OK, done that.

oswaldquek (approved):

PR #2582, opened 2021-02-18
PP-7709 Filter products by type in products not selfservice

alexbishop1:
When we need all the payment links for a gateway account or all the prototype links for a gateway account, use the new endpoint on the products microservice that filters by type.

This is more efficient than retrieving every live product for a gateway account from the database, dragging them all across the network and then filtering out the ones we do not want.


rorymalcolm (approved):
LGTM

PR #2584, opened 2021-02-19
PP-7785 All service transactions status filter

sfount:
Currently the all users accounts permissions model is based on filtering
only live accounts.

Generalise this to allow the calling code to specify if it should return
live or test accounts.

Updates the tests and add coverage specifically for filtering test
accounts.

For the all service transactions page, indicate if the current status filter is live or test with a tag next to
the title as in the design proposal.

Add a link to switch between live and test environments with the
following logic:
* viewing test data and has live accounts - show switch to live
* viewing test data with no live accounts - nothing
* viewing live data - show switch to test

Screenshot 2021-02-19 at 12 37 27

Screenshot 2021-02-19 at 12 37 50

Screenshot 2021-02-19 at 12 38 00


app/controllers/all-service-transactions/get.controller.js
```diff
@@ -20,26 +20,34 @@ const { CORRELATION_HEADER } = require('../../utils/correlation-header.js')
...
+    logger.info('Listing all services transactions', logContext)
 
     if (!userPermittedAccountsSummary.gatewayAccountIds.length) {
-      return next(new NoServicesWithPermissionError('You do not have any associated services with rights to view live transactions.'))
+      return next(new NoServicesWithPermissionError('You do not have any associated services with rights to view these transactions.'))
     }
     const searchResultOutput = await transactionService.search(userPermittedAccountsSummary.gatewayAccountIds, filters.result)
     const cardTypes = await client.getAllCardTypes(correlationId)
-    const model = buildPaymentList(searchResultOutput, cardTypes, null, filters.result, router.paths.allServiceTransactions.download, req.session.backPath)
+    const route = filterLiveAccounts ? paths.allServiceTransactions.download : paths.formattedPathFor(paths.allServiceTransactions.downloadStatusFilter, 'test')
```
kbottla, line 47: (minor) can we say `downloadRoute` to make it clear what it is for ?

app/paths.js
```diff
@@ -128,7 +128,9 @@ module.exports = {
   },
   allServiceTransactions: {
     index: '/all-service-transactions',
+    indexStatusFilter: '/all-service-transactions/:statusFilter(test|live)',
```
kbottla, line 131: Nice 👏🏽

app/views/transaction-detail/index.njk
```diff
@@ -12,7 +12,7 @@
   {{ super() }}
   {% set defaultBackLink %}{{formatAccountPathsFor(routes.account.transactions.index, currentGatewayAccount.external_id)}}{% if indexFilters %}?{{indexFilters}}{% endif %}{% endset %}
```
kbottla, line 13: Should this also factor in all service transactions test URL ?

app/views/transactions/index.njk
```diff
@@ -41,11 +41,23 @@
...
        Transactions
     {% endif %}
   </h1>
+
+  {% if allServiceTransactions %}
+    {% if filterLiveAccounts %}
+      <p><a class="govuk-link govuk-link--no-visited-state" href="{{ routes.formattedPathFor(routes.allServiceTransactions.indexStatusFilter, 'test') }}">Switch to test</a></p>
+    {% else %}
+      {% if hasLiveAccounts %}
+        <p><a class="govuk-link govuk-link--no-visited-state" href="{{ routes.allServiceTransactions.index }}">Switch to live</a></p>
```
kbottla, line 56: This will be available immediately (without setting ENABLE_TEST_REPORTS). There will be small window that users see `switch to test` and directed to old instance. Fine as it is too as it is very unlikely lot of users see this or access test transactions

app/views/transactions/index.njk
```diff
@@ -41,11 +41,23 @@
...
+       Transactions for all services
+       <strong class="govuk-tag govuk-tag--grey">{{ "LIVE" if filterLiveAccounts else "TEST" }}</strong>
     {% else %}
        Transactions
     {% endif %}
   </h1>
+
+  {% if allServiceTransactions %}
+    {% if filterLiveAccounts %}
+      <p><a class="govuk-link govuk-link--no-visited-state" href="{{ routes.formattedPathFor(routes.allServiceTransactions.indexStatusFilter, 'test') }}">Switch to test</a></p>
```
kbottla, line 53: Can we say `Switch to test accounts`? Content updated (final now) this afternoon.

Similarly `Switch to live accounts` below

kbottla (commented):
Looks great! Just added few minor comments

app/views/transaction-detail/index.njk
sfount: I don't think this has anything to do with this PR

app/views/transactions/index.njk
sfount: Yeah it seemed like the only users that would see this are already familiar with this tool so the fact it works for the test environment too shouldn't throw them, a good point though and should be fairly easy to change if we need to

app/controllers/all-service-transactions/get.controller.js
sfount: Yeah good suggestion I think, it was called route as that's the param name for the code that then uses it but this should help clarify in context

kbottla (approved):

PR #2585, opened 2021-02-19
PP-7766 Add request Stripe test account link

kbottla:
## WHAT
- Added `Request Stripe test account` link to sandbox account dashboard.
- Available when the flag ENABLE_STRIPE_TEST_ACCOUNT_REQUEST is enabled.
- Also removed 'Manage team members' link on account dashboard (as per design prototype) as the link is available on Myservices page (which will be the default landing page soon) and doesn't belong to specific gateway account.


app/views/dashboard/_links.njk
```diff
@@ -1,4 +1,6 @@
-{% if linksToDisplay.length % 3 == 0 %}
+
+{% set noOfLinks = linksToDisplay.length + (1 if requestedStripeTestAccount else 0) %}
```
stephencdaly, line 2: Would this be slightly simpler if we sent the `requestPspTestAccount` in the linksToDisplay when test account is requested, but then checked `requestedStripeTestAccount` value to check what content to display later? This would be closer to how it works for the go live links.

We are displaying the "link", it just has different content and isn't a link.

test/cypress/integration/dashboard/dashboard-links.cy.test.js
```diff
@@ -67,27 +65,43 @@ describe('the links are displayed correctly on the dashboard', () => {
...
-      cy.get('#payment-links-link').should('have.class', 'flex-grid--column-third')
+      cy.get('#payment-links-link').should('have.class', 'flex-grid--column-half')
 
-      cy.get('#manage-service-link').should('exist')
-      cy.get('#manage-service-link').should('have.class', 'flex-grid--column-third')
+      cy.get('#request-to-go-live-link').should('exist')
+      cy.get('#request-to-go-live-link').should('have.class', 'flex-grid--column-half')
+    })
+
+    it('should display `Stripe test account requested` section if request has been submitted', () => {
```
stephencdaly, line 88: It might be good to check on the content of the link in this test rather than just ids?

stephencdaly (commented):
Some possible suggested improvements.

Does this PR also remove the manage team members link? Is it worth saying that in the commit message?

test/cypress/integration/dashboard/dashboard-links.cy.test.js
kbottla: We don't usually test content in cypress test. We could add templates test to test this ? I will raise separate PR once content is finalised

app/views/dashboard/_links.njk
kbottla: Thanks, updated!

test/cypress/integration/dashboard/dashboard-links.cy.test.js
stephencdaly: We don't test full content, but we tend to do assertions on elements/text rather than IDs usually as I reckon it gives more confidence that we're displaying the right thing.

stephencdaly (approved):

test/cypress/integration/dashboard/dashboard-links.cy.test.js
stephencdaly: Usually we also don't use IDs in templates if they're just for testing, we only use them if they're required for application code. We should be able to test most things we used to test in template tests in the Cypress tests as they give us a higher degree of confidence that the template and the server code work together.

PR #2586, opened 2021-02-19
PP-7786 Payments to your bank account status filter

sfount:
Depends on https://github.com/alphagov/pay-selfservice/pull/2584

Allow the payments to your bank account page to filter live or test
accounts based on the updated cross account permissions utility.

Add a tag to show what environment is being displayed following the
ticket design.

Update the download link for live and test environments.

Add a link to switch between live and test environments with the
following logic:
* viewing test data and has live accounts - show switch to live
* viewing test data with no live accounts - nothing
* viewing live data - show switch to test

Screenshot 2021-02-19 at 15 41 37

Screenshot 2021-02-19 at 15 41 47

### DIFF
--- a/app/views/payouts/list.njk
+++ b/app/views/payouts/list.njk
@@ -26,7 +26,9 @@ Payments to your bank account - GOV.UK Pay
   {% if enableTestReports %}
   <p class="govuk-body">
     {% if filterLiveAccounts %}
+      {% if hasTestStripeAccount %}
       <a class="govuk-link govuk-link--no-visited-state" href="{{ routes.formattedPathFor(routes.payouts.listStatusFilter, 'test') }}">Switch to test accounts</a>
+      {% endif %}
     {% else %}
       {% if hasLiveAccounts %}
         <a class="govuk-link govuk-link--no-visited-state" href="{{ routes.payouts.list }}">Switch to live accounts</a>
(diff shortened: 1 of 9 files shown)


app/views/payouts/list.njk
```diff
@@ -16,12 +16,27 @@ Payments to your bank account - GOV.UK Pay
...
+  </h1>
 </div>
-<div class="govuk-grid-column-two-thirds">
-  <p class="govuk-body govuk-!-margin-bottom-8">Payments Stripe has made to your bank account.</p>
+<div class="govuk-grid-column-two-thirds govuk-!-margin-bottom-8">
+  <p class="govuk-body">Payments Stripe has made to your bank account.</p>
+  {% if enableTestReports %}
+  <p class="govuk-body">
+    {% if filterLiveAccounts %}
+      <a class="govuk-link govuk-link--no-visited-state" href="{{ routes.formattedPathFor(routes.payouts.listStatusFilter, 'test') }}">Switch to test accounts</a>
```
kbottla, line 29: Can we only show `Switch to test accounts` if there are any test Stripe accounts ?

sfount: We could do this as a custom property further up in the utility that checks permissions, do we think it's something that's needed though? Potentially showing the link helps keep the tools predictable and symmetrical as a reminder to teams that you can test this stuff with Stripe test accounts (when the content is added)

kbottla: Displaying link could indicate service has access to test reports which will never be the case. In case, if service has already got live Stripe account, it is very unlikely service would want to setup a test account (given service would already know about additional functionality that comes with Stripe).

kbottla (approved):
LGTM!

PR #2601, opened 2021-02-23
PP-7789 Breadcrumbs navigation header

sfount:
Update the service/ account navigation "phase_banner" using the new
breadcrumb according to the ticket specification.

Move the account specific navigation down to the line below.

Update pages that currently have a manual link back to the my services
page as they can now rely on the breadcrumb for that functionality.

All changes are behind the future method feature flag.

Beginning to formalising page types
* All services page
* Service page
* Account page 
* Top level

Removes the "My services" link from the navigation header in favour of the breadcrumbs, behind the future feature flag.

**All services page header**

Screenshot 2021-02-23 at 12 46 46

**Account page header**

Screenshot 2021-02-23 at 12 47 04

**Service page header**

Screenshot 2021-02-23 at 12 47 24

**Top level**
Screenshot 2021-02-23 at 12 47 57

with @stephencdaly

### DIFF
--- a/app/views/payouts/list.njk
+++ b/app/views/payouts/list.njk
@@ -1,3 +1,4 @@
+{% from '../macro/breadcrumbs.njk' import breadcrumbs %}
 {% extends "../layout.njk" %}
 
 {% block pageTitle %}
@@ -5,13 +6,20 @@ Payments to your bank account - GOV.UK Pay
 {% endblock %}
 
 {% block beforeContent %}
-  {{ super() }}
+  {% if enabledMyServicesAsDefaultView %}
+
+  {{ breadcrumbs([
+    { text: "My services", href: routes.serviceSwitcher.index },
+    { text: "Payments to your bank account" }
+  ]) }}
+  {% else %}
   {{
     govukBackLink({
       text: "Back to My services",
       href: routes.serviceSwitcher.index
     })
   }}
+  {% endif %}
 {% endblock %}
 
 {% block mainContent %}
--- a/app/views/services/index.njk
+++ b/app/views/services/index.njk
@@ -9,6 +9,10 @@
   <script src="/public/js/components/notification-banner.js"></script>
 {% endblock %}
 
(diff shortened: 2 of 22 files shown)


app/views/services/index.njk
```diff
@@ -9,6 +9,10 @@
   <script src="/public/js/components/notification-banner.js"></script>
 {% endblock %}
 
+{% block beforeContent %}
```
kbottla, line 12: Is this supposed to have anything here or is it intended ?

app/views/payouts/list.njk
```diff
@@ -1,17 +1,25 @@
+{% from '../macro/breadcrumbs.njk' import breadcrumbs %}
 {% extends "../layout.njk" %}
 
 {% block pageTitle %}
 Payments to your bank account - GOV.UK Pay
 {% endblock %}
 
 {% block beforeContent %}
-  {{ super() }}
+  {% if enabledMyServicesAsDefaultView %}
```
kbottla, line 9: Shall we change this similar to all service transactions with live/test mode https://github.com/alphagov/pay-selfservice/blob/73d8712ab809bdc2b9125ef59d2e8d5fcc7b307e/app/views/transactions/index.njk#L31-L45

app/views/services/index.njk
sfount: It's intended, the extension mechanism in Nunjucks means that this template will by default use the `beforeContent` inherited from the main layout. As this page implements its own unique behaviour (the one page without any kind of header) it overrides this

app/views/payouts/list.njk
kbottla: As discussed, that can be a separate as changes are behind feature flag

kbottla (approved):

app/views/payouts/list.njk
sfount: 👍  great spot, I'll push the tag for breadcrumbs on payouts as separate PR

PR #2605, opened 2021-02-24
PP-7484 API Keys - Add API security text.

iqbalgds:
- Add new content to API keys index page.
- Add `API security` text to `create API flow`


SandorArpa (approved):
LOKTM

PR #2614, opened 2021-03-01
PP-7835 Enable all services test reports

kbottla:
## WHAT
- Removed `ENABLE_TEST_REPORTS` flag and usages which enables remaining functionality for all services test reports
- Removed `ENABLE_STRIPE_TEST_ACCOUNT_REQUEST` flag and usage as Stripe test requests functionality is fully enabled


app/views/services/index.njk
```diff
@@ -72,7 +72,7 @@
 
     {% set allServiceTransactionsPath = routes.allServiceTransactions.index if has_live_account else routes.formattedPathFor(routes.allServiceTransactions.indexStatusFilter, 'test') %}
 
-    {% if has_live_account or env.ENABLE_TEST_REPORTS === 'true' %}
+    {% if has_live_account %}
```
sfount, line 75: If I understand correctly I think this condition should be entirely removed?

The logic before read "If the flag is on, the new behaviour (always show)" OR "The existing behaviour (only show with live accounts). 

I think now it should read "The new behaviour (always show)"

sfount (changes requested):

sfount (approved):

PR #2616, opened 2021-03-02
PP-4339 Trim gateway credentials before sending to Connector

sfount:
When users have copied and pasted credentials from their payment service
providers they will sometimes come with hidden whitespace. This should
be removed before sending it to the backend as it can lead to invalid
credentials.


adityapahuja (approved):
Looks good to me!

PR #2629, opened 2021-03-08
PP-7859 Metadata Transactions filter - add UI elements

iqbalgds:
- Update controller - to pass down 'isExperimentalFeaturesEnabled' to view.
- Add 'Advance filters' toggle.
  - Only show this if the service has 'isExperimentalFeaturesEnabled' = true.
  - The toggle contains the metadata inbox filter.

### DIFF
--- a/app/views/transactions/filter.njk
+++ b/app/views/transactions/filter.njk
@@ -209,6 +209,37 @@
         </fieldset>
       </fieldset>
     </div>
+    {% if isExperimentalFeaturesEnabled %}
+      <div class="govuk-grid-column-full">
+        <div class="govuk-grid-column-one-half">
+          <details class="govuk-details" data-module="govuk-details">
+            <summary class="govuk-details__summary">
+              <span class="govuk-details__summary-text">
+                Advanced filters
+              </span>
+            </summary>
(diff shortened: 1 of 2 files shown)


app/views/transactions/filter.njk
```diff
@@ -209,6 +209,37 @@
...
+    {% if isExperimentalFeaturesEnabled %}
+      <div class="govuk-grid-column-full">
+        <div class="govuk-grid-column-one-half">
+          <details class="govuk-details" data-module="govuk-details">
+            <summary class="govuk-details__summary">
+              <span class="govuk-details__summary-text">
+                Advanced filters
+              </span>
+            </summary>
+            <div class="govuk-details__text">
```
kbottla, line 221: Detail is currently not expanded when search is performed on metadata field. I am assuming this is expected and to be added in a separate PR ?

kbottla (approved):
LGTM!

app/views/transactions/filter.njk
iqbalgds: Good spot - planning to do that in another PR.
Thanks.

PR #2630, opened 2021-03-08
PP-7859 Metadata Transactions filter - add UI elements to page

iqbalgds:
- Update controller - to pass down 'isExperimentalFeaturesEnabled' to view.
- Add 'Advance filters' toggle.
  - Only show this if the service has 'isExperimentalFeaturesEnabled' = true.
  - The toggle contains the metadata inbox filter.


kbottla:
Duplicate of https://github.com/alphagov/pay-selfservice/pull/2629

PR #2631, opened 2021-03-10
PP-7786 Payouts list handle no stripe test accounts

sfount:
Currently the live payouts page is restricted to only users that have
access to live Stripe accounts.

Update the test page to only show to users with test Stripe accounts.

Add Cypress coverage.


SandorArpa (approved):
LOKTM

PR #2633, opened 2021-03-11
PP-7859 Pass metadata_value to Ledger for transaction search

kbottla:
## WHAT
- Pass metadataValue search param & value to Ledger
- Rename query `metadata` to `metadataValue` as per the story and naming is inline with other query params
- Added pact to search transactions by metadata value


SandorArpa (approved):
LOKTM

PR #2641, opened 2021-03-15
Pp 7859 metadata filter controller updates

iqbalgds:
- Add the 'Advanced filters' to 'All transactions' view.
- Keep the 'Advanced filters' toggle open if a user has already submitted
  a metadata filter search.
- Update the controller, to pass the metadata filter values to the base client.
- Update the `download transactions` link to include query param for metadata filter.
- Update Cypress test for `transaction search` to include testing by metadata.


kbottla (approved):
LGTM!

PR #2642, opened 2021-03-16
PP-7702 Update tests when 'ENABLE_MY_SERVICES_AS_DEFAULT_VIEW=true'

iqbalgds:
- Update the test.env - make the flag true.
- Update the single failing test.


SandorArpa (approved):

PR #2643, opened 2021-03-16
BAU Fix double line below My services breadcrumb

stephencdaly:
There was an additional `govuk-phase-banner` element below the banner
containing the "My services" link on the edit phone number and change
second factor method pages under "My profile".

Add these views to the list of views that don't display the account
navigation so this doesn't happen.

This is what we were seeing:

Screenshot 2021-03-16 at 17 17 56


alexbishop1 (approved):
💯

PR #2651, opened 2021-03-23
PP-7893 Use new designs for API key pages

stephencdaly:
Create a separate template `revoked-keys.njk` for the revoked keys page as it has significantly diverged now. Continue using `_keys.njk` sub template but move content specific for active keys into `index.njk`.

Modify tests to be integration tests rather than functional tests so they're faster and easier to debug!

Screenshot 2021-03-23 at 17 50 50
Screenshot 2021-03-23 at 17 50 57
Screenshot 2021-03-23 at 17 51 13
Screenshot 2021-03-23 at 17 51 18


kbottla (approved):
LGTM!

PR #2652, opened 2021-03-24
PP-7733 Remove my services notification banner

stephencdaly:
We no longer show the notification banner as we only showed it until a date specified by an environment variable that has now passed.

Remove all code related to it.


SandorArpa (approved):
LOKTM

PR #2653, opened 2021-03-24
PP-7893 Add back id used by endtoend tests

stephencdaly:
The id on the header is used by end to end tests to check which page it's on. It's not a nice way of doing it but add it back in to avoid the hassle of editing the end to end tests.


SandorArpa (approved):
LOKTM

PR #2654, opened 2021-03-24
PP-7733 Remove old navigation from templates

stephencdaly:
Remove checks on the `ENABLE_MY_SERVICES_AS_DEFAULT_VIEW` feature flag in templates which decide whether to display the new or old navigation and always display the new navigation.

Remove `transaction-list.ui.test.js` as everything this tests is now covered by Cypress tests.


kbottla (approved):
Nice one 🎉 🎉 !

PR #2656, opened 2021-03-25
PP-7733 Remove feature flag

stephencdaly:
Remove all usages of the `ENABLE_MY_SERVICES_AS_DEFAULT_VIEW` feature flag and always default to "My services" page rather than account dashboard.


SandorArpa (approved):
LOKTM

PR #2657, opened 2021-03-25
PP-7733 Remove unnecessary middleware

stephencdaly:
Remove the getAccount, hasServices and resolveService middleware on all
routes that no longer require it.

Replace usages of the `enforceUserAuthenticated` middleware with the new
`is-user-authorised` middleware which does all the same checks.

Change tests that expect a redirect to `/otp-login` if the user has not
completed second factor authentication. This was never actually the
behaviour, as there was always a check on the session version first
which is updated after second factor authentication was completed, and
if the session version didn't match we redirected to `/login`. The
new `is-user-authorised` middleware now just always directs to `/login`
so this ambiguity is removed without changing actual behaviour.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json",
     "lines": null
   },
-  "generated_at": "2021-03-24T11:54:45Z",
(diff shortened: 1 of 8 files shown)


kbottla (approved):
LGTM!

PR #2658, opened 2021-03-26
PP-7715 Fix redirect after refund

stephencdaly:
Fix the redirect after making a refund if the user has come from the
"Transactions for all services" page, rather than the transactions list
for a particular account.

Do this by adding a hidden input to the refund form to indicate that we
have come from the all services transactions page which is passed along
with the redirect.

Store the filters for the all services transaction search on the session
as we do for the account transaction search so we can construct the back
link from these, even after redirecting following a refund. Also include
the status filter (live/test) to ensure we get back to the correct
search.

### DIFF
--- a/app/controllers/all-service-transactions/get.controller.js
+++ b/app/controllers/all-service-transactions/get.controller.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const _ = require('lodash')
+const url = require('url')
 
 const { response } = require('../../utils/response')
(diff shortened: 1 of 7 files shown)


sfount (approved):
The methodology of recreating the filters/ URL based on the what's in the session looks like it may get confusing, especially with the all services/ individual service possibilities.

It may be worth reviewing this in the future to see if there's a way to simplify that but this looks like a great fix. It follows the current pattern and is well covered.

PR #2667, opened 2021-03-30
PP-7735 Make service URLs uniform

stephencdaly:
Add new URLs for the organisation details page that use the pattern
`/service/:externalServiceId` so that we can use a Router for all
service routes.

Also change the URL for view/edit team members to be
`/service/:externalServiceId/team-members` rather than
`/service/:externalServiceId/` as this isn't really a root URL for a
service.

Keep routes for the old URLs which will be removed after this change
has been deployed. This is to avoid breaking user journeys during
deployment.

### DIFF
--- a/app/paths.js
+++ b/app/paths.js
@@ -163,11 +163,14 @@ module.exports = {
     update: '/service/:externalServiceId/edit-name'
   },
   merchantDetails: {
-    index: '/organisation-details/:externalServiceId',
-    edit: '/organisation-details/edit/:externalServiceId'
+    index: '/service/:externalServiceId/organisation-details',
+    edit: '/service/:externalServiceId/organisation-details/edit',
+    indexOld: '/organisation-details/:externalServiceId',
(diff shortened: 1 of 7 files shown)


app/paths.js
```diff
@@ -163,11 +163,14 @@ module.exports = {
     update: '/service/:externalServiceId/edit-name'
   },
   merchantDetails: {
-    index: '/organisation-details/:externalServiceId',
-    edit: '/organisation-details/edit/:externalServiceId'
+    index: '/service/:externalServiceId/organisation-details',
```
sfount, line 166: Do we want these paths to be organised into `service` similar to the `account` paths?
```js 
{
  service: { },
  account: {}
}
```

How paths are organised wont change the functionality so totally fine either way

sfount (approved):

app/paths.js
stephencdaly: They will be in the PR to add the router, just bodged in this change for now so didn't have to change things twice.

PR #2668, opened 2021-03-30
PP-7735 Remove direct debit code

stephencdaly:
Remove code that requires an email for the organisation details for a direct debit account, as we no longer support direct debit.


kbottla (approved):
LGTM!

PR #2674, opened 2021-04-01
PP-7734 Remove get-gateway-account middleware

stephencdaly:
The only thing this middleware was still used for was setting the gateway account cookie to the first gateway account found on login if it wasn't already set.

The cookie is still used for upgrading old account URLs without the external ID to the new URLs based on the account in the cookie. We will now redirect to the "My services" page if no cookie is set.


sfount (approved):
🎉

PR #2675, opened 2021-04-01
PP-7734 Remove usages of gateway account cookie

stephencdaly:
Remove all usages of the `gateway_account` cookie that are no longer required.

For now, continue setting `gateway_account.currentGatewayAccountExternalId` which is used for upgrading old account URLs, but we may remove this soon or only update it if the user already has the cookie set as we want to retire this behaviour.


kbottla (approved):
LGTM!

PR #2684, opened 2021-04-09
PP-7570 Use AsyncLocalStorage to provide shared logging context

stephencdaly:
Add middleware to open an AsyncLocalStorage store on a request. Add the correlation ID when the store is initialised.

Add a format function to the logger that pulls fields that have been added to the store and adds them to the log line.

Add middleware to add user info to the log context when it is available. Add account and service info to the log context in the middleware that resolves them.

Re-order the server middleware so that the auth middleware is run after the global middleware that initialises logging amongst other things.

### DIFF
--- a/app/services/auth.service.js
+++ b/app/services/auth.service.js
@@ -4,6 +4,8 @@ const lodash = require('lodash')
 const passport = require('passport')
 const LocalStrategy = require('passport-local').Strategy
 const CustomStrategy = require('passport-custom').Strategy
+const { addLoggingField } = require('../utils/log-context')
+const { USER_EXTERNAL_ID } = require('@govuk-pay/pay-js-commons').logging.keys
 
 // TODO: Remove when issue solved
 /*
@@ -112,6 +114,14 @@ function hasValidSession (req) {
   return isValid
 }
 
+function addUserFieldsToLogContext (req, res, next) {
+  if (req.user) {
+    addLoggingField(USER_EXTERNAL_ID, req.user.externalId)
+    addLoggingField('is_internal_user', req.user.internalUser)
+  }
+  next()
+}
+
 function initialise (app) {
   app.use(passport.initialize())
   app.use(passport.session())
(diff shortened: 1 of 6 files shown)


app/utils/log-context.js
```diff
@@ -0,0 +1,41 @@
...
+
+const supplementSharedLoggingFields = format((info) => {
+  if (!asyncLocalStorage.getStore()) {
+    return info
+  }
+
+  return Object.assign(info, asyncLocalStorage.getStore())
+})
+
+function addUserLoggingFields (user) {
```
sfount: The pattern before this felt a bit too specific for the logging context utility but it feels like this finds the right balance. For common entities extracting and including info we think is relevant in one place feels quite clean. 

We could potentially do this for services and accounts too for routes that pertain to those. 

Some ideas 

```js 
// service external id
// go live status is "LIVE" yet or not? - unsure if that's useful
function addServiceLoggingFields(service) {}

// live or test
// account id
// account external id?
// payment provider
function addAccountLogginFields(account) {}
```

app/services/auth.service.js
```diff
@@ -120,6 +121,10 @@ function initialise (app) {
   passport.use('localDirect', new CustomStrategy(localDirectStrategy))
   passport.serializeUser(serializeUser)
   passport.deserializeUser(deserializeUser)
+  app.use((req, res, next) => { 
+    req.user && addUserLoggingFields(req.user)
```
sfount: (very minor!) I'm a fan of short circuit evaluation for assignment but have seen it get a bit confusing to read when used for method calls that have side effects. 

Interested in your opinion but I think these would benefit from being a bit more verbose but a bit clearer 
```suggestion
    if (req.user) {
      addUserLoggingFields(req.user) 
    }
```

sfount (approved):
Looks great 🍬 . This along using the shared context for passing on the correlation ID to downstream services will clean up a lot of code

sfount (approved):

PR #2685, opened 2021-04-09
PP-7570 Clean up logging

stephencdaly:
As we now attach the correlation id, user, service, and account info to all log lines when it is available using AsyncLocalStorage, we no need to attach this information to individual log lines.

See b18c7571d774ab3c591d4bf614f24ac38a923de0 for implementation of this.

Remove occurrences where we log this info in individual log lines.

### DIFF
--- a/app/services/auth.service.js
+++ b/app/services/auth.service.js
@@ -38,8 +38,7 @@ module.exports = {
 // Middleware
 function lockOutDisabledUsers (req, res, next) {
   if (req.user && req.user.disabled) {
-    const correlationId = req.headers[CORRELATION_HEADER] || ''
-    logger.info(`[${correlationId}] user: ${lodash.get(req, 'user.externalId')} locked out due to many password attempts`)
+    logger.info('User locked out due to many password attempts')
     return noAccess(req, res, next)
   }
   return next()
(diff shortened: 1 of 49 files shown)


app/services/auth.service.js
```diff
@@ -104,20 +103,20 @@ function setSessionVersion (req) {
...
-  if (!isValid) logger.info(`[${req.correlationId}] Invalid session version for user. User session_version: ${lodash.get(req, 'user.sessionVersion', 0)}, session version ${lodash.get(req, 'session.version')}`)
+  if (!isValid) logger.info(`Invalid session version for user. User session_version: ${lodash.get(req, 'user.sessionVersion', 0)}, session version ${lodash.get(req, 'session.version')}`)
   return isValid
 }
 
 function addUserFieldsToLogContext (req, res, next) {
   if (req.user) {
     addLoggingField(USER_EXTERNAL_ID, req.user.externalId)
-    addLoggingField('is_internal_user', req.user.internalUser)
+    addLoggingField('internal_user', req.user.internalUser)
```
kbottla, line 119: This affects few selfservice Splunk dashboards

kbottla (approved):
Looks good to me. Renaming `is_internal_user` log field to `internal_user` affects few dashboards. Happy to rename here and update dashboards or change it back to `is_internal_user`

app/services/auth.service.js
stephencdaly: we inconsistently have `is_internal_user` and `internal_user` for different log lines, so I picked the one we used for the transaction search log lines. Good point though, I'll check if any dashboards need updating.

stephencdaly: I'll keep it as `internal_user` and update the dashboards

PR #2691, opened 2021-04-16
PP-7570 Fix logging context for POST requests

stephencdaly:
There is an issue when body-parser middleware is added after an AsyncLocalStorage store is opened, where the async context is lost.

This is identified here https://stackoverflow.com/questions/64330910/nodejs-asynclocalstorage-getstore-return-undefined, although no cause is identified.

This ordering was also required in Toolbox when using a different library using async_hooks.

Change middleware ordering to fix this.

### DIFF
--- a/server.js
+++ b/server.js
@@ -58,6 +58,9 @@ function addCsrfMiddleware (app) {
 }
 
 function initialiseGlobalMiddleware (app) {
+  app.use(bodyParser.json())
+  app.use(bodyParser.urlencoded({ extended: true }))
+
   app.use(cookieParser())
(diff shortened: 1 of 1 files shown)


sfount (approved):
Similar behaviour observed in Toolbox but not much context from the time https://github.com/alphagov/pay-toolbox/commit/fc224728fb70105c59997c36e6ed86904391e0e5

Potentially we can submit an issue questioning the behaviour to have someone who knows the details of the `json()` body parser comment, it may be something we can pitch in to help fix upstream in express.

PR #2695, opened 2021-04-20
PP-8003 - Release Stripe content changes

nimalank7:
Description:
- Display 'No payments to your bank acount found on or after...' only for live accounts with no payouts
- Display sample report and inset text explaining test reports only for test accounts with no payouts
- Added a cypress test to verify if sample report is displayed on test account if there are no payouts
- Added a cypress test to verify if text is shown indicating no payouts on a live account if there are no payouts

### DIFF
--- a/app/assets/csv/sample_reporting_payout.csv
+++ b/app/assets/csv/sample_reporting_payout.csv
@@ -1,0 +1,2 @@
+Reference,Description,Email,Amount,Card Brand,Cardholder Name,Card Expiry Date,Card Number,State,Finished,Error Code,Error Message,Provider ID,GOV.UK Payment ID,Issued By,Date Created,Time Created,Corporate Card Surcharge,Total Amount,Wallet Type,Fee,Net,Card Type,MOTO
+4S8CKS3FN2,test kb,useremail@example.org,10,Visa,test,11/21,4242,Success,true,,,pi_1HG0y2Hj08j2jFuBmM9r9Ybe,9mim88o2gt7atngu9m7i78en6a,,14 Aug 2020,11:20:48,0,10,,0.5,9.5,credit,false
--- a/app/views/payouts/list.njk
+++ b/app/views/payouts/list.njk
@@ -24,9 +24,8 @@ Payments to your bank account - GOV.UK Pay
   </h1>
 </div>
 <div class="govuk-grid-column-two-thirds govuk-!-margin-bottom-0">
-  {% if filterLiveAccounts %}
-    <p class="govuk-body">Payments Stripe has made to your bank account.</p>
-  {% else %}
+  <p class="govuk-body">Payments Stripe has made to your bank account.</p>
+  {% if not filterLiveAccounts %}
     <div class="govuk-inset-text">
     <p class="govuk-body">Test reports represent how payments to your bank account appear. They show the total amount received from Stripe in each payment and detail all the transactions.</p>
     <p class="govuk-body">The reports are available the next working day after your test transactions. With test transactions, no payments are made to your bank account.</p>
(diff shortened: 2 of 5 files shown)


app/views/payouts/list.njk
```diff
@@ -83,12 +82,39 @@ Payments to your bank account - GOV.UK Pay
...
+            <th scope="col" class="govuk-table__header">Service</th>
+            <th scope="col" class="govuk-table__header govuk-table__header--numeric">Amount</th>
+            <th scope="col" class="govuk-table__header">View transactions</th>
+          </tr>
+        </thead>
+        <tbody class="govuk-table__body">
+          <tr class="govuk-table__row">
+            <td scope="row" class="govuk-table__cell">Test Service</td>
+            <td class="govuk-table__cell govuk-table__cell--numeric">
+              {{ 10000 | penceToPoundsWithCurrency }}
```
kbottla, line 108: We could also use `&#163;100.00` as this is static, but fine with ^

app/views/payouts/list.njk
```diff
@@ -83,12 +82,39 @@ Payments to your bank account - GOV.UK Pay
...
+      {% if payoutsReleaseDate %}
+        <p class="govuk-body">No payments to your bank account found on or after {{ payoutsReleaseDate.format('D MMMM YYYY') }}.</p>
+        <p class="govuk-body">For historic payments to your bank account please <a class="govuk-link" href="https://www.payments.service.gov.uk/support/">contact support.</a></p>
+      {% else %}
+        <p class="govuk-body">No payments to your bank account found.</p>
+      {% endif %}
     {% else %}
-    <p class="govuk-body">No payments to your bank account found.</p>
+      <h2 class="govuk-heading-m">Sample Report</h2>
+
```
kbottla, line 95: Some text missing here about sample report which reads `This is a sample report .......`

nimalank7: Fixed.

app/views/payouts/list.njk
nimalank7: Sure.

app/views/payouts/list.njk
```diff
@@ -83,12 +82,40 @@ Payments to your bank account - GOV.UK Pay
...
+          </tr>
+        </thead>
+        <tbody class="govuk-table__body">
+          <tr class="govuk-table__row">
+            <td scope="row" class="govuk-table__cell">Test Service</td>
+            <td class="govuk-table__cell govuk-table__cell--numeric">
+              {{ 10000 | penceToPoundsWithCurrency }}
+            </td>
+            <td class="govuk-table__cell">
+              <a class="govuk-link govuk-link--no-visited-state" href="public/csv/sample_reporting_payout.csv">
```
stephencdaly, line 112: Does this link work? I tried testing locally and got page not found. Does it also need to have the `download` attribute?

stephencdaly: Looks like the href should be `/csv/sample_reporting_payout.csv`

app/assets/csv/sample_reporting_payout.csv
```diff
@@ -0,0 +1,2 @@
+Reference,Description,Email,Amount,Card Brand,Cardholder Name,Card Expiry Date,Card Number,State,Finished,Error Code,Error Message,Provider ID,GOV.UK Payment ID,Issued By,Date Created,Time Created,Corporate Card Surcharge,Total Amount,Wallet Type,Fee,Net,Card Type,MOTO
+4S8CKS3FN2,test kb,useremail@example.org,10,Visa,test,11/21,4242,Success,true,,,pi_1HG0y2Hj08j2jFuBmM9r9Ybe,9mim88o2gt7atngu9m7i78en6a,,14 Aug 2020,11:20:48,0,10,,0.5,9.5,credit,false
```
stephencdaly, line 2: Could we put a more realistic test payment description, something like "Pay your council tax", and realist cardholder name etc. maybe lift from the API documentation? https://docs.payments.service.gov.uk/making_payments/#take-a-payment

Also, the amount should match the total on the payouts page. Have we talked about whether we want a number of transactions that add up to that amount to make this a more realistic report? The total amount on the payouts page should match the total net amount in the CSV

stephencdaly (changes requested):

app/views/payouts/list.njk
nimalank7: Nice spot will change that now!

PR #2725, opened 2021-05-04
PP-8032 Only send unexpected errors to Sentry

stephencdaly:
We have moved towards the pattern of middleware and controllers passing errors to `next` rather than the middleware/controller handling all errors themselves. All errors are then handled in the `error-handler.js` middleware.

The change to this pattern meant we were sending all errors that are handled by the `error-handler.js` middleware to Sentry, even business as usual errors such as a user not being authenticated.

Fix this by not using the `Sentry.Handlers.errorHandler()` middleware and instead only sending unexpected errors to Sentry using `Sentry.captureException` in our error handling middleware.

Also remove the `error-logger.js` middleware which logs middleware errors at error level. This predated the `error-handler.js` middleware and just adds duplicate logging.


app/middleware/error-handler.js
```diff
@@ -60,5 +62,6 @@ module.exports = function errorHandler (err, req, res, next) {
   logger.error(`Unhandled error caught: ${err.message}`, {
     stack: err.stack
   })
+  Sentry.captureException(err)
```
sfount, line 65: Along with capturing the exception with Sentry here, is there a reflective part to remove it from the existing method of sending an event to sentry on `error` level logs? 

An example of the design consideration there is that the line of code above registers a `logger.entry` which I _think_ (would need confirming) will be treated as a unique event by Sentry. 

Code that sets that up https://github.com/alphagov/pay-selfservice/blob/master/app/utils/logger.js#L29

sfount (approved):
Changes look good 👍  worth sense checking the error logging middleware format isn't used by any saved searches -- it seems like that will have been duplicating logs that are logged by handlers for a long time, it does include some extra info but something I assume would be taken care of my logging nearer the action rather than in generic middleware

PR #2726, opened 2021-05-06
PP-8032 Standardise error handling

stephencdaly:
For unexpected errors, pass the error to `next` to be handled by our error handling middleware. This will allow for errors to be handled uniformly, displaying the same page for all unexpected errors as recommended by the design system.

This will also allow us to in future stop logging at error level for all error pages shown, as the error handler will send unexpected errors to Sentry explicitly. When we do this the noise in Sentry caused by business as usual errors will be reduced.

## Note

Hiding whitespace changes will make this PR easier to review.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json",
     "lines": null
   },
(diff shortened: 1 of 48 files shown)


kbottla (approved):
LGTM!

PR #2727, opened 2021-05-06
PP-8032 Handle errors uniformly for registration

stephencdaly:
Add a new error type to handle in the error handler which is thrown when the registration data isn't found in the session cookie. This allows us to handle these errors uniformly. Use a 400 error code for this as it makes more sense than 404.

Handle unexpected errors by passing an error to next to show the standard 500 error page.


SandorArpa (approved):
LOKTM

PR #2729, opened 2021-05-07
PP-8032 Don't log at error level every time we show error

stephencdaly:
All unexpected errors are now handled by the error handler middleware which sends the error to Sentry before rendering the internal error page.

There are many other times we render the error view, when there are errors for things such as cookies missing or invites having expired which are "business as usual" and do not need to be investigated by a developer. This will stop events being sent to Sentry for these type of errors.


SandorArpa (approved):
LGTM

PR #2741, opened 2021-05-14
PP-8048 Add step to workflow job to run Cypress tests

stephencdaly:
We might want this as a separate job we can run in parallel, particularly if we can share the build artefacts so they only need to be built once for the workflow. For now add as another step to the same job.


SandorArpa (approved):
LOKTM

PR #2742, opened 2021-05-14
PP-8048 Run unit and Cypress tests in parallel

stephencdaly:
- First job runs install and compile, and caches the output directories
- Second job restores cache of build directories and runs unit tests
- Third job restores cache of build directories and runs Cypress tests

Unit and Cypress test jobs run in parallel. The build dependencies are cached using the commit SHA and the HEAD reference in the key, so they will be unique to the commit and the branch it's on, so should only be cached for a single workflow run or repeated workflow runs for the same commit.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -10,9 +10,9 @@ permissions:
   contents: read
 
 jobs:
-  unit-tests:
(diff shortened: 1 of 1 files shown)


SandorArpa (approved):
LGTM 👍

PR #2764, opened 2021-06-03
PP-8131 Namespace PSP and credential pages by provider

sfount:
Link specficially to payment provider versions of PSP and credential
configuration pages.

This willl support individual accounts to link to multiple credential
configurations.

It will also allow supported PSP credential specific pages, Worldpay
with Worldpay credential validation.

### How to test

- The "Your PSP" and "Credentials" pages should work exactly as they do now, no user facing changes


stephencdaly (approved):
Looks good!

PR #2765, opened 2021-06-03
PP-8131 Worldpay credentials specific creds page

sfount:
Use new `/worldpay/check-credentials` Connector route to make sure service users have provided correct credentials before allowing submission.

Also
- Implement worldpay credentials against current language best practices
- Simple backend form validation meets design system recommendations
- Page is covered by existing credentials Cypress tests

### DIFF
--- a/app/controllers/credentials/credentials-form.js
+++ b/app/controllers/credentials/credentials-form.js
@@ -40,18 +40,10 @@ class CredentialsForm {
       errorSummaryList: formatErrorsForSummaryList(errors)
     }
   }
-
-  // render(formData = {}) {
-  //   return this.fields.reduce((aggregate, field) => {
-  //     const key = field.key || field.id
-  //     aggregate[key] = typeof formData[field.id] === 'string' ? formData[field.id].trim() : formData[field.id]
-  //     return aggregate
-  //   }, {})
-  // }
 }
 
 function isNotEmpty (value) {
   return value && value.length !== 0
 }
 
-module.exports = { CredentialsForm, isNotEmpty }
+module.exports = { CredentialsForm, isNotEmpty, formatErrorsForSummaryList }
--- a/app/controllers/credentials/credentials-form.js
+++ b/app/controllers/credentials/credentials-form.js
@@ -1,0 +1,57 @@
+function formatErrorsForSummaryList (errors = {}) {
+  return Object.entries(errors).map(([id, message]) => ({
+    href: `#${id}`,
+    text: message
+  }))
+}
+class CredentialsForm {
+  constructor (fields = []) {
+    this.fields = fields
+    this.values = {}
+  }
+
+  from (entity = {}) {
+    const values = this.fields.reduce((aggregate, field) => {
+      const key = field.key || field.id
+      aggregate[field.id] = entity[key]
+      return aggregate
+    }, {})
+    return { values }
+  }
+
+  validate (formData = {}) {
+    const errors = {}
+    const values = {}
+
+    this.fields.forEach((field) => {
+      const valid = field.valid || []
+      values[field.id] = typeof formData[field.id] === 'string' ? formData[field.id].trim() : formData[field.id]
+      valid.some((validator) => {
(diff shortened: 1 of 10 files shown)


sfount:
🤔 it would be helpful if there was logging around the "invalid" credentials scenario -- how often that is presented to users might prove useful for support queries.

sfount:
Playing around with setting `autocomplete=` on the input elements could try and stop Chrome from its invasive credentials management on generically named "username" and "password"

app/controllers/credentials/credentials-form.js
```diff
@@ -0,0 +1,49 @@
...
+        }
+      })
+    })
+    return {
+      values,
+      errors,
+      errorSummaryList: formatErrorsForSummaryList(errors)
+    }
+  }
+}
```
stephencdaly, line 43: Is the intention that most of this code could be re-used for any form, not just credentials?

app/controllers/credentials/credentials-form.js
```diff
@@ -0,0 +1,49 @@
...
+
+function isNotEmpty (value) {
```
stephencdaly, line 45: Even when we're sharing this code for any form, does it make more sense for this function live in a separate validation helper class, which would maybe replace `server-side-form-validations.js` when we more widely use this pattern?

app/controllers/credentials/credentials-form.js
```diff
@@ -0,0 +1,49 @@
...
+    }, {})
+    return { values }
+  }
+
+  validate (formData = {}) {
+    const errors = {}
+    const values = {}
+
+    this.fields.forEach((field) => {
+      const valid = field.valid || []
```
stephencdaly, line 27: this is where typescript would be so useful! I know that it's not gonna happen anytime soon though :(

app/controllers/credentials/worldpay.controller.js
```diff
@@ -0,0 +1,48 @@
...
+  const results = credentialsForm.validate(req.body)
+
+  if (results.errorSummaryList.length) {
+    return response(req, res, 'credentials/worldpay', { form: results })
+  }
+
+  try {
+    const checkCredentialsWithWorldpay = await connectorClient.postCheckWorldpayCredentials({ correlationId, gatewayAccountId, payload: results.values })
+    if (checkCredentialsWithWorldpay.result !== 'valid') {
+      results.errorSummaryList = formatErrorsForSummaryList({ 'merchant_id': 'Unable to use credentials with Worldpay, check entered credentials' })
```
stephencdaly, line 33: Does this linking to the the merchant_id field suggest that there's a problem with the merchant id specifically? Not sure of the answer to this...

test/cypress/stubs/gateway-account-stubs.js
```diff
@@ -196,6 +196,14 @@ function postCheckWorldpay3dsFlexCredentials (opts) {
   }
 }
 
+function postCheckWorldpayCredentials (opts) {
+  const path = `/v1/api/accounts/${opts.gatewayAccountId}/worldpay/check-credentials`
+  return stubBuilder('POST', path, 200, {
+    request: opts.credentials,
+    response: { result: 'valid' }
```
stephencdaly, line 203: Should use a fixture with a corresponding pact test. Fine for that to be in a separate PR though.

stephencdaly (approved):
Looks good and the controller code is very elegant using this pattern.

It would be good to make sure this form pattern is generified and pointed to from our node guidance as part of this work to ensure this doesn't become yet another pattern that isn't replicated in other places.

Left a question about the linking from the error summary when credentials fail validation with Worldpay which might be one to check with someone else if we aren't confident of the answer.

And needs a pact test, but can be added in the next PR.

sfount:
Thanks for the review @stephencdaly

Recorded for a follow up PR to tackle: 
- consumer pact tests covering the credentials checking route
- logging around valid/ invalid response handling
- iterate the top level generic form error handling pattern if appropriate

I'll take this through to a test environment to check connecting all the dots.

PR #2787, opened 2021-06-14
BAU Fix setup for flakey worldpay flex pact

sfount:
I'm not familiar with the chai `should` syntax but this unit test seems
to fail every so often for no reason.

Use async/ await syntax to make it clearer what's going on and review
behaviour to make sure it will only return once the request and
assertion have completed.

### How to test
- test should pass 
- test should always pass, independent of CPU cycles, temperature, day of the week, alignment of the stars


kbottla (dismissed):
LGTM!

kbottla (approved):

PR #2788, opened 2021-06-14
PP-8213 Add gateway account credentials fixtures and util (bump-ci)

sfount:
- adds fixtures for `provider_switch_enabled` and
`gateway_account_credentials` account properties
- add simplest case utility to make use of the new credentials

### DIFF
--- a/app/utils/credentials.js
+++ b/app/utils/credentials.js
@@ -5,7 +5,7 @@ const { InvalidConfigurationError } = require('../errors')
 const CREDENTIAL_STATE = {
   CREATED: 'CREATED',
   ENTERED: 'ENTERED',
-  VALIDATED: 'VALIDATED_WITH_LIVE_PAYMENT',
+  VERIFIED: 'VERIFIED_WITH_LIVE_PAYMENT',
   ACTIVE: 'ACTIVE',
   RETIRED: 'RETIRED'
 }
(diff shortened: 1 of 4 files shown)


app/utils/credentials.js
```diff
@@ -0,0 +1,37 @@
+'use strict'
+
+const { InvalidConfigurationError } = require('../errors')
+
+const CREDENTIAL_STATE = {
+  CREATED: 'CREATED',
+  ENTERED: 'ENTERED',
+  VALIDATED: 'VALIDATED_WITH_LIVE_PAYMENT',
```
stephencdaly, line 8: This should be `VERIFIED_WITH_LIVE_PAYMENT` https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/model/GatewayAccountCredentialState.java

test/fixtures/gateway-account.fixtures.js
```diff
@@ -15,6 +15,22 @@ function validCredentials (opts = {}) {
   return credentials
 }
 
+function validGatewayAccountCredential (credentialOpts = {}, gatewayAccountOpts = {}) {
+  const gatewayAccountCredential = {
+    gateway_account_credential_id: credentialOpts.id || 1,
+    payment_provider: gatewayAccountOpts.payment_provider || 'sandbox',
+    state: credentialOpts.state || 'ACTIVE',
+    gateway_account_id: gatewayAccountOpts.gateway_account_id || 31,
+    // '2021-05-10T01:00:00.000Z'
```
stephencdaly, line 24: what's this comment for?

test/fixtures/gateway-account.fixtures.js
```diff
@@ -73,6 +97,11 @@ function validGatewayAccount (opts) {
     gatewayAccount.worldpay_3ds_flex = validWorldpay3dsFlexCredentials(opts.worldpay_3ds_flex)
   }
 
+  // provider switch enabled is only available to the "frontend" resource, it isn't always guaranteed
+  if (opts.provider_switch_enabled) {
```
stephencdaly, line 101: We might want to set to false in the opts? should this check if it is undefined?

stephencdaly (changes requested):
Looks good, just need to fix the enum value - and I left a few other minor comments

test/fixtures/gateway-account.fixtures.js
sfount: I originally had this set to opts or default `false` but it fails pacts because the confusing `/v1/api` routes won't return the value, I guess as it fails the same fixture is used for both

test/fixtures/gateway-account.fixtures.js
sfount: This was another thing that was originally default but isn't guaranteed from the provider, the comment shows what format it expects but I forgot it was in there to be honest, probably isn't needed

stephencdaly (approved):

PR #2789, opened 2021-06-14
PP-8213 Switching PSP base page

sfount:
Add controllers, template and base test structure for the switching PSP page.

Future pull requests will add individual tasks and completion in conjunction with the backend specs becoming available.

* if an account is not configured to switch, navigating to the page `404`s 
* if an ambiguous switching configuration is found (too many or too few credentials to fill in) an invalid configuration error is raised and handled
* page descriptions and requirements templated in based on the provider being switched to
* cover new middleware and controllers

**Base page that can be integrated with individual future steps**

Screenshot 2021-06-14 at 18 16 06

### DIFF
--- a/app/views/switch-psp/switch-psp.njk
+++ b/app/views/switch-psp/switch-psp.njk
@@ -1,0 +1,58 @@
+{#
+  - title, into paragraph, task list category headers, final paragraph and action button all just template in the provider going to
+  - description paragraph with what you'll need selected based on provider (set innline)
+  - task list items likely set by controller, content for it can be set inline? not exactly sure how that will look
+
+  {% set tasks %}
+  - manually <li> the tasks, they come with aria descriptions etc. that will take more effort to generalise
+  - should add a `tasks` to the context, it will be based on which psp is being switched to and have an interface like task[id] = { status: }
+ #}
+{% extends "layout.njk" %}
+
+{% block pageTitle %}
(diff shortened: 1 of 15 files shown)


app/views/switch-psp/switch-psp.njk
```diff
@@ -0,0 +1,49 @@
...
+    <li>
+        <h2 class="app-task-list__section">
+          <span class="app-task-list__section-number">1. </span> Get ready to switch PSP
+        </h2>
+        <ul class="app-task-list__items"></ul>
+      </li>
+
+      <li>
+        <h2 class="app-task-list__section" id="switch-psp-action-step">
+          <span class="app-task-list__section-number">2. </span> Switch PSP to {{ targetCredential.payment_provider | formatPSPname }}
```
stephencdaly, line 37: It looks like this is working for you, but I've run `npm install` and `npm run compile` and this isn't appearing correctly for me. Where does this CSS come from?

Screenshot 2021-06-15 at 12 14 46

stephencdaly (commented):
Spotted an edge case, and question about CSS

stephencdaly (approved):
Looks good 👍

app/views/switch-psp/switch-psp.njk
sfount: Added missing SCSS

PR #2792, opened 2021-06-17
PP-8216 Add Cypress tests for task list

stephencdaly:
Add Cypress tests to check the task list is displayed properly with correct tags etc. and to test the journey for completing the first step.


kbottla (approved):
LGTM!

PR #2793, opened 2021-06-18
PP-8215 Switching payment validator tool

sfount:
Adds pages
* verify PSP integration page, routes to start and receive test payment journey hooks

Extends connector helpers
* add helper methods to interact with Connector charges (creating and
checking in-flight status)
* fixtures for charge interaction (POST and GET return the same shape of
data)

Updates switching PSP tasks
* refactored to use a new `taskListItem` macro that will keep formatting consistent for all future tasks
* added task for verifying PSP integration

Proposed to add in a follow up PR
* pacts with Connector for new endpoints
* Cypress happy path for this task list item

Recommend stepping through commits as each piece was done in a logical order, happy to split into multiple PRs if more appropriate for reviewing.

Journey step page

Screenshot 2021-06-18 at 11 56 03

Successful task item completion 

Screenshot 2021-06-18 at 14 56 49

Failed task item completion

Screenshot 2021-06-18 at 14 57 04

PR #2794, opened 2021-06-18
PP-8215 Cypress coverage for payment validator task 

sfount:
Depends on https://github.com/alphagov/pay-selfservice/pull/2793

Add Cypress coverage for task lisk item to verify PSP integration.
Asserts that high level success and non-success payment routes are
appropriately handled.


kbottla (approved):
LGTM!

PR #2803, opened 2021-06-21
PP-8215 Fix using absolute path for cypress environment

sfount:
Jenkins sets a different URL for end to end tests that other
envrionments. Use a relative URL for the `next_url` on switch PSP tests
so that it doesn't matter what environment it is run in.


kbottla (approved):

SandorArpa (approved):
LGTM

PR #2806, opened 2021-06-22
PP-8190 Your PSP and credential pages use credentials list

sfount:
Existing psp pages use the top level gateway account `credentials` blob
to display and manage account credentials. Move this to be name-spaced by
the gateway account credential entity list to support switching.

Use the list of credentials to provide links in the settings menu.

The primary change is from `currentGatewayAccount.credentials` -> `getCredentialByExternalId().credentials`

Adds logic with coverage to make sure we're picking out terminal state credentials 
for when an account is switching.

Proposes a new URL structure to make things consistent and clear for devs 
`/your-psp/:id/`
`/your-psp/:id/credentials`
`/switch-psp/:id/credentials`


SandorArpa (approved):
LOKTM

PR #2810, opened 2021-06-24
PP-8277 add patch gateway merchant id and patch state pact test

SandorArpa:
## WHAT
- add a pact test to patch Google Pay gateway merchant id
- add a pact test to patch gateway credential state

### DIFF
--- a/test/fixtures/gateway-account.fixtures.js
+++ b/test/fixtures/gateway-account.fixtures.js
@@ -195,7 +195,7 @@ function validUpdateGatewayAccountCredentialsRequest (opts = {}) {
   const defaultCredentials = {
     username: 'a-username',
     password: 'a-password', // pragma: allowlist secret
-    'merchant_id': 'a-merchant-id'
+    merchant_id: 'a-merchant-id'
   }
   return [
     {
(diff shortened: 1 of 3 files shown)


test/fixtures/gateway-account.fixtures.js
```diff
@@ -237,6 +237,84 @@ function validGatewayAccountCredentialsResponse (opts = {}) {
...
+  return [
+    {
+      op: 'replace',
+      path: 'state',
+      value: opts.state || 'VERIFIED_WITH_LIVE_PAYMENT'
+    }
+  ]
+}
+
+function validPatchAccountGatewayAccountCredentialsStateResponse (opts = {}) {
```
sfount, line 292: `validPatchAccountGatewayAccountCredentialsStateResponse` and `validPatchGatewayMerchantIdResponse` seem to contain exactly the same code, as this is just a fixture builder could they just be the same method?

It already has a pattern to do
```
if (opts.x !== undefined) {
  data.x = opts.s
}
```

so maybe this could be done for `gateway_merchant_id` to avoid the duplication (I think that's the only difference?)

sfount (approved):

PR #2813, opened 2021-06-25
PP-8277 add post switch payment provider pact test

SandorArpa:
## WHAT
- add new pact test for switching payment provider (psp)


nimalank7 (approved):
LGTM!

PR #2814, opened 2021-06-25
PP-8277 add post create charge pact test

SandorArpa:
## WHAT
- add new pact test for creating a charge

### DIFF
--- a/test/fixtures/charge.fixtures.js
+++ b/test/fixtures/charge.fixtures.js
@@ -1,0 +1,57 @@
+'use strict'
+
+const { string } = require('@pact-foundation/pact').Matchers
+
+function validPostChargeRequestRequest (opts = {}) {
+  return {
+    amount: opts.amount || 100,
+    reference: opts.reference || 'a reference',
+    description: opts.description || 'a description',
+    return_url: opts.returnUrl || 'https://somewhere.gov.uk/rainbow/1'
+  }
+}
+
+// this fixture is using matchers (string = type)
+function validPostChargeRequestResponse (opts = {}) {
+  return {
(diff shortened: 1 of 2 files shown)


test/fixtures/charge.fixtures.js
```diff
@@ -0,0 +1,63 @@
...
+  return {
+    charge_id: string('ch-ab2341da231434l'),
+    amount: 100,
+    reference: opts.reference || 'a reference',
+    description: opts.description || 'a description',
+    return_url: opts.returnUrl || 'https://somewhere.gov.uk/rainbow/1',
+    state: opts.state || { status: 'created',
+      finished: false
+    },
+    links: opts.links || [
```
stephencdaly, line 25: We probably only need to include the `next_url` in the pact as it's the only link that selfservice consumes which would simplify this a little.

Also do we definitely need the regex matcher on the links? The pact needs to ensure that the contract doesn't change, but the unit/integration tests for connector should ensure that the URL being returned is of the correct format, so selfservice should be able to trust what it receives, unless it has to manipulate the URL in some way.

stephencdaly (dismissed):
Nothing wrong with the pact test so fine to go in, just questioned whether we could simplify the pact generation a little.

test/fixtures/charge.fixtures.js
stephencdaly: Discussed, and there isn't a way to verify that a particular element exists in an array using matchers, so we need to add all the links to the pact to check that there is an element in the array with "rel": "next_url".

stephencdaly (approved):

PR #2824, opened 2021-07-02
PP-8214 Switch PSP - update dashboard banner and labels

kbottla:
## WHAT
- Updated dashboard banner & labels to latest content
- Banner and switch PSP labels are only available to admin users of service


stephencdaly (approved):
LGTM!

PR #2832, opened 2021-07-07
PP-8219 Add psp switch status inset text

stephencdaly:
Add inset text on the "Your PSP" page to indicate:
- When switch is in progress
- On the "Old PSP" page after switch, when the provider switched
- On the "Your PSP" page after switch, when the current provider became active

Display dates in the format "7 July 2021".

### DIFF
--- a/app/views/your-psp/index.njk
+++ b/app/views/your-psp/index.njk
@@ -30,7 +30,20 @@
 
   {% if credential.active_end_date %}
   <div id="switched-psp-status" class="govuk-inset-text">
-    This service is taking payments with {{ activeCredential.payment_provider | formatPSPname }}. It switched from using {{ credential.payment_provider | formatPSPname }} on {{ credential.active_end_date | datetime('date') }}
+    This service is taking payments with {{ activeCredential.payment_provider | formatPSPname }}. It switched from using {{ credential.payment_provider | formatPSPname }} on {{ credential.active_end_date | datetime('datelong') }}
+  </div>
+  {% elif switchedProvider and credential.active_start_date %}
+  <div id="switched-psp-status" class="govuk-inset-text">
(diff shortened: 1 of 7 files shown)


app/views/your-psp/index.njk
```diff
@@ -30,7 +30,20 @@
...
+    This service is taking payments with {{ activeCredential.payment_provider | formatPSPname }}. It switched from using {{ credential.payment_provider | formatPSPname }} on {{ credential.active_end_date | datetime('datelong') }}
+  </div>
+  {% elif switchedProvider and credential.active_start_date %}
+  <div id="switched-psp-status" class="govuk-inset-text">
+    This service started taking payments with {{ credential.payment_provider | formatPSPname }} on {{ credential.active_start_date | datetime('datelong') }}.
+  </div>
+  {% elif switchingCredential %}
+  <div id="switched-psp-status" class="govuk-inset-text">
+    <p class="govuk-body">
+      Your service is ready to switch PSP from {{ credential.payment_provider | formatPSPname }} to {{ switchingCredential.payment_provider | formatPSPname }}.
```
sfount, line 42: This uses the credential we're updating on the page -- I think the active credential is more specific and available to the template. 

For the currently supported use case this will be fine but this might help avoid future confusion.

```suggestion
      Your service is ready to switch PSP from {{ activeCredential.payment_provider | formatPSPname }} to {{ switchingCredential.payment_provider | formatPSPname }}.
```

sfount (commented):
LGTM 👍 small suggestion and coverage of exported utils as we discussed

app/views/your-psp/index.njk
stephencdaly: 👍  updated

sfount (approved):

PR #2834, opened 2021-07-08
PP-8272 Redirect to test connection page on payment failure

stephencdaly:
Previously, we were redirecting to the "Switch PSP" page if the test payment failed. Instead redirect to the "Test connection" page.

Use `req.flash` to display the error/success messages and remove some of the boiler plate as it makes less sense now they're displayed on different pages.


kbottla (approved):
LGTM!

PR #2839, opened 2021-07-12
PP-8298 Patch user_external_id when patching credentials state

kbottla:
## WHAT
- Patches `last_updated_by_user_external_id` when patching credentials state (to VERIFIED_WITH_LIVE_PAYMENT)


SandorArpa (approved):
LOKTM

PR #2841, opened 2021-07-12
PP-8272 Update PSP switch content

stephencdaly:
- Update content to finalised version
- Move "Switch PSP" item in settings nav below "Your PSP"


SandorArpa (approved):
LOKTM

PR #2842, opened 2021-07-13
PP-8218 Switch PSP to Stripe tasks

sfount:
Namespace Stripe account info pages by credential ID. Update pages to work contextually in the go live and switching PSP flows.
- add support for Stripe steps in the PSP switching page
- add stripe specific content for switching page
- updates task list item to support showing `completed` for items that
shouldn't be retry-able (disabled but completed items)
- add all stripe steps and update tests to work with new structure

Screenshot 2021-07-13 at 10 05 41

Screenshot 2021-07-13 at 10 06 01

### DIFF
--- a/app/controllers/stripe-setup/bank-details/post.controller.js
+++ b/app/controllers/stripe-setup/bank-details/post.controller.js
@@ -3,7 +3,7 @@
 const lodash = require('lodash')
 
 const { response } = require('../../../utils/response')
-const { isSwitchingCredentialsRoute, getSwitchingCredentialIfExists } = require('../../../utils/credentials')
+const { isSwitchingCredentialsRoute, getSwitchingCredential } = require('../../../utils/credentials')
 const bankDetailsValidations = require('./bank-details-validations')
 const { updateBankAccount } = require('../../../services/clients/stripe/stripe.client')
 const { ConnectorClient } = require('../../../services/clients/connector.client')
@@ -17,7 +17,7 @@ const ACCOUNT_NUMBER_FIELD = 'account-number'
 const SORT_CODE_FIELD = 'sort-code'
 
 module.exports = async (req, res, next) => {
-  const switchingToCredentials = isSwitchingCredentialsRoute(req)
+  const isSwitchingCredentials = isSwitchingCredentialsRoute(req)
   const stripeAccountSetup = req.account.connectorGatewayAccountStripeProgress
   if (!stripeAccountSetup) {
     return next(new Error('Stripe setup progress is not available on request'))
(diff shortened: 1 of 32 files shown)


app/controllers/stripe-setup/bank-details/post.controller.js
```diff
@@ -35,20 +37,34 @@ module.exports = async (req, res, next) => {
...
     })
   }
 
   try {
-    const stripeAccount = await connector.getStripeAccount(req.account.gateway_account_id, req.correlationId)
+    const switchingCredential = getSwitchingCredentialIfExists(req.account)
+    let stripeAccountId
 
-    await updateBankAccount(stripeAccount.stripeAccountId, {
+    if (switchingToCredentials) {
```
stephencdaly, line 49: `switchingToCredentials` is quite similar to `switchingCredential`
maybe renaming to something like `isSwitchingCredentials` would improve readability

app/controllers/stripe-setup/bank-details/post.controller.js
```diff
@@ -35,20 +37,34 @@ module.exports = async (req, res, next) => {
...
       accountNumber: rawAccountNumber,
       sortCode: rawSortCode,
+      switchingToCredentials,
       errors
     })
   }
 
   try {
-    const stripeAccount = await connector.getStripeAccount(req.account.gateway_account_id, req.correlationId)
+    const switchingCredential = getSwitchingCredentialIfExists(req.account)
```
stephencdaly, line 46: Moving this variable into the scope of the if block might make it clearer when it's relevant. 

Also should we use the variation of this method that throws an exception if the switching credential does not exist - `getSwitchingCredential`, as we would expect it to if on the switching page?

app/views/switch-psp/switch-psp.njk
```diff
@@ -102,6 +102,11 @@
           formatAccountPathsFor(routes.account.switchPSP.stripeSetup.companyNumber, currentGatewayAccount.external_id, targetCredential.external_id),
           taskList.ENTER_COMPANY_NUMBER
         ) }}
+        {{ taskListItem(
+          "Make a live payment to test your Worldpay PSP",
```
stephencdaly, line 106: "Stripe"

stephencdaly (changes requested):
Looks good, just a copy-paste error and some optional minor suggestions to improve readability.

app/views/switch-psp/switch-psp.njk
sfount: 🤦

app/controllers/stripe-setup/bank-details/post.controller.js
sfount: Really good catch -- I hadn't picked up on that but using the non wrapped method will make any edge cases much clearer

stephencdaly (approved):
👍

PR #2877, opened 2021-08-04
PP-8296 Improve view payment links page for view-only users

stephencdaly:
- Show the "Payment links" option in the account navigation for view-only users and make it link to the "View payment links" page.
- Hide the left hand nav for view-only users as they are not able to create payment links.
- Hide the text about adding metadata for view-only users.
- Move the warning text for view only users below the h1.

Screenshot 2021-08-04 at 16 27 53


SandorArpa (approved):
LOKTM

PR #2881, opened 2021-08-05
PP-8296 Add access control for accepting invite

stephencdaly:
Previously, for invites to invite an existing user to a service, the invite was accepted regardless of whether the invite was for the logged in user.

It was also possible to view the added to service success message for any service by manipulating the URL.

Fix this by:
- Requiring login before an invite for an existing user is completed. Check that the invite is for the logged in user before completing.
- If the invite is not for the user, redirect to "My services" without completing the invite, and show no message to the user.
- Don't pass the service id to display the success message in a query parameter for the redirect to "My services". Instead, pass it in the cookie using `req.flash`.

Other improvements:
- Use the GOV.UK frontend notification component for the success message
- Replace the functional tests for this controller with unit tests


app/controllers/register-user.controller.js
```diff
@@ -40,16 +41,24 @@ const showRegistration = function showRegistration (req, res, next) {
...
-  if (!sessionData || !sessionData.code) {
+  if (!sessionData || !sessionData.code || !sessionData.email) {
     return next(new RegistrationSessionMissingError())
   }
 
   const inviteCode = sessionData.code
   const correlationId = req.correlationId
 
+  if (sessionData.email !== req.user.email) {
+    logger.info('Attempt to accept invite for a diiferent user', {
```
kbottla, line 52:
```suggestion
    logger.info('Attempt to accept invite for a different user', {
```

kbottla (approved):
LGTM!

PR #2890, opened 2021-08-11
PP-8438 Associate service ID when creating default test card account

sfount:
Test card gateway accounts are created by default when a new service is
created, this happens:
- when an existing user adds a new service
- when a new user signs up for Pay, either themselves or following an
invite link

We are now associating connector integrations with services rather than
storing the internal IDs of card connector gateway accounts in the
service object.

- send `service_id` through with all calls to create card gateway
account
- remove `gateway_account_ids` attribute from calls to create service
(backend resources set this as optional, nothing provided will still
result in a created service)
- change order of operations so that a service is created first,
followed by the test card gateway account
- pass the service id correctly to the call to create a default card
gateway account

In order to support the existing behaviour until the data migration is
carried out:
- make a separate call to admin users to associate the internal card
gateway account ID with the service

### DIFF
--- a/app/services/service.service.js
+++ b/app/services/service.service.js
@@ -54,7 +54,7 @@ async function createService (serviceName, serviceNameCy, user, correlationId) {
   if (!serviceNameCy) serviceNameCy = ''
 
   const service = await adminUsersClient.createService(serviceName, serviceNameCy, correlationId)
-  logger.info('New service created for existing user')
+  logger.info('New service added by existing user')
 
   const gatewayAccount = await connectorClient.createGatewayAccount('sandbox', 'test', serviceName, null, service.externalId, correlationId)
   logger.info('New test card gateway account registered with service')
--- a/app/services/service.service.js
+++ b/app/services/service.service.js
@@ -53,9 +53,16 @@ async function createService (serviceName, serviceNameCy, user, correlationId) {
   if (!serviceName) serviceName = 'System Generated'
   if (!serviceNameCy) serviceNameCy = ''
 
-  const gatewayAccount = await connectorClient.createGatewayAccount('sandbox', 'test', serviceName, null, correlationId)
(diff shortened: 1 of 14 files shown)


stephencdaly (dismissed):
Looks good to me, minor comment about the connector client method but nothing to stop it going in.

app/services/service.service.js
```diff
@@ -53,9 +53,16 @@ async function createService (serviceName, serviceNameCy, user, correlationId) {
...
+  const service = await adminUsersClient.createService(serviceName, serviceNameCy, correlationId)
+  logger.info('New service created for existing user')
+
+  const gatewayAccount = await connectorClient.createGatewayAccount('sandbox', 'test', serviceName, null, service.externalId, correlationId)
+  logger.info('New test card gateway account registered with service')
+
+  // @TODO(sfount) PP-8438 support existing method of associating services with internal card accounts, this should be
+  //               removed once connector integration indexed by services have been migrated
+  await adminUsersClient.addGatewayAccountsToService(service.externalId, [ gatewayAccount.gateway_account_id ])
+  logger.info('Service associated with internal gateway account ID with legacy mapping')
```
stephencdaly, line 65: Is it worth adding a small unit test for this, or do we think it's already covered?

sfount: Yeah good question, sad paths would probably make more sense as unit tests but it's very well covered by the integration/ cypress tests which require the appropriate mocks/ stubbings for the code to progress (and some make explicit assertions that the stub has been fulfilled)

sfount:
@stephencdaly thanks for the feedback and good catch re: dashboards. I've reverted to using the existing log lines to keep it simple and have added logging context where you flagged that had previously been broken
https://github.com/alphagov/pay-selfservice/pull/2890/commits/360622695e866c93dd7bfe2178379be7ec20ad64

PR #2905, opened 2021-08-24
PP-8445 Make settings redirects consistent

stephencdaly:
Previously, some settings pages would stay on the same page and display an error message when submitted, and others would redirect back to the main settings page.

Make this consistent so that:
- We always redirect to the main settings page when setting is saved
- We always display a success message

Additionally, have refactored email-notifications.controller.js to adhere to our style guide.


SandorArpa (approved):
LOKTM

PR #2920, opened 2021-09-08
PP-8555 Pass httpAgent to Stripe constructor

kbottla:
## WHAT
- Pass httpAgent to Stripe constructor instead of using deprecated `setHttpAgent` method
- Also sets `timeout` to retrieve account details to 1 second which was originally intended rather than 10 seconds

### DIFF
--- a/app/services/clients/stripe/stripe.client.js
+++ b/app/services/clients/stripe/stripe.client.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const stripe = require('stripe')(process.env.STRIPE_ACCOUNT_API_KEY)
 const ProxyAgent = require('https-proxy-agent')
 
 const StripeBankAccount = require('./StripeBankAccount.class')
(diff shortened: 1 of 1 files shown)


SandorArpa (approved):
LGTM

app/services/clients/stripe/stripe.client.js
```diff
@@ -46,7 +51,7 @@ module.exports = {
 
   retrieveAccountDetails: function (stripeAccountId) {
     return stripe.accounts.retrieve(stripeAccountId, {
```
sfount, line 53: The interface for this method is `(id, option, config)`, this is still passing a configuration into the request options

kbottla: Stripe lib seems to have an overloaded method without account params (ie. `(id, options)`). Not sure though why it didn't work.

https://github.com/stripe/stripe-node/blob/master/types/2020-08-27/Accounts.d.ts#L3046-L3057

```
  /**
   * Retrieves the details of an account.
   */
  retrieve(
    id: string,
    params?: AccountRetrieveParams,
    options?: RequestOptions
  ): Promise<Stripe.Response<Stripe.Account>>;
  retrieve(
    id: string,
    options?: RequestOptions
  ): Promise<Stripe.Response<Stripe.Account>>;
```

sfount: yeah you're right, briefly looking at the interface you've linked and the code looks like this should be respected

PR #2928, opened 2021-09-14
PP-8571 Fix print button issues and update layout

iqbalgds:
- Update `print-button.js`
  - Makes it correctly select the Print button
  - Update the print button styling.
- Update the layout of the page
  - Change download heading to `Download PDF version`
  - Change the read online heading to `HTML Version: <contract name>`


kbottla (approved):
LGTM!

PR #2943, opened 2021-09-27
PP-8565 Use referer on headers to get hostname

kbottla:
## WHAT
- Use `req.headers.referer` to get hostname of selfservice to provide as the return URL for payment. `req.headers.origin` is not supported on browsers like IE or Edge browsers (Before Edge 79, origin header was not sent with POST requests - see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin).


app/controllers/switch-psp/verify-psp-integration.controller.js
```diff
@@ -34,7 +39,7 @@ async function startPaymentJourney (req, res, next) {
       payment_provider: targetCredential.payment_provider,
       description: 'Live payment to verify new PSP',
       reference: 'VERIFY_PSP_INTEGRATION',
-      return_url: urljoin(req.headers && req.headers.origin, formatAccountPathsFor(paths.account.switchPSP.receiveVerifyPSPIntegrationPayment, req.account.external_id))
+      return_url: urljoin(req.headers && getHost(req.headers.referer), formatAccountPathsFor(paths.account.switchPSP.receiveVerifyPSPIntegrationPayment, req.account.external_id))
```
sfount, line 42: Another option we might consider here is using the self service base URL which should be templated into the environment, that would avoid any ambiguity with browser header support/ extensions messing or spoofing headers

PR #2945, opened 2021-09-28
PP-8715 Update authenticator app links

alexbishop1:
Update the authenticator app links (shown to a user on the page that lets them choose their sign-in method) to be more helpful.

The Duo homepage is very enterprisey, so link instead to the specific page for the Duo Mobile app, which still is not aimed at end users but a least has links to download the app.

Update the Microsoft Authenticator link (the old link redirects to the new one).


gidsg (approved):
👍 🔗

PR #3012, opened 2021-10-29
PP-8498 Create/ update Webhook page

sfount:
Controller, paths and routes to connect the create and update routes.

Use the same template with an `isEditing` flag to reduce boilerplate. Explicit controllers for creating and updating call separate backend endpoints.

- bump pay js commons dependency to bring in new Webhooks constants definitions
- add simple service method to normalise frontend checkbox form
formatting to always send an array of subscriptions to the backend

This pull request doesn't do any validation on the create/ update inputs. If this should be done as a one off (following a similar example to other admin tool pages) or used as the first instance of a shared forms validation library should be discussed.

Screenshot 2021-10-29 at 16 45 58

Screenshot 2021-10-29 at 16 45 21

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -448,10 +448,10 @@ module.exports.bind = function (app) {
   account.get(stripe.addPspAccountDetails, permission('stripe-account-details:update'), restrictToStripeAccountContext, stripeSetupAddPspAccountDetailsController.get)
 
   futureAccountStrategy.get(webhooks.index, permission('webhooks:read'), webhooksController.listWebhooksPage)
-  futureAccountStrategy.get(webhooks.create, permission('webhooks:read'), webhooksController.createWebhookPage)
-  futureAccountStrategy.post(webhooks.create, permission('webhooks:read'), webhooksController.createWebhook)
-  futureAccountStrategy.get(webhooks.update, permission('webhooks:read'), webhooksController.updateWebhookPage)
-  futureAccountStrategy.post(webhooks.update, permission('webhooks:read'), webhooksController.updateWebhook)
+  futureAccountStrategy.get(webhooks.create, permission('webhooks:update'), webhooksController.createWebhookPage)
+  futureAccountStrategy.post(webhooks.create, permission('webhooks:update'), webhooksController.createWebhook)
+  futureAccountStrategy.get(webhooks.update, permission('webhooks:update'), webhooksController.updateWebhookPage)
+  futureAccountStrategy.post(webhooks.update, permission('webhooks:update'), webhooksController.updateWebhook)
 
   app.use(paths.account.root, account)
   app.use(paths.service.root, service)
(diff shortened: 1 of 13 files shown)


app/routes.js
```diff
@@ -439,6 +439,10 @@ module.exports.bind = function (app) {
   account.get(stripe.addPspAccountDetails, permission('stripe-account-details:update'), restrictToStripeAccountContext, stripeSetupAddPspAccountDetailsController.get)
 
   futureAccountStrategy.get(webhooks.index, permission('webhooks:read'), webhooksController.listWebhooksPage)
+  futureAccountStrategy.get(webhooks.create, permission('webhooks:read'), webhooksController.createWebhookPage)
+  futureAccountStrategy.post(webhooks.create, permission('webhooks:read'), webhooksController.createWebhook)
+  futureAccountStrategy.get(webhooks.update, permission('webhooks:read'), webhooksController.updateWebhookPage)
+  futureAccountStrategy.post(webhooks.update, permission('webhooks:read'), webhooksController.updateWebhook)
```
gidsg, line 445: Do we need a different permission for these routes? As `read` implies read-only access?

app/views/webhooks/edit.njk
```diff
@@ -0,0 +1,66 @@
+{% extends "layout.njk" %}
+
+{% set heading = "Update your Webhook details" if isEditing else "Create a new Webhook" %}
+{% set submitURL = formatFutureStrategyAccountPathsFor(routes.futureAccountStrategy.webhooks.update, currentGatewayAccount.type, currentGatewayAccount.service_id, currentGatewayAccount.external_id, webhook.external_id) if isEditing else formatFutureStrategyAccountPathsFor(routes.futureAccountStrategy.webhooks.create, currentGatewayAccount.type, currentGatewayAccount.service_id, currentGatewayAccount.external_id) %}
+
+{% block pageTitle %}
+  {{ heading }} - {{currentService.name}} {{currentGatewayAccount.full_type}} - GOV.UK Pay
```
gidsg, line 7: Can we get `{{currentGatewayAccount.full_type}}` from the service rather than the gateway account?

gidsg:
Should we have tests for the update (PATCH) request, as well as the create one?

test/cypress/integration/webhooks/webhooks.cy.test.js
```diff
@@ -32,4 +33,36 @@ describe('Webhooks', () => {
 
     cy.get('[data-webhook-entry]').should('have.length', 1)
   })
+
+  it('should create a webhook with a valid properties', () => {
```
gidsg, line 37:
```suggestion
  it('should create a webhook with valid properties', () => {
```
tiny typo

gidsg (approved):
Looks good, aside from minor comment that can be fixed in subsequent PR.

PR #3019, opened 2021-11-01
PP-8801 Add tasks for Stripe KYC

kbottla:
## WHAT
- Add new tasks list to Your PSP - Stripe page

image

## Responsible person complete
image

## Organisation URL complete
image

### All tasks complete
Won't list tasks


stephencdaly (approved):
Changes look good.

Do we have an issue here that although we aren't showing the "Your PSP - Stripe" link in the settings nav bar, if the user completes the tasks and then tries to click the back button they'll see a blank page? Will this be changed in a future PR to display some content before we make this feature available?

Not a question that needs to be answered before this PR goes through though, as the page won't be linked to for anyone yet.

kbottla:
Thanks @stephencdaly. Yeah blank page won't be very useful here. Content is being reviewed and possibly we have some here or at least show all tasks as completed (if this is page is not available for all services)

PR #3020, opened 2021-11-01
PP-8609 - Rename Policy Page files to stop using HTML in the name and references

iqbalgds:
- Want to refer to everything as just `policy` and no longer use `html`.
  The policy pages contain both the `download link` and the `html` version.
  As this makes it clearer to understand what the files and references refer to.
- Rename `policy-html.njk` to `policy.njk`
  - Update refenrces to this file.
- Update paths.js,routes.js to use new variable name - `policyPage`
- Update the controller name to just `policy.controller`
- Removed all references to `html`


kbottla (approved):
LGTM!

PR #3023, opened 2021-11-02
PP-8801 Add cypress tests for KYC tasks

kbottla:
## WHAT
- Added cypress tests for KYC tasks
- Removed functional tests for KYC tasks as these are now covered by cypress tests


SandorArpa (approved):
LOKTM

PR #3025, opened 2021-11-02
PP-8519 Google Analytics - set cookie domain.

iqbalgds:
- Fix bug so that when the Google Analytic cookies are created -
  we explicitly set a domain.
- Update the `ga('create'...)` to use the cookieFunctions.getDomainCookie
  method.


kbottla (approved):
LGTM!

kbottla (approved):

PR #3028, opened 2021-11-04
BAU Fix custom email paragraph validation length

sfount:
Client side validation was refined and improved in
https://github.com/alphagov/pay-selfservice/pull/2966.

It looks like the maximum paragraph length for custom emails was set in
error, revert it to its previous value.


alexbishop1 (approved):
A literal 10x improvement.

The database column that stores the custom paragraph (`template_body` in `email_notifications` in connector) is of type `TEXT` (which is practically unlimited in terms of characters) and there does not seem to be any validation in connector that restricts the number of characters.

PR #3035, opened 2021-11-08
PP-8499 Webhook detail page

sfount:
* adds template and controller for webhooks detail page
  * conditionally display data that may not be supported by the backend, show a degraded message if not available
* updates existing create/ update flow to work with the detail page

Depends on https://github.com/alphagov/pay-selfservice/pull/3012

Screenshot 2021-11-23 at 11 30 02

### DIFF
--- a/app/views/webhooks/detail.njk
+++ b/app/views/webhooks/detail.njk
@@ -1,0 +1,93 @@
+{% extends "layout.njk" %}
+
+{% block pageTitle %}
+  Webhook {{ webhook.callback_url }} - {{currentService.name}} {{ humanReadableEnvironment }} - GOV.UK Pay
+{% endblock %}
+
+{% block side_navigation %}
+  {% include "includes/side-navigation.njk" %}
+{% endblock %}
+
+{% block mainContent %}
+<div class="govuk-grid-column-two-thirds">
+  <h1 class="govuk-heading-l page-title" title="{{ webhook.callback_url }}">
+    {{ webhook.callback_url | truncate(40) }}
+    <strong class="govuk-tag govuk-tag--blue">{{ webhook.status }}</strong>
+  </h1>
+
+  <p class="govuk-body">{{ webhook.description }}</p>
+
+  <p class="govuk-body">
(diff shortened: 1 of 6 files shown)


app/views/webhooks/detail.njk
```diff
@@ -0,0 +1,93 @@
...
+{% block mainContent %}
+<div class="govuk-grid-column-two-thirds">
+  <h1 class="govuk-heading-l page-title" title="{{ webhook.callback_url }}">
+    {{ webhook.callback_url | truncate(40) }}
+    <strong class="govuk-tag govuk-tag--blue">{{ webhook.status }}</strong>
+  </h1>
+
+  <p class="govuk-body">{{ webhook.description }}</p>
+
+  <p class="govuk-body">
```
gidsg, line 20: Not sure what's in the design here, but seems like this is unlabelled. Should it have a heading like `Payment events`?

test/cypress/integration/webhooks/webhooks.cy.test.js
```diff
@@ -50,14 +50,28 @@ describe('Webhooks', () => {
     cy.get('button').contains('Create Webhook').click()
   })
 
+  it('should display a valid webhooks details', () => {
```
gidsg, line 53:
```suggestion
  it('should display a valid webhooks details page', () => {
```

app/views/webhooks/edit.njk
```diff
@@ -14,9 +15,9 @@
 {% block mainContent %}
 <div class="govuk-grid-column-two-thirds">
   {{ govukBackLink({
-    text: "Back to Webhooks",
+    text: "Back to Webhook" if isEditing else "Back to Webhooks",
```
gidsg, line 18:
```suggestion
    text: "Back to Webhook details" if isEditing else "Back to Webhooks",
```
Maybe makes more sense, as the user is kind of on the webhook already, if they are editing it? Not sure though!

gidsg (dismissed):
Looks good to me aside from minor comments 👍

PR #3036, opened 2021-11-08
PP-8804 Add StripeDirector class

kbottla:
## WHAT
- Added new class for StripeDirector which will be used when adding new director for stripe accounts.
   - Separate class instead of using StripePerson to keep classes simple. 
        - Other options could be i) to use StripePerson for both responsible person and director ii) define base class and inherit as needed.
- Added new stripe client method to create and update director.

### DIFF
--- a/app/services/clients/stripe/StripeDirector.class.js
+++ b/app/services/clients/stripe/StripeDirector.class.js
@@ -1,0 +1,49 @@
+'use strict'
+
+const Joi = require('joi')
+
+const schema = {
+  first_name: Joi.string().required(),
+  last_name: Joi.string().required(),
+  email: Joi.string().required(),
+  dob_day: Joi.number().integer().strict().min(1).max(31),
+  dob_month: Joi.number().integer().strict().min(1).max(12),
+  dob_year: Joi.number().integer().strict().min(1900).max(2100),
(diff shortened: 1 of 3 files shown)


app/services/clients/stripe/StripeDirector.class.js
```diff
@@ -0,0 +1,49 @@
...
+
+const Joi = require('joi')
+
+const schema = {
+  first_name: Joi.string().required(),
+  last_name: Joi.string().required(),
+  email: Joi.string().required(),
+  dob_day: Joi.number().integer().strict().min(1).max(31),
+  dob_month: Joi.number().integer().strict().min(1).max(12),
+  dob_year: Joi.number().integer().strict().min(1000).max(9999),
```
SandorArpa, line 11: Shouldn't it be min(1900).max(2020) or something like that?

SandorArpa (changes requested):

app/services/clients/stripe/StripeDirector.class.js
kbottla: yeah good point! updated to `min(1900).max(2100)`. DOB validation should prevent if year is in future.

SandorArpa (approved):
LOKTM

PR #3039, opened 2021-11-09
PP-8804 Stripe setup: Add controller for director

kbottla:
## WHAT
- Added new controller (GET) for director page
- Unit and cypress test for new page
- Updated routes, paths and navigation for director page

image


stephencdaly (approved):
Looks good to me! The ordering and width of the email input doesn't look to match the prototype, but has the design been changed since?

kbottla:
Thanks @stephencdaly. Good spots! Didn't realise email has different width. Will update it in next PR and also the order

PR #3040, opened 2021-11-09
PP-8832 make vat number optional

SandorArpa:
## WHAT
- At present, VAT number is mandatory for services onboarding with Stripe. Services will have enter a valid VAT number and if they don’t have one, we need to override tasks (manual update to DB). VAT is not a mandatory field as per Stripe. Thus make it optional in onboarding journey.
- update cypress tests

### DIFF
--- a/app/controllers/stripe-setup/vat-number/post.controller.test.js
+++ b/app/controllers/stripe-setup/vat-number/post.controller.test.js
@@ -6,7 +6,8 @@ const paths = require('../../../paths')
 
 describe('VAT number POST controller', () => {
   const postBody = {
-    'vat-number': 'GB999999973'
+    'vat-number': 'GB999999973',
+    'vat-number-declaration': 'true'
   }
 
   let req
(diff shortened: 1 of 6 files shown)


app/controllers/stripe-setup/vat-number/post.controller.test.js
```diff
@@ -71,6 +72,22 @@ describe('VAT number POST controller', () => {
     sinon.assert.calledWith(res.redirect, 303, `/account/a-valid-external-id${paths.account.stripe.addPspAccountDetails}`)
   })
 
+  it('should call not Stripe, then connector, then redirect to add psp account details route', async function () {
```
nimalank7, line 75: Should this be 'should not call Stripe when vat number not provided'?

nimalank7 (changes requested):
Minor comment regarding test name but apart from that fine.

nimalank7 (approved):
LGTM!

PR #3041, opened 2021-11-09
PP-8831 Send phone number to stripe with international code

stephencdaly:
Stripe rejects phone numbers without an international code. Format the phone number entered for the responsible person to include the UK dialling code if no international dialling code is provided.

Have tested this locally, connecting to the Stripe test environment and the phone number formatted to include the international dialling code is accepted.

## Notes 

`telephone-number-validation.js` was renamed to `telephone-number-utils.js` and a function was added to format the phone number with the dialling code. The validation method and tests are unchanged, but git has not recognised this as a file rename.


kbottla (approved):
LGTM!

PR #3042, opened 2021-11-11
PP-8804 Stripe setup - Update director controller for POST

kbottla:
## WHAT
- Updated director controller for POST
  - POST is context aware (i.e., new services onboarding, kyc, switch PSP)
- Will raise separate PR for cypress tests

### DIFF
--- a/app/controllers/stripe-setup/director/post.controller.js
+++ b/app/controllers/stripe-setup/director/post.controller.js
@@ -1,0 +1,168 @@
+'use strict'
+
+const lodash = require('lodash')
+
+const logger = require('../../../utils/logger')(__filename)
+const paths = require('../../../paths')
+const formatAccountPathsFor = require('../../../utils/format-account-paths-for')
+const { isSwitchingCredentialsRoute, isAdditionalKycDataRoute, getCurrentCredential, getSwitchingCredential } = require('../../../utils/credentials')
+const { response } = require('../../../utils/response')
+const { validateMandatoryField, validateEmail } = require('../../../utils/validation/server-side-form-validations')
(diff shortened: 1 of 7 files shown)


app/controllers/stripe-setup/director/post.controller.js
```diff
@@ -0,0 +1,168 @@
...
+      dobMonth: formFields[DOB_MONTH_FIELD],
+      dobYear: formFields[DOB_YEAR_FIELD]
+    }
+
+    pageData['errors'] = errors
+
+    return response(req, res, 'stripe-setup/director/index', {
+      ...pageData, isSwitchingCredentials, collectingAdditionalKycData, currentCredential
+    })
+  } else {
```
stephencdaly, line 74: As we return if there are errors we could remove the `else` to reduce nesting. Not something that needs to be fixed, just a personal preference for readability

app/controllers/stripe-setup/director/post.controller.js
```diff
@@ -0,0 +1,168 @@
...
+      } else {
+        return res.redirect(303, formatAccountPathsFor(paths.account.stripe.addPspAccountDetails, req.account && req.account.external_id))
+      }
+    } catch (err) {
+      next(err)
+    }
+  }
+}
+
+async function getStripeAccountId (account, isSwitchingCredentials, correlationId) {
```
stephencdaly, line 109: Could this be shared as a utility functions, as I'm sure we do this in a lot of places?

stephencdaly (approved):
Looks good to me.

app/controllers/stripe-setup/director/post.controller.js
kbottla: yes, going to introduce utility and refactor separately.

PR #3045, opened 2021-11-11
PP-8804 Add Stripe setup utility

kbottla:
## WHAT
- Added a new Stripe setup utility which has common functions used across stripe setup controllers (responsible person, director, bank details and so on).
  Some could possibly (like `getFormFields`) be used in all selfservice controllers.


SandorArpa (approved):
LOKTM

nimalank7 (approved):
LGTM!

PR #3048, opened 2021-11-11
PP-8804 Stripe setup: Use utility methods

kbottla:
## WHAT
- Updated all stripe setup controllers (POST method) to use utility methods to get Stripe connect account ID and some validation.


stephencdaly (approved):
LGTM!

PR #3052, opened 2021-11-12
PP-8830 Update pact test to include URL

stephencdaly:
Update the PATCH service update patch request to include updating the
merchant_details URL.


SandorArpa (approved):
LOKTM

PR #3053, opened 2021-11-12
PP-8802 - Add KYC Organisation URL feature

iqbalgds:
- Add new route for Org URL page
- Add new Nunjucks file
- Update content
  
image

### DIFF
--- a/app/views/kyc/organisation-url.njk
+++ b/app/views/kyc/organisation-url.njk
@@ -1,0 +1,47 @@
+{% extends "../layout.njk" %}
+
+{% block pageTitle %}
+  Add organisation website address - {{ currentService.name }} - GOV.UK Pay
+{% endblock %}
+
+{% block side_navigation %}
+  {% include "includes/side-navigation.njk" %}
+{% endblock %}
+
+{% block mainContent %}
+
+<div class="govuk-grid-column-two-thirds">
+
+  {{ govukBackLink({
+    text: "Back",
+    classes: "govuk-!-margin-top-0",
+    href: formatAccountPathsFor(routes.account.yourPsp.index, currentGatewayAccount.external_id, credential.external_id)
+  }) }}
+
+  <h1 class="govuk-heading-l govuk-!-margin-bottom-6">
+    <label for="organisation-url">Add organisation website address</label>
+  </h1>
+
+  <form id="organisation-url-form" method="post" novalidate>
+    <input id="csrf" name="csrfToken" type="hidden" value="{{ csrf }}"/>
+
+    {{ govukInput({
+      name: "organisation-url",
+      id: "organisation-url",
+      classes: "govuk-input--width-30",
+      value: organisationUrl,
+      hint: {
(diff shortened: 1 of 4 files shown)


app/services/clients/adminusers.client.js
```diff
@@ -673,6 +673,26 @@ module.exports = function (clientOptions = {}) {
     )
   }
 
+  const updateMerchantUrl = (serviceExternalId, merchantUrl, correlationId) => {
```
stephencdaly, line 676: We could use the `updateService` method here https://github.com/alphagov/pay-selfservice/blob/master/app/services/service.service.js#L44 rather than needing to add an additional client method.
This was introduced more recently, so not all places where we patch a service haven't been updated to use this, but using it would reduce the amount of code required.

app/views/kyc/organisation-url.njk
```diff
@@ -0,0 +1,46 @@
...
+
+<div class="govuk-grid-column-two-thirds">
+
+  {{ govukBackLink({
+    text: "Back to Your PSP",
+    classes: "govuk-!-margin-top-0",
+    href: formatAccountPathsFor(routes.account.yourPsp.index, currentGatewayAccount.external_id, credential.external_id)
+  }) }}
+
+  <h1 class="govuk-heading-l govuk-!-margin-bottom-6"><label for="vat-number">What is your organisation’s URL?</label></h1>
```
stephencdaly, line 17: It looks like this page might have been missed in the content review. There is a separate story to make sure all the pages are using the finalised content at the end, but might want to raise with Justin that this page needs reviewing. The content document is here https://docs.google.com/document/d/1-_ohNR4DjT26alRaIAdu2cR5Rqq9TEbhGYJEkoFlQkA/edit#heading=h.l61akk9fle6h

app/services/clients/adminusers.client.js
iqbalgds: Thanks - I have removed the `updateMerchantUrl` method.

app/views/kyc/organisation-url.njk
iqbalgds: The content for this page is in that Google doc - under `Add Organisation detail`.
I have updated the content to the laters version as per that Google doc.

app/utils/display-converter.js
```diff
@@ -34,7 +34,8 @@ const hideServiceNavTemplates = [
   'request-to-go-live/organisation-name',
   'request-psp-test-account/index',
   'two-factor-auth/index',
-  'two-factor-auth/configure'
+  'two-factor-auth/configure',
+  'kyc/organisation-url'
```
stephencdaly, line 38: I don't think we want to hide the service nav for this page? The add kyc details is done from within the context of a gateway account, so think it should be shown

iqbalgds: Good spot - I also added the side menu, as the mockup has this included.

stephencdaly (approved):
LGTM

PR #3062, opened 2021-11-15
PP-8830 Update content for organisation details page

stephencdaly:
Update the page to use the latest content. This changes where the
fieldset is - to just contain the address, with the telephone number
and website address URLs separate.

Screenshot 2021-11-15 at 13 13 40


nimalank7 (approved):
LGTM!

PR #3064, opened 2021-11-16
PP-8830 Reuse same controller for both organisation details pages

stephencdaly:
There are two pages where users can enter organisation details:
- A page accessible from "My services"
- As part of the request to go live journey

The main difference between these pages is that for the request to go
live journey, the organisation name is collected on a separate page.

The other page has been left behind when content changes have been made,
and having them separate allows them to keep getting out of sync.

Combine the pages to use the same controllers and view, but for the
page that is accessed from "My services" also collect the organisation
name.

Did some additional tidy up - renamed some code references to merchant
details to organisation details to reflect the name of the page. Removed
the functional tests and replaced them with Cypress tests.

## Review notes

Selecting "Hide whitespace changes" will make this a bit easier to review

Screenshot 2021-11-16 at 14 18 30

## How it looks 

Screenshot 2021-11-16 at 15 01 18


kbottla (approved):
LGTM!

PR #3072, opened 2021-11-18
PP-8902 some correcting changes

SandorArpa:
## WHAT
- These changes are some things that we missed in the first round and/or later additions.


kbottla (approved):
LGTM!

PR #3073, opened 2021-11-18
PP-8803 Show all responsible person fields when change is clicked

stephencdaly:
When the "Change" button is clicked next to the name of the existing responsible person on the KYC additional information page, reload the responsible person page with all fields present.

Screenshot 2021-11-18 at 12 14 51


kbottla (approved):
LGTM!

PR #3076, opened 2021-11-19
PP-8803 Implement submission of additional responsible person details

stephencdaly:
Create new GET and POST controllers for the page to submit only the
additional details required for KYC (phone and email). Have done this
because attempting to have this logic in the same controllers as
submitting the full responsible details was becoming quite convoluted.

Getting and posting the responsible details for KYC when the "Change"
has been clicked uses the GET and POST controllers for submitting
the responsible details for go-live, as most of the logic is the same
with the exception of setting the stripe setup task in connector.


SandorArpa (approved):
Tidy KISS, LBTM

PR #3077, opened 2021-11-19
PP-8902 fix restricted account banner

SandorArpa:
## WHAT
- fix rendering of Stripe restricted account banner


stephencdaly (approved):
Banner now appears as expected

Screenshot 2021-11-19 at 12 07 40

stephencdaly (approved):

PR #3089, opened 2021-11-22
PP-8928 Improve error summaries

stephencdaly:
For the request to go live and stripe setup pages, improve the error
summaries to be in line with the design system recommendations that the
error messages in the summary should be the same as the inline errors.

Improve inline errors to change the "This field cannot be blank" errors
to reference the field name.

Deleted a few tests that were superfluous.


SandorArpa (approved):
LOKTM

PR #3091, opened 2021-11-22
PP-8928 Improve error messages on choose how to process payments page

stephencdaly:
Use the error summary component for these error messages and display inline error messages.

Removed some stuff from the get controller involving the session which was unused, and inline some functions in the post controller to make it simpler.

## Before
Screenshot 2021-11-22 at 14 19 25

## After

Screenshot 2021-11-22 at 14 17 04
Screenshot 2021-11-22 at 14 17 28


SandorArpa (approved):
LGTM

PR #3092, opened 2021-11-22
PP-8928 Improve error message on agreement page

stephencdaly:
Use the error summary component and display an inline error message.

Screenshot 2021-11-22 at 14 57 31


kbottla (approved):
LGTM!

PR #3096, opened 2021-11-23
PP-8802 Org URL - Update error messages to be consistent with Design System

iqbalgds:
- Update the `organisation-url.njk` template
  - Use the `error-summary.njk` macro so the summary is consistent
    with Design System guidelines.
- Update Cypress test


stephencdaly (approved):
LGTM!

PR #3097, opened 2021-11-23
PP-8835 Switch PSP - Add organisation URL task

kbottla:
## WHAT
- Added organisation URL task to switch PSP tasks
- Moved organisation url validation cypress test (from kyc specific tests as there are short lived) to organisation-url cypress test


stephencdaly (approved):
LGTM!

PR #3098, opened 2021-11-23
PP-8861 Complete KYC when all tasks are finished

stephencdaly:
When the final task for KYC additional data collection is finished, which can be done in any order:
- Update the Stripe account to add the new capabilities as well as the MCC code and product description
- Patch connector to set the `additional_kyc_data` Stripe setup step
- Patch connector to disable the `requires_additional_kyc_data` flag on the gateway account
- Show a success banner saying that all tasks have been completed


SandorArpa (approved):
LOKTM

PR #3099, opened 2021-11-23
PP-8928 Fix link in error summary

stephencdaly:
For radio buttons, there is no ID for the containing fieldset, so error summary messages need to link to the first radio button item. Fix the ID for the error on the VAT number page so it links to the first radio button item, which has a custom ID.


kbottla (approved):
LGTM!

PR #3100, opened 2021-11-23
PP-8928 Capitalise Company registration number

stephencdaly:
This should be capitalised according to content guidance, so capitalise it everywhere it's used.


SandorArpa (approved):
LGTM

PR #3102, opened 2021-11-24
PP-8922 Update Stripe API version to the latest

SandorArpa:
## WHAT
- Update Stripe API version to the latest


iqbalgds (approved):

0atman (commented):
This is the correct version according to stripe's docs https://stripe.com/docs/api/versioning

PR #3103, opened 2021-11-24
PP-8901 `npm install` -> `npm ci`

0atman:
## WHAT
switched `npm install` to `npm ci`, as reported by @heathd, which creates reproducible builds.

## HOW 
`docker build .` observe how `package-lock.json is unmodified`

On my machine, both take 84s to build, so there appears to be no change in build time.


0atman:
The dockerfile being based on Alpine I'm not overly delighted with. However, that's outside of the scope of this PR

dj-maisy (approved):
A quality improvement to the selfservice Dockerfile. 👍

PR #3104, opened 2021-11-24
PP-8861 Show KYC task list after all tasks complete

stephencdaly:
Show the completed task list with a different paragraph above it when all KYC tasks have been completed.

Also continue to show the "Your PSP - Stripe" link in the settings navigation after tasks have been completed.

We disable the `requires_additional_kyc_data` flag after all tasks have been completed, but we set the `additional_kyc_data` Stripe setup task on the gateway account, so look at this to determine whether to show the navigation link.


SandorArpa (approved):
LGTM

PR #3107, opened 2021-11-26
PP-8861 Don't show KYC task list after Stripe switch complete

stephencdaly:
We were showing the KYC task list after switching PSP to Stripe was
complete, as we were loading the Your PSP page.

Only show the task list when the service is providing KYC data or after
they have completed providing KYC data to make sure this doesn't happen.

## Note

Most of the affected lines are just indentation changes, hiding whitespace changes might help
 
This is what was happening:
Screenshot 2021-11-26 at 14 35 23

### DIFF
--- a/app/views/your-psp/_stripe.njk
+++ b/app/views/your-psp/_stripe.njk
@@ -18,46 +18,49 @@
   </li>
 {% endmacro %}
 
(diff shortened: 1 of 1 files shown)


kbottla (approved):
LGTM!

PR #3114, opened 2021-11-30
PP-8861 Complete KYC if responsible person is last task

stephencdaly:
When adding the functionality to enable the new Stripe account capabilities and setting the flag that KYC was complete for the gateway account, the page where the user adds just the email and telephone to the responsible person was missed.

When this page is submitted, check whether the KYC task list is now complete, and if so call the function to do complete the KYC process and show a success banner notifying the user that KYC is complete.


SandorArpa (approved):
LGTM

PR #3115, opened 2021-11-30
PP-8962 Add get controller for government entity document task

kbottla:
## WHAT
- Added get controller for new Stripe onboarding task (Proof of Government Entity document)
- Also added a pact test for equivalent stripe setup task in connector

image

### DIFF
--- a/app/controllers/stripe-setup/government-entity-document/get.controller.js
+++ b/app/controllers/stripe-setup/government-entity-document/get.controller.js
@@ -1,0 +1,25 @@
+'use strict'
+
+const { response } = require('../../../utils/response')
+const { isSwitchingCredentialsRoute, isAdditionalKycDataRoute, getCurrentCredential } = require('../../../utils/credentials')
+const { getAlreadySubmittedErrorPageData } = require('../stripe-setup.util')
+
+module.exports = (req, res, next) => {
+  const isSwitchingCredentials = isSwitchingCredentialsRoute(req)
+  const collectingAdditionalKycData = isAdditionalKycDataRoute(req)
+  const currentCredential = getCurrentCredential(req.account)
+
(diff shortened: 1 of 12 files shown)


app/controllers/stripe-setup/government-entity-document/get.controller.js
```diff
@@ -0,0 +1,25 @@
+'use strict'
+
+const { response } = require('../../../utils/response')
+const { isSwitchingCredentialsRoute, isAdditionalKycDataRoute, getCurrentCredential } = require('../../../utils/credentials')
+const { getAlreadySubmittedErrorPageData } = require('../stripe-setup.util')
+
+module.exports = (req, res, next) => {
```
stephencdaly, line 7: Should name the controller function rather than having it anonymous to adhere to our style guide

test/cypress/integration/stripe-setup/government-entity-document.cy.test.js
```diff
@@ -0,0 +1,145 @@
...
+      cy.visit(governmentEntityDocumentUrl)
+
+      cy.get('h1').should('contain', 'An error occurred')
+      cy.get('#back-link').should('contain', 'Back to dashboard')
+      cy.get('#back-link').should('have.attr', 'href', dashboardUrl)
+      cy.get('#error-message').should('contain', 'You’ve already provided proof of government entity document. Contact GOV.UK Pay support if you need to update it.')
+    })
+  })
+
+  describe('when it is not a Stripe gateway account', () => {
```
stephencdaly, line 93: We could probably think about dropping these edge case tests for errors as they're covered sufficiently by the unit tests you've added. Just to reduce maintenance burden and reduce runtime of Cypress tests marginally.

stephencdaly (approved):
Looks good. Just minor style comment and a comment about Cypress tests, but can be merged.

PR #3116, opened 2021-12-01
PP-8962 Install multer dependency and add Stripe client methods for file upload

kbottla:
## WHAT
- Installed `multer` package which is a expressjs middleware for handling multipart/form-data (to upload file for Stripe task - Proof of Government entity document).
- Added stripe client method to create file for government entity verification on Stripe
- Updated StripeAccount class for the payload required to attach verification document to stripe account.


SandorArpa (approved):
LGTM

PR #3117, opened 2021-12-01
PP-8962 Add post controller for government entity document task

kbottla:
## WHAT
- Added post controller for new Stripe onboarding task (Proof of Government Entity document)
- Starts using csrfToken available on query param as for multipart form we will be specifying csrfToken as the query param
- Use multer middleware to handle multipart/form-data and configured to use in-memory storage by default.

Can't add Cypress tests for full post journey as stripe hardcodes hostname (files.stripe.com) for file upload and cannot be overridden. Unit tests asserts that correct payload is passed to stripe client methods.

### DIFF
--- a/app/controllers/stripe-setup/government-entity-document/post.controller.js
+++ b/app/controllers/stripe-setup/government-entity-document/post.controller.js
@@ -1,0 +1,95 @@
+'use strict'
+
+const lodash = require('lodash')
+
+const logger = require('../../../utils/logger')(__filename)
+const { response } = require('../../../utils/response')
+const { isSwitchingCredentialsRoute, isAdditionalKycDataRoute, getCurrentCredential } = require('../../../utils/credentials')
+const { getStripeAccountId, getAlreadySubmittedErrorPageData, completeKyc } = require('../stripe-setup.util')
+const { uploadFile, updateAccount } = require('../../../services/clients/stripe/stripe.client')
+const { isKycTaskListComplete } = require('../../../controllers/your-psp/kyc-tasks.service')
+const { ConnectorClient } = require('../../../services/clients/connector.client')
+const connector = new ConnectorClient(process.env.CONNECTOR_URL)
+const paths = require('../../../paths')
+const formatAccountPathsFor = require('../../../utils/format-account-paths-for')
+
+const GOVERNMENT_ENTITY_DOCUMENT_FIELD = 'government-entity-document'
+
+async function postGovernmentEntityDocument (req, res, next) {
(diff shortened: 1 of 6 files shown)


app/controllers/stripe-setup/government-entity-document/post.controller.js
```diff
@@ -0,0 +1,95 @@
...
+          req.flash('generic', 'You’ve successfully added all the Know your customer details for this service.')
+        } else {
+          req.flash('generic', 'Document has been submitted successfully')
+        }
+        return res.redirect(303, formatAccountPathsFor(paths.account.yourPsp.index, req.account && req.account.external_id, currentCredential.external_id))
+      }
+
+      return res.redirect(303, formatAccountPathsFor(paths.account.stripe.addPspAccountDetails, req.account && req.account.external_id))
+    } catch (err) {
+      next(err)
```
stephencdaly, line 74: I've tested this locally, and multer breaks the AsyncLocalStorage session we use for logging with useful info such as the gateway_account_id, user_external_id etc.

Getting
```
{
  "logger_name": "/Users/stephendaly/projects/gds/pay-selfservice/app/controllers/stripe-setup/government-entity-document/post.controller.js",
  "message": "Logging from the POST controller",
  "level": "INFO",
  "@timestamp": "2021-12-01T16:18:41.639Z",
  "@version": 1,
  "container": "selfservice"
}
```

When should be getting something like 
```
{
  "logger_name": "/Users/stephendaly/projects/gds/pay-selfservice/app/controllers/stripe-setup/government-entity-document/get.controller.js",
  "message": "Logging from the GET controller",
  "level": "INFO",
  "user_external_id": "136f59fa302843dfa38618169b73cdf4",
  "internal_user": false,
  "gateway_account_id": 10,
  "gateway_account_type": "live",
  "service_external_id": "c7dcda05ae444e118ec1e5b216ce01ae",
  "@timestamp": "2021-12-01T16:18:14.632Z",
  "@version": 1,
  "container": "selfservice"
}
```

This means that in particular, we won't get the `x_request_id` on log lines. I think we can make this better this for this route by surrounding the whole controller with a try/catch block and manually logging an error line with all the additional details we expect before passing the error to next. Then at least if there is an error we can link it back to a request and an account.

stephencdaly (commented):
Code looks good.
Left a comment about how multer breaks logging a bit and how we might resolve it.

stephencdaly (approved):
Looks good, we'll add in a solution for the missing log context in a separate PR

PR #3119, opened 2021-12-01
PP-8961 Add government entity document task to switch PSP and KYC tasks

kbottla:
## WHAT
- Added government entity document task to switch PSP and KYC tasks

### Switch PSP page with new task
image

### Your PSP page
image


nimalank7 (approved):
LGTM!

PR #3120, opened 2021-12-01
PP-8983 Stop setting default value for company number

kbottla:
## WHAT
- We are creating new stripe accounts with capabilities -card_payments, transfers. Companies house registration number is no longer mandatory for accounts with these capabilities. It was mandatory for legacy_payments capabilities and we had to set something to enable stripe accounts.


stephencdaly (approved):
LGTM

PR #3121, opened 2021-12-02
PP-8860 Stripe KYC: Set additional required fields and remove legacy capabilities

kbottla:
## WHAT
- Set responsible person as executive when updating only phone number and email address.
- For stripe accounts switched to new capabilities (card_payments) company phone number is mandatory. We provided organisation phone number at the time of creating stripe account (accounts with legacy_payments capabilities) but this appears in a different place (business_profile.support_phone_number) for accounts switched to new capabilities.
- Also remove legacy_payments capability (if exists on account) once new capabilities (card_payments, transfers) are enabled - stripe api https://stripe.com/docs/api/capabilities/update?lang=node to update capability


SandorArpa (approved):
LGTM

PR #3122, opened 2021-12-03
PP-8860 Fix getting telephone number from merchant details

kbottla:
## WHAT
- Use snake case `telephone_number` as it is not converted to camelcase when transforming service response from adminusers


SandorArpa (approved):
LGTM

PR #3128, opened 2021-12-07
PP-8929 Stripe KYC: stop moving stripe accounts to new capabilities

kbottla:
## WHAT
- As stripe accounts are being restricted (both payments & payouts), stop moving accounts to new capabilities while stripe addresses the issue of providing grace period for services to provide missing info.
- mcc is currently set when switching stripe accounts to new capabilities. But mcc cannot be updated if it is already set on account (some live accounts on Stripe has mcc set). So set mcc only when it is not available on stripe account.

### DIFF
--- a/app/controllers/stripe-setup/stripe-setup.util.js
+++ b/app/controllers/stripe-setup/stripe-setup.util.js
@@ -8,7 +8,7 @@ const connector = new ConnectorClient(process.env.CONNECTOR_URL)
 const { validateDateOfBirth } = require('../../utils/validation/server-side-form-validations')
 const paths = require('../../paths')
 const formatAccountPathsFor = require('../../utils/format-account-paths-for')
-const { listPersons, addNewCapabilities, removeLegacyPaymentsCapability } = require('../../services/clients/stripe/stripe.client')
+const { listPersons, addNewCapabilities, retrieveAccountDetails } = require('../../services/clients/stripe/stripe.client')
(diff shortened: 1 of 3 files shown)


app/controllers/stripe-setup/stripe-setup.util.js
```diff
@@ -72,25 +71,12 @@ async function getExistingResponsiblePersonName (account, isSwitchingCredentials
...
     connector.disableCollectAdditionalKyc(gatewayAccountId, correlationId)
   ])
 
-  const stripeLegacyPayments = lodash.get(stripeAccountResponse, 'capabilities.legacy_payments')
-  if (stripeLegacyPayments === 'active') {
-    await removeLegacyPaymentsCapability(stripeAccountId)
-  }
-
-  logger.info('KYC additional information completed for Stripe account', {
+  logger.info('KYC additional information completed for Stripe account (but not moved to new capabilities)', {
```
stephencdaly, line 79: Do we want to continue adding the additional information - MCC, product description, telephone so that when we or Stripe come to change it, it's only the capabilities that need updating?

stephencdaly (approved):
LGTM!

app/controllers/stripe-setup/stripe-setup.util.js
kbottla: good point! updated. Also included few changes from here https://github.com/alphagov/pay-selfservice/pull/3127

PR #3129, opened 2021-12-07
PP-8929 Wrap multer upload with a Promise

kbottla:
## WHAT
- We are using multer for document upload and noticed request context (gateway_account_id, service id,...) added as part of middleware are missing from the logs. This works sometimes and not sure yet why.
- There is an open issue multer where content is lost #/multer/issues/814 and the work around is to wrap upload in a Promise.

with @stephencdaly, @sandorarpa, @nlsteers, @nimalank7


stephencdaly (approved):

SandorArpa (approved):

PR #3133, opened 2021-12-08
PP-8929 Stripe KYC: enable new capabilities

kbottla:
## WHAT
- Enable new capabilities for stripe accounts when services submits all info (but does not remove legacy payments capabilities). legacy_payments capability will be removed (manually and in batch) after verifying that the stripe accounts do not have any information due.


stephencdaly (approved):
LGTM!

PR #3135, opened 2021-12-08
BAU Fix Welsh service name validation

stephencdaly:
If there were validation errors when adding a Welsh service name, these were not being passed correctly through to the view, and not displayed inline correctly.

This affects adding new services and editing service names.

Remove validation for Welsh service names from the controller for naming a service when registering an account, as the page shown for this does not have an option to add a Welsh name.

Also fix the page for adding a new service so that the checkbox for adding a Welsh name is checked if a Welsh name has been entered but  there are errors.

Screenshot 2021-12-08 at 16 07 12


kbottla (approved):
LGTM!

PR #3142, opened 2021-12-14
PP-8499 Manage Webhook pages

sfount:
Adds pages to manage an individual webhooks, supporting
* accessing a given webhooks signing secret
* deactivating and activating a given webhook

Adds helper methods for accessing singing secret endpoint and service wrapper to toggle status.

Happy path Cypress coverage.

**Signing secret page**

Screenshot 2021-12-14 at 13 16 18

**Toggle status page**

Screenshot 2021-12-14 at 13 16 29

**Detail page (following toggle status action**

Screenshot 2021-12-14 at 13 16 42

### DIFF
--- a/app/views/webhooks/signing_secret.njk
+++ b/app/views/webhooks/signing_secret.njk
@@ -1,0 +1,71 @@
+{% extends "layout.njk" %}
+
+{% block pageTitle %}
+  Webhook signing secret {{ webhook.callback_url }} - {{currentService.name}} {{ humanReadableEnvironment }} - GOV.UK Pay
+{% endblock %}
+
+{% block side_navigation %}
+  {% include "includes/side-navigation.njk" %}
+{% endblock %}
+
+{% block mainContent %}
+<div class="govuk-grid-column-two-thirds">
+  {{ govukBackLink({
+    text: "Back to Webhook",
+    classes: "govuk-!-margin-top-0",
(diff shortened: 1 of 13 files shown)


app/views/webhooks/signing_secret.njk
```diff
@@ -0,0 +1,71 @@
...
+  {{ govukBackLink({
+    text: "Back to Webhook",
+    classes: "govuk-!-margin-top-0",
+    href: formatFutureStrategyAccountPathsFor(routes.futureAccountStrategy.webhooks.detail, currentGatewayAccount.type, currentGatewayAccount.service_id, currentGatewayAccount.external_id, webhook.external_id)
+  }) }}
+
+  <h1 class="govuk-heading-l page-title">Manage signing secret</h1>
+
+  <p class="govuk-body">
+    The contents of Webhooks are encrypted with a signing secret. Only your application should store this secret to unencrypt Webhooks content.
```
gidsg, line 22: This text isn't really accurate. You don't use the signing key to 'unencrypt' the content, you use it to verify the authenticity of the message. Can we at least put a TODO here and maybe we can pick up the content with @NathanD-GDS et al?

sfount: 👍  this is lifted and shifted from some very early days stuff. I'll add it to a list of comments being addressed in the next pass

gidsg (approved):
Looks good to me (aside from minor comment) 🚥 👍

PR #3159, opened 2022-01-05
PP-8989 Stop prefixing VAT numbers with GB

nlsteers:
## WHAT
This PR:
- removes the need for a `GB` prefix when submitting VAT numbers during the Stripe onboarding flow
- changes the validation errors and page wording to align with [HMRC design patterns on taking VAT information from customers](https://design.tax.service.gov.uk/hmrc-design-patterns/vat-registration-number/)
- simplifies some logic
- adds a new RegEx for checking VAT numbers ([RegExr playground](https://regexr.com/6cmtd))

## HOW
Run relevant tests:
```bash
npx mocha app/utils/validation/field-validation-checks.vat-number.test.js
npx mocha app/controllers/stripe-setup/vat-number/vat-number-validations.test.js
npx cypress run --spec 'test/cypress/integration/stripe-setup/vat-number.cy.test.js'
```


app/views/stripe-setup/vat-number/index.njk
```diff
@@ -41,10 +41,10 @@
           name: "vat-number",
           id: "vat-number",
           label: {
-            text: "Enter your organisation's VAT number"
+            text: "VAT registration number"
           },
           hint: {
-            text: "VAT numbers start with ‘GB’, for example ‘GBGD123’"
+            text: "This is 9 numbers, sometimes with ‘GB’ at the start, for example 123456789 or GB123456789. You can find it on your VAT registration certificate."
```
stephencdaly, line 47: I wonder whether this content needs a slight modification, as it seems that the VAT number isn't always 9 numbers. Possibly it should say "This is usually 9 number..." although should ask Nathan about this.

stephencdaly (commented):
Looks good, just a comment about the content.

app/views/stripe-setup/vat-number/index.njk
nlsteers: I used the suggested wording from HMRCs design system re VAT reg numbers, it is marked `experimental` though. Presumably the most common format is `GB?[0-9]{9}`?

stephencdaly: I guess for HMRC, the usual users are businesses so have the 9 character VAT number, whereas we're more likely to be dealing with government departments and health authorities. Don't have any stats as to how often people have VAT numbers with fewer than 9 digits though, as we don't store these.

NathanD-GDS: Agree with @stephencdaly and his tweak. I'd also adjust HMRC's content a bit as well.

Does this work?

`This is usually 9 numbers and may start with `GB`. For example, `123456789` or `GB123456789`. You can find it on your VAT registration certificate.`

NathanD-GDS: My only concern would be services getting confused if their VAT number is fewer than 9 digits. I think the text gets a bit messy though if we start saying 'This is usually 9 characters but it could be less. It also could start with GB.... or maybe not.'

stephencdaly (approved):
LGTM!

PR #3160, opened 2022-01-06
PP-9080 add pages for 2022 contracts

nlsteers:
## WHAT
Added accessible pages for the 2022 contracts, these are PDF download only for now with a view adding HTML at a later date.

Paired with @stephencdaly


kbottla (approved):
LGTM!

PR #3170, opened 2022-01-11
PP-8501 Update webhook detail page to match backend

sfount:
Lines webhook detail/ message list up with https://github.com/alphagov/pay-webhooks/pull/43

Match backend design changes in the webhook service. Match fixtures and
process to the new schema specified in
https://github.com/alphagov/pay-webhooks/pull/43.

Delivery attempts and message details to follow in separate change.


gidsg (approved):
LGTM ✌️ 👍

PR #3171, opened 2022-01-12
PP-8501 Webhook event detail page

sfount:
Depends on https://github.com/alphagov/pay-selfservice/pull/3170

An individual message, showing the history of delivery attempts

Screenshot 2022-01-12 at 11 15 49

All messages for a given webhook

Screenshot 2022-01-12 at 11 15 26

### DIFF
--- a/app/controllers/webhooks/webhooks.controller.js
+++ b/app/controllers/webhooks/webhooks.controller.js
@@ -15,7 +15,7 @@ async function webhookDetailPage (req, res, next) {
 
   try {
     const webhook = await webhooksService.getWebhook(req.params.webhookId, req.service.externalId)
-    const messages = await webhooksService.getWebhookMessages(req.params.webhookId, { page, ...status && status })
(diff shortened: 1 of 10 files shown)


app/controllers/webhooks/webhooks.controller.js
```diff
@@ -107,6 +118,16 @@ async function toggleActiveWebhook(req, res, next) {
   }
 }
 
+async function resendWebhookMessage(req, res, next) {
```
gidsg, line 121: we don't actually have the backend for this yet, but good that we're thinking about the UI for it!

sfount: yeah we'll need to capture getting this done on the backend! I've removed the button from actually showing up until that's in as for now it would 404 trying to talk to the webhooks service

gidsg (dismissed):
✌️ 👍 LGTM

gidsg (approved):

PR #3185, opened 2022-01-20
PP-9085: Reduce docker image size and increase security with multi-stage build

jfharden:
## WHAT
For now this is just a demo and part of thinking about future ci, upgrading alpine and improving image builds. I also included a Dockerfile which changes the build to a debian buster-slim (appmetrics wont build on bullseye) and bumps the patch node version to 12.22.9 for security updates.

Change the build to be a muilti-stage buiild. This allows appmetrics to be built and not to keep the installed tool-chain or any intermediate files in the build (object files etc). Copy the final installed node_modules into the production stage of the build. For comparison the current latest-master is 497MB, and this build is 223MB. The Debian build is 290MB

```
$ docker image ls | grep selfservice
govukpay/selfservice              multi-stage               6f0bc14f9932   14 seconds ago      223MB
govukpay/selfservice              latest-master             0127d28947be   2 days ago          497MB
govukpay/selfservice              debian-slim               796dc2052499   7 minutes ago       290MB
```

~This works for me in a pay local but that of course doe not guarantee it works on live.~

To check metrics are still sent locally I made a simple UDP socket listener in ruby
```
#!/usr/bin/env ruby

require "socket"

PORT = 8125

sock = UDPSocket.new
sock.bind("", PORT)
puts "Bound  socket on port #{PORT}"

loop do
  data, _ = sock.recvfrom(1024)
  puts "[#{data}]"
end
```

and ran this.

I modified the pay-local docker-compose template to set the env var `METRICS_HOST=host.docker.internal` and to use my custom build of selfservice instead of latest-master.

Then I ran self-service and I can see the stats getting sent as expected:

```
[selfservice.memory.process.private:302825472|g]
[selfservice.memory.process.physical:109064192|g]
[selfservice.memory.process.virtual:334819328|g]
[selfservice.memory.system.used:7730221056|g]
[selfservice.memory.system.total:8346603520|g]
[selfservice.cpu.process:0.000555362|g]
[selfservice.cpu.system:0.0211038|g]
[selfservice.eventloop.latency.max:0.3438|g]
[selfservice.eventloop.latency.min:0.1182|g]
[selfservice.eventloop.latency.avg:0.19147|g]
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12.22.7-alpine3.12@sha256:99eaf1312b1926bc6db27d7230c8b3118d4ad2db64cc6a8a8304adeb8bad283b AS base
+FROM node:12.22.7-alpine3.12 as base
 
 RUN apk --no-cache upgrade && apk add --no-cache tini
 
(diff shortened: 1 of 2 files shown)


jfharden:
This is no longer required since we removed appmetrics

PR #3198, opened 2022-01-24
PP-9128 KYC banner update

nlsteers:
## WHAT
Amend the KYC banner to reflect the passed deadline. Additionally, show this banner to non-admin users of the service without the update CTA


app/views/dashboard/_additional_kyc_info_banner.njk
```diff
@@ -1,29 +1,48 @@
...
-          You have until {{ stripeKycAdditionalDataDueDate }} to add these details. If you do not add them, your service will stop taking payments and Stripe will stop paying out to your bank account.
-      </p>
-      <p class="govuk-body">
-        Continue to
-        <a id="add-additional-kyc-details" class="govuk-notification-banner__link"
-           href="{{ formatAccountPathsFor(routes.account.yourPsp.index, currentGatewayAccount.external_id, activeCredential.external_id) }}">
-          Add new KYC Information</a>.
-      </p>
+        
+        <h2 class="govuk-notification-banner__heading">
```
kbottla, line 9: (minor) Any reason this is a h2 ? It is making both notification title and this text h2

image

nlsteers: My reasoning was that the first `p` was classed `banner__heading` and should probably be a heading element (the design system uses a mix of heading or paragraph tags for this class).

I can revert back / use a lower order heading if preferable?

kbottla: component has two elements (header & heading). `banner__header` is h2 and `banner__heading` is usually a paragraph with styling to emphasize text.

kbottla: yeah can use paragraph so it will be inline with design system.

nlsteers: updated

kbottla (approved):

PR #3201, opened 2022-01-27
PP-9052 Fix name service during account creation bug

stephencdaly:
Ensure that the service we are setting the name for as the last step of a user signing up for a user account is the test service created for them during the sign-up process.

There was an issue where if the user started registering for an account and then accepted an invite to join a service before completing it, they could end up changing the name of the service they were invited to by mistake, as our code was only expecting them to have one service at this point.

To avoid this, check for a service with the default name, which is always "System Generated", and rename that service. If the user does not have a service with the default name, show an error page.

If the user tries to access the page for setting the service name when they do not have a service with the default name, redirect to the "My services" page.

Extract the default service name, which is "System Generated" into a constant referenced from all places we use it in application code.

## Review notes

Hiding whitespace changes will make the tests easier to review, as there are indentation changes
Screenshot 2022-01-27 at 15 06 44

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -405,7 +405,7 @@
         "filename": "test/unit/controller/register-service-controller/register-service.controller.test.js",
         "hashed_secret": "e6b6afbd6d76bb5d2041542d7d2e3fac5bb05593",
         "is_verified": false,
-        "line_number": 18,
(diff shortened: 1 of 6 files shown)


kbottla (approved):
LGTM!

PR #3211, opened 2022-02-01
PP-9161 Transaction Table - update styling in the mobile view

iqbalgds:
- Set the background colour of the table to white so that in smaller
  screens, you do not get the grey overlap.
- Tables with an overflow do not style well on small screens.
  So cannot add a right margin.  So have just left it to go
  to the edge of the screen.

### DIFF
--- a/app/assets/sass/components/transaction-list.scss
+++ b/app/assets/sass/components/transaction-list.scss
@@ -1,3 +1,7 @@
+.transactions-list-table{
+  background-color: govuk-colour("white");
+}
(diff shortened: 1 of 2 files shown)


stephencdaly (approved):
It's unclear from the ticket whether the intention was to make the whole of the grey background white or just the table. If just the table, looks good.

iqbalgds:
Thanks - I will update the ticket to make it clear that only the table background is being set to white.

PR #3227, opened 2022-02-11
PP-9187 re-adding glibc

SandorArpa:
## WHAT
- glibc needed to build Docker image


docker/build_and_test.Dockerfile
```diff
@@ -1,5 +1,11 @@
 FROM node:12.22.4-alpine3.12@sha256:a199f64c814f12fb00fa81a6e1c7f964e260a127f2b64ac526a358f3152c0f66
 
+### Needed to build docker image
```
stephencdaly, line 3: Probably the comment before was more helpful, as surely all commands in this file are needed to build the docker image?

SandorArpa: 👍

PR #3235, opened 2022-02-14
PP-9155 refactor get current credential

SandorArpa:
## WHAT
- refactor getCurrentCredential() to return any credentials when there is only one credential and it's ACTIVE. This is needed as Stripe account might not be ACTIVE yet.
- add new method getActiveCredential() and use it in switch-psp instead of getCurrentCredential()
- update tests


app/utils/credentials.test.js
```diff
@@ -24,11 +24,12 @@ describe('credentials utility', () => {
       expect(credential.state).is.equal('ACTIVE')
     })
 
-    it('validly returns nothing when no active credentials exist', () => {
+    it('validly returns the only credential when no active credentials exist', () => {
```
stephencdaly, line 27: Worth adding additional unit tests for the new getActiveCredential method?

stephencdaly (dismissed):

PR #3248, opened 2022-02-21
PP-9264 Update Footer links to new versions of contracts.

iqbalgds:
- Revert change with Footer links point to the new 2022 URL.
  - Make them link to the original URL.
- Update the orignal contract URLs to use the new V2 policy template files.
- policy.controller
  - Remove call to get HTML from S3.
  - We will add this back later once we have HTML versions of the contracts.

### DIFF
--- a/app/controllers/policy/supported-policy-documents.js
+++ b/app/controllers/policy/supported-policy-documents.js
@@ -6,17 +6,17 @@ const supportedPolicyDocuments = [
   {
     key: 'memorandum-of-understanding-for-crown-bodies',
     title: 'Pay memorandum of understanding',
-    template: 'policy/document/memorandum-of-understanding-for-crown-bodies'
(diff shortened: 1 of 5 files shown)


app/controllers/policy/supported-policy-documents.js
```diff
@@ -6,17 +6,17 @@ const supportedPolicyDocuments = [
   {
     key: 'memorandum-of-understanding-for-crown-bodies',
     title: 'Pay memorandum of understanding',
-    template: 'policy/document/memorandum-of-understanding-for-crown-bodies'
+    template: 'policy/document/v2/memorandum-of-understanding-for-crown-bodies'
```
kbottla, line 9: We also need to upload new files to S3. Download is still pointing to old versions

kbottla (approved):

PR #3256, opened 2022-03-01
BAU — Fix ‘Request a live account’ link in payment link success box

alexbishop1:
Fix the ‘Request a live account’ link that appears in the success message after creating a test payment.


SandorArpa (approved):
LGTM

PR #3265, opened 2022-03-10
PP-9377 Upgrade Node 12 to 16

sfount:
- update major Node version to 16 LTS
- move to multi stage docker builds for Node projects

Relates to https://github.com/alphagov/pay-frontend/pull/2692


.github/workflows/run-tests.yml
```diff
@@ -61,7 +64,7 @@ jobs:
       - name: Setup
         uses: actions/setup-node@9ced9a43a244f3ac94f13bfd896db8c8f30da67a
         with:
-          node-version: 12.22.10
+          node-version: "${{ steps.parse-node-version.outputs.NVMRC }}"
```
gidsg, line 67: Does this share between jobs?

sfount: 🤔 very good question, I'm guessing yes as this worked agains this PR but it would be helpful to prove it!

sfount: No I've verified that subsequent jobs are just defaulting to whatever behaviour the node runner defaults to. 

This should be updated along with https://github.com/alphagov/pay-products-ui

.github/workflows/run-tests.yml
```diff
@@ -10,17 +10,29 @@ permissions:
...
+    runs-on: ubuntu-18.04
+    name: Parse versions
+    outputs:
+      node-version: ${{ steps.parse-node-version.outputs.nvmrc }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846
+      - name: Parse Node version
+        id: parse-node-version
+				run: echo "::set-output name=nvmrc::$(cat .nvmrc)"
```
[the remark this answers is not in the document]
sfount: 🤦 thanks, sorted

gidsg (approved):
LGTM 🆙 👍

PR #3270, opened 2022-03-15
PP-9148 override minimist to 1.2.5

nlsteers:
## WHAT
Override `minimist` to `1.2.5` to prevent security warnings


SandorArpa (approved):
LOKTM

PR #3279, opened 2022-03-22
BAU: Remove unused PaaS files

gidsg:
## WHAT
This was added for PaaS in https://github.com/alphagov/pay-selfservice/pull/1815 so can be safely deleted

## HOW 
build should still pass


jfharden:
@gidsg You'll need to rebase this to get the cypress tests to pass

katstevens (approved):

PR #3301, opened 2022-04-08
PP-9460 Collect amount hint for payment links

stephencdaly:
When the option that the user can pick the amount is clicked, display
a text input that allows them to specify a hint to be shown on the
amount page.

Send this hint when creating the product.

Hide this behind a new feature flag, PAYMENT_LINKS_FUTURE_ENABLED so it
is not shown unless this feature flag is enabled. For editing a payment
link show the input either if the environment variable is enabled, or
the `new_payment_link_journey_enabled` flag is true for the product.

Screenshot 2022-04-08 at 12 39 50

Screenshot 2022-04-08 at 12 40 01


app/controllers/payment-links/get-edit-amount.controller.js
```diff
@@ -21,14 +21,17 @@ module.exports = function showEditAmountPage (req, res) {
...
 
   const pageData = {
     amountType,
     amountInPence,
+    amountHint,
     isWelsh,
     errors: recovered.errors,
-    isEditing: true
+    isEditing: true,
+    displayFuturePaymentLinksContent: process.env.PAYMENT_LINKS_FUTURE_ENABLED === 'true' || sessionData.newPaymentLinkJourneyEnabled
```
oswaldquek, line 34: where does this `newPaymentLinkJourneyEnabled` get set or come from?

app/views/payment-links/edit.njk
```diff
@@ -38,6 +38,11 @@
       {{ product.price | penceToPoundsWithCurrency }}
     {% else %}
       <span class="pay-text-grey">User can choose</span>
+      {% if product.amountHint %}
+        <span class="govuk-!-display-block pay-text-grey">
+          {{ product.amountHint | striptags(true) | escape | nl2br }}
```
oswaldquek, line 43: What does `escape | nl2br` do?

app/controllers/payment-links/get-edit-amount.controller.js
stephencdaly: Yeh this is a bit of a tricky one to follow through. When you're editing an existing payment link, we take the whole product and basically set `session.editPaymentLinkData` equal to the product object. So this comes from the Product we get from products

app/views/payment-links/edit.njk
stephencdaly: This is so we display back the hint correctly as the user entered it.
`escape` will escape special characters so they are displayed correctly when the HTML is rendered
`nl2br` converts new line characters to `<br>` tags so that line breaks put in by the user are displayed back correctly.

This is copied from above, where we display back the reference hint.

test/cypress/integration/payment-links/amount-hint.cy.test.js
```diff
@@ -0,0 +1,143 @@
...
+const gatewayAccountExternalId = 'a-valid-account-id'
+const serviceExternalId = 'a-valid-service-id'
+const serviceName = {
+  en: 'pay for something',
+  cy: 'talu am rywbeth'
+}
+const productName = 'Test'
+const amountHint = 'A hint'
+
+describe('A payment link with an amount hint set', () => {
```
oswaldquek, line 16: Do we need a test where the amount hint is not shown if the env var PAYMENT_LINKS_FUTURE_ENABLED is false / not set?

app/controllers/payment-links/get-edit-amount.controller.js
oswaldquek: Err I don't follow, especially the "So this comes from the Product we get from products" bits. Are there any links?

stephencdaly: It's done here.
https://github.com/alphagov/pay-selfservice/blob/master/app/controllers/payment-links/get-edit.controller.js#L50

It's a little bit confusing, it gets the product from products
```const product = await productsClient.product.getByProductExternalId(gatewayAccountId, productExternalId)```

Then merges it with some other data
```pageData.product = lodash.merge(product, editPaymentLinkData)```

Then sets it on the session
```lodash.set(req, 'session.editPaymentLinkData', pageData.product)```

So in the end, `session.editPaymentLinkData` contains all the fields on a GET product response. 

And the response from products is transformed using this `Product` class https://github.com/alphagov/pay-selfservice/blob/master/app/models/Product.class.js#L24

Although just spotted, we aren't loading in that `newPaymentLinkJourneyEnabled` for that class, so I'll make a small change to update that now.

stephencdaly: Updated

test/cypress/integration/payment-links/amount-hint.cy.test.js
stephencdaly: Ideally, yes - but I haven't found a way to change the value of environment variables between different cypress tests.

So this will be a little bit of a blind spot for testing. But as this is selfservice we can probably get away without everything being tested in a super strict way. They worst that could happen if this does show up is the user enters a hint that then isn't currently used by anything, but will be in the future.

app/controllers/payment-links/get-edit-amount.controller.js
oswaldquek: > Although just spotted, we aren't loading in that newPaymentLinkJourneyEnabled for that class, so I'll make a small change to update that now.

I think that's it 😄

oswaldquek (approved):

PR #3324, opened 2022-04-28
PP-9541 Agreements list page

sfount:
Agreement list page template with some simple macros that can be shared
between any pages showing agreements.

* agreement status tag - provide uniform human readable description and
  colour for a given agreement status
* payment instrument - display a card payment instrument with card brand
  logo, default to unknown logo if not registered

Agreement controller connected up with paths and routes, support basic:

* pagination
* filtering

Controller passes some basic filters to the ledger agreements utility
which are added to the request query string.

* status - a single string (multiple statuses not supported for now)
* reference - full or partial reference string
* page - number, defaults to 1

Functionality gated by account property `recurring_enabled`, this can be
used to enable the functionality for certain accounts as the feature is
rolled out.

Cypress coverage to assert the page loads and provides basic
interactions as expected.

Screenshot 2022-04-28 at 10 57 02


alexbishop1 (changes requested):
Looks good!

Except the card logos, which are old!

Mastercard hasn’t use that logo for years. You can get an updated logo from https://brand.mastercard.com/brandcenter/artwork.html.

I don’t think the Visa logo has yellow in the ‘V’ any more and they seem to have even abandoned the stripes (and lightened the blue) if https://merchantsignageeu.visa.com/product.asp?dptID=696 is anything to go by (I highly recommend reading the Visa Digital Brand Requirements, if only to discover that there is a Visa Haptic Vibration).

For now, we should probably reuse https://github.com/alphagov/pay-frontend/blob/master/app/assets/images/accepted-cards-sprite.png (even though it looks like the Visa logo might be outdated because they seem to enjoy changing it just to amuse themselves).

alexbishop1 (approved):
We can fix the images later.

alexbishop1 (approved):

PR #3333, opened 2022-05-12
PP-8608 remove deprecated stripe contract fee handling

SandorArpa:
## WHAT
- Stripe contract fee V2 is now applicable to all services which use Stripe as PSP, thus we no longer need `includeFeeBreakdownHeaders`. All Stripe csv download will contain the fee break down. Remove fee break down flag.


iqbalgds (approved):
LGTM

PR #3379, opened 2022-06-16
PP-9587: Display banner on dashboard for disabled accounts

stephencdaly:
If gateway account has the `disabled` flag set to true, display a banner on the dashboard that indicates the account is disabled. Do not display any of the other possible banners if this is displayed.

Add a Cypress test for this. The Pact with connector is updated to expect the `disabled` flag, as we've updated the fixture used to generate the Pact.


alan-gds (approved):
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay.

PR #3380, opened 2022-06-17
PP-9587 Show refund error for account disabled

stephencdaly:
If a refund fails because the account is disabled, show a specific error telling the user that the refund failed for this reason.


alan-gds (approved):
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay.

PR #3425, opened 2022-07-22
PP-9671 Display not refundable message when payment disputed

stephencdaly:
When a payment has "disputed"=true and the refund_summary status of
"unavailable", display a paragraph on the transaction view indicating
that it is not possible to refund the payment.

Add validation that the "disputed" field exists on transactions
retrieved from ledger to the Pact tests. This is done by updating the
fixture that is used to build the Pact.


SandorArpa (approved):
LGTM

PR #3467, opened 2022-08-15
PP-9016 Refine webhooks status tags

sfount:
Following the pattern set by agreements for recurring card payments.
Move webhook and webhook message status tags to macros that can be
shared between templates. Add entries for new statuses.

**Webhook status**

Screenshot 2022-08-15 at 16 09 41

**Message status**

Screenshot 2022-08-15 at 17 46 10

### DIFF
--- a/app/views/webhooks/_message_status.njk
+++ b/app/views/webhooks/_message_status.njk
@@ -1,0 +1,16 @@
+{% set statusStyleMap = {
+"PENDING": "govuk-tag--grey",
+"SUCCESSFUL": "govuk-tag--green",
+"FAILED": "govuk-tag--red",
+"WILL_NOT_SEND": "govuk-tag--yellow"
+} %}
+{% set statusTextMap = {
+"PENDING": "Pending",
+"SUCCESSFUL": "Successful",
+"FAILED": "Failed",
+"WILL_NOT_SEND": "Will not send"
+} %}
+
+{% macro messageStatusTag(status) %}
+<strong class="govuk-tag {{ statusStyleMap[status] }}">{{ statusTextMap[status] }}</strong>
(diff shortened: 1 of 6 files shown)


app/views/webhooks/_message_status.njk
```diff
@@ -0,0 +1,16 @@
+{% set statusStyleMap = {
+"PENDING": "govuk-tag--grey",
+"SUCCESSFUL": "govuk-tag--green",
+"FAILED": "govuk-tag--red",
+"WILL_NOT_SEND": "govuk-tag--yellow"
+} %}
+{% set statusTextMap = {
+"PENDING": "PENDING",
```
alexbishop1, line 8: Should the labels be upper-case? I would have thought we’d use title (or maybe title) case and use CSS to make it all caps.

app/views/webhooks/_message_status.njk
```diff
@@ -0,0 +1,16 @@
...
+"SUCCESSFUL": "SUCCESSFUL",
+"FAILED": "FAILED",
+"WILL_NOT_SEND": "WON’T SEND"
```
alexbishop1, line 11: May as well make the label WILL NOT SEND because [we shouldn’t use negative contractions](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#contractions).

alexbishop1 (changes requested):

app/views/webhooks/_message_status.njk
sfount: Yeah CSS already makes it all caps by virtue of it being a tag, the result will be the same, happy with either

alexbishop1 (approved):

PR #3484, opened 2022-08-24
PP-9784 Add cancelled filter value to agreement status filter

sfount:
Include the `CANCELLED` status for filter values on agreement state.

This is propagated to Ledger and matches https://github.com/alphagov/pay-java-commons/blob/master/model/src/main/java/uk/gov/service/payments/commons/model/agreement/AgreementStatus.java


sfount:
This has been superseded by other work on statuses

PR #3490, opened 2022-08-30
PP-9882 Create payment link page: warn if Worldpay MOTO service

alexbishop1:
Add a warning to the create payment link page imploring the user to not create a payment link if and only if the gateway account has current credentials for Worldpay using a merchant code that ends in ‘MOTO’ (which is how Worldpay identify MOTO merchant codes).

The existing warning on the create payment link page, which informs users about the existence of telephone payment links (agent-initiated MOTO payments) is changed to inset text so we do not end up with two warnings on single page.

image

### DIFF
--- a/app/controllers/payment-links/get-start.controller.js
+++ b/app/controllers/payment-links/get-start.controller.js
@@ -3,9 +3,18 @@
 const lodash = require('lodash')
 
 const { response } = require('../../utils/response.js')
+const { getCurrentCredential } = require('../../utils/credentials')
 
 module.exports = (req, res) => {
   lodash.set(req, 'session.pageData.createPaymentLink', {})
 
-  return response(req, res, 'payment-links/index', {})
+  const credential = getCurrentCredential(req.account)
+
+  const accountUsesWorldpayMotoMerchantCode = lodash.get(credential, 'payment_provider', '') === 'worldpay'
+      && lodash.get(credential, 'credentials.merchant_id', '').endsWith('MOTO')
+
+  return response(req, res, 'payment-links/index', {
+    accountUsesWorldpayMotoMerchantCode: accountUsesWorldpayMotoMerchantCode
+  })
+
 }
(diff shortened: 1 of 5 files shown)


app/controllers/payment-links/get-start.controller.js
```diff
@@ -3,9 +3,19 @@
...
+const { getCurrentCredential } = require('../../utils/credentials')
 
 module.exports = (req, res) => {
   lodash.set(req, 'session.pageData.createPaymentLink', {})
 
-  return response(req, res, 'payment-links/index', {})
+  const credential = getCurrentCredential(req.account)
+
+  const accountUsesWorldpayMotoMerchantCode = (credential && credential.payment_provider === 'worldpay'
+      && credential.credentials && credential.credentials.merchant_id
```
kbottla, line 14: (minor) Could use `lodash.get(credential, 'credentials.merchant_id', '')` to reduce few conditions but fine as it is.

test/unit/controller/payment-links/get-start.controller.test.js
```diff
@@ -0,0 +1,163 @@
...
+    '../../utils/response': mockResponses
+  })
+}
+
+describe('The Worldpay MOTO account warning', () => {
+  let req
+  let res
+
+  describe('when the gateway account has a Worldpay credentials object with a merchant code ending with ‘MOTO’', () => {
+    before(() => {
```
kbottla, line 19: (not blocking) Any reason each test has its own setup (`before()`) method instead of a global one? 

We could have a common setup and within tests, `req` object values can be set as per the test.

kbottla (approved):
Looks good. Added a few minor comments but not blocking.

app/controllers/payment-links/get-start.controller.js
alexbishop1: Good idea, will do.

test/unit/controller/payment-links/get-start.controller.test.js
alexbishop1: Will move some stuff into a global `beforeEach(…)` function.

kbottla (approved):

PR #3506, opened 2022-09-12
PP-10041: Send OtpCode to 'complete' endpoint

alan-gds:
* Include an otp code in the request body when calling `${inviteResource}/${inviteCode}/complete` on adminusers.
  * Except for when the invite is for an existing user - the otp code is not required in that case.
* adminusers will not yet actually use this code, but we are soon going to make changes that mean it will.
* This PR also includes some cleanup relating to https://github.com/alphagov/pay-selfservice/pull/2890 - the pact tests were still passing gatewayAccountIds in the request body, but that has not been necessary or useful for quite a while now.

### DIFF
--- a/app/controllers/register-service.controller.js
+++ b/app/controllers/register-service.controller.js
@@ -190,7 +190,7 @@ async function createPopulatedService (req, res, next) {
   }
 
   try {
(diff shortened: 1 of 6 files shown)


sfount (approved):
Change looks good, we've discussed adding pacts for the new properties in a future PR.

Great cleanup from previous change 👍

PR #3513, opened 2022-09-20
Revert "PP-10041: Send OtpCode to 'complete' endpoint"

alan-gds:
Reverts alphagov/pay-selfservice#3506

We realised that unifying OTP validation and invite completion isn't a good idea (it makes it too difficult to define suitable test coverage).  So, #3506 is no longer helpful and should be reverted.  See https://payments-platform.atlassian.net/browse/PP-10041?focusedCommentId=48911

### DIFF
--- a/app/controllers/register-service.controller.js
+++ b/app/controllers/register-service.controller.js
@@ -190,7 +190,7 @@ async function createPopulatedService (req, res, next) {
   }
 
   try {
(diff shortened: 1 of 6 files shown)


sfount (approved):
👍 Looks like it would still be good to remove some of the no-longer-used-but-not-removed properties on method descriptions/ stub builders etc. that have been cleaned up in this PR. Maybe that can attached to the ticket as low priority follow up

PR #3542, opened 2022-10-10
WIP Commit

iqbalgds:
- Update the `kyc-tasks` service
  - Add new task for `confirm org details`
- Checking for the business name and business address line 1 to check if the task is complete.


kbottla:
Don't need to update KYC tasks for organisation details. We should be updating the switch PSP tasks https://github.com/alphagov/pay-selfservice/blob/master/app/controllers/switch-psp/switch-tasks.service.js and relevant UI

PR #3546, opened 2022-10-12
PP-10041: Prepare for switching to new InviteType strings

alan-gds:
## WHAT
We have added several new InviteType enum values in pay-adminusers.  Like the old ones (which are currently still in use), these get serialised into self-service requests as lower-cased strings.  Self-service needs to be ready to handle these strings as potential values for the `type` field of invites.

## HOW 
Please check the test coverage and let me know what you think.  I've made duplicates of the tests in `invite-validation.controller.ft.test.js`, but there are a few other test classes (`service-registration.service.it.test.js`, `self-reigstration-otp-verify.ft.test.js`, `complete-service-invite.pact.test.js` and `complete-user-invite.pact.test.js`) which reference the old enum values in a more incidental way (I think?), and it's not clear to me whether it's worth making duplicates of these at this stage (especially for cases such as `complete-user-invite.pact.test.js`, where 'user' is only referenced in the setup of a `describe` block).

### DIFF
--- a/app/controllers/invite-validation.controller.js
+++ b/app/controllers/invite-validation.controller.js
@@ -38,7 +38,13 @@ async function validateInvite (req, res, next) {
       req.register_invite.email = invite.email
       const redirectTarget = invite.user_exist ? paths.registerUser.subscribeService : paths.registerUser.registration
       res.redirect(302, redirectTarget)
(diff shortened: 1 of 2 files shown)


stephencdaly (approved):
LGTM!

PR #3547, opened 2022-10-13
PP-10048 Don't display phone number when not present

stephencdaly:
It will be possible for a User to not have a telephone number set if they have registered using an Authenticator App as their second factor method.

If a phone number isn't set, don't display the phone number row on the "My profile" screen. In future, we will amend the user journey to change the second factor method SMS to ask for a phone number if the user does not have one set.


alan-gds (approved):
LGTM

PR #3549, opened 2022-10-13
BAU: Stripe: Mark responsible person as provided

kbottla:
## WHAT
- For new accounts with Stripe we are confirming manually on the Stripe dashboard that all executives are provided. Do this as soon as services enter the responsible person's information.


SandorArpa (approved):
LGTM

PR #3561, opened 2022-10-18
PP-10047 Restructure controllers

stephencdaly:
To make controller directory structure make more sense, move `two-factor-auth-controller` directory to be under the `user` directory. Also remove `-controller` from the directory name as this is not a pattern we use elsewhere.


alan-gds (approved):
👍

PR #3562, opened 2022-10-18
PP-10041: Switch to using new /otp/validate endpoint

alan-gds:
In `adminusers.client`, stop calling `/v1/api/invites/otp/validate/service` and instead call `/v2/api/invites/otp/validate`.


stephencdaly (approved):
LGTM! I'd probably recommending pinning the version on deploy-to-staging in concourse before merging this so we can manually check on the test environment haven't accidentally broken anything as the contract testing is currently not there.

PR #3563, opened 2022-10-19
PP-10047 Use constants for second factor method

stephencdaly:
Rather than hardcoding the strings "SMS" or "APP" for the second factor method in many places in the code, define constants that are used in all places.

### DIFF
--- a/app/controllers/user/two-factor-auth/post-index.controller.js
+++ b/app/controllers/user/two-factor-auth/post-index.controller.js
@@ -5,20 +5,21 @@ const lodash = require('lodash')
 const logger = require('../../../utils/logger')(__filename)
 const userService = require('../../../services/user.service.js')
 const paths = require('../../../paths')
+const secondFactorMethod = require('../../../models/second-factor-method')
 
 module.exports = async (req, res) => {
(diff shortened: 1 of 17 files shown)


app/controllers/user/two-factor-auth/post-index.controller.js
```diff
@@ -5,20 +5,21 @@ const lodash = require('lodash')
...
+    try {
+      const user = await userService.provisionNewOtpKey(req.user.externalId, req.correlationId)
+      if (method === secondFactorMethod.SMS) {
+        await userService.sendProvisionalOTP(user, req.correlationId)
+      }
+      return res.redirect(paths.user.profile.twoFactorAuth.configure)
+    } catch (err) {
+      logger.error(`Provisioning new OTP key failed - ${err.message}`)
+      req.flash('genericError', 'Something went wrong. Please try again or contact support.')
+      return res.redirect(paths.user.profile.twoFactorAuth.index)
```
alan-gds, line 23: The indentation looks incorrect here - compare with line 12.

alan-gds (approved):
Looks fine, apart from a possible indentation issue.

PR #3564, opened 2022-10-19
PP-10041: Follow new flow for new user/existing service invite journey

alan-gds:
* Instead of calling /v1/invites/otp/validate, make a call to /v2/invites/otp/validate and then /v1/invites/${inviteCode}/complete.
* Pact tests for the old flow have been deleted without replacement; the new flow would be tricky to pact test.


app/controllers/register-user.controller.js
```diff
@@ -158,18 +158,28 @@ const submitOtpVerify = async function submitOtpVerify (req, res, next) {
...
+      return next(err)
+    }
+  }
+
+  try {
+    const completeResponse = await registrationService.completeInvite(sessionData.code, correlationId)
+    loginController.setupDirectLoginAfterRegister(req, res, completeResponse.user_external_id)
+    return res.redirect(303, paths.registerUser.logUserIn)
+  } catch (err) {
+    if (err.errorCode === 410) {
```
stephencdaly, line 182: Looks like we don't have tests that cover the various error responses from adminusers which we handle differently. I'd be wary to add new functional tests for this. Maybe this is a case where we could start writing some unit tests for this class to add this particular coverage. Maybe something to pair on?

stephencdaly (commented):
Changes look correct, but think we could do with adding some test coverage that was already missing for this controller.

test/unit/controller/register-user.controller.test.js
```diff
@@ -103,4 +109,94 @@ describe('Register user controller', () => {
...
+      return proxyquire('../../../app/controllers/register-user.controller', {
+        '../services/user-registration.service.js': {
+          verifyOtp: verifyOtpSpy,
+          completeInvite: completeInviteSpy
+        }
+      })
+    }
+
+    describe('Submitted OTP code is correct', () => {
+      it('', async () => {
```
stephencdaly, line 124: This test needs a name

test/unit/controller/register-user.controller.test.js
```diff
@@ -7,55 +7,61 @@ const userFixtures = require('../../fixtures/user.fixtures')
...
 
 describe('Register user controller', () => {
   const email = 'invited-user@example.com'
-  const code = 'a-code'
+  const inviteCode = 'a-code'
   const correlationId = 'a-request-id'
   const serviceExternalId = 'a-service-id'
 
+  const verifyOtpSuccessStub = sinon.spy(() => Promise.resolve())
   const completeInviteSuccessStub = sinon.stub().returns(inviteFixtures.validInviteCompleteResponse({
```
stephencdaly, line 20: This already existed, but I think it should return `Promise.resolve(InviteFixtures.validInviteCompleteResponse(...))` as the complete method returns a Promise. This will just ensure that we're awaiting it correctly in the tests - otherwise they'd still pass even if we didn't `await` in the controller.

Also looks like stubs and spies can be used interchangeably here. Would be good to use the same thing for both these spy/stubs just to avoid confusion. I guess using a stub is maybe technically more correct but we use spy in most places.

stephencdaly (commented):
Looks good, a couple of minor things

test/unit/controller/register-user.controller.test.js
alan-gds: Ah, I did want to do it the way you're suggesting and I went and looked for something like `InviteFixtures.validInviteCompleteResponse`, but originally failed to spot it.  I agree that it would be best to be consistent here, and now that I know what fixture I can return, I will change the verifyOtp one.

test/unit/controller/register-user.controller.test.js
alan-gds: 🤦

test/unit/controller/register-user.controller.test.js
alan-gds: Wait, hold on a sec, isn't `InviteFixtures.validInviteCompleteResponse` only appropriate for invite completion?  Surely what we need is something along the line of `validOtpVerifyResponse`?  Sorry, I misread your comment.

alan-gds: I did indeed misread the comment, but not in the way I thought; Stephen has explained it to me on a call.

stephencdaly (approved):
LGTM!

PR #3566, opened 2022-10-20
PP-10047 Refactor user service

stephencdaly:
Refactor methods in user.service to accept a user external ID rather than an entire user object. This will make testing controllers that call this service easier.

There are 2 commits in this PR, the second is to run the linter on the whole of selfservice.


sfount (approved):
Change looks great, much easier to stub and avoids the methods using data they shouldn't have.

I'm guessing there are no unit tests for these as I don't see any updated but confident they're covered by higher level tests.

PR #3568, opened 2022-10-21
PP-10240 Remove the "old" client error handler

stephencdaly:
There is currently the concept of an "old" and a "new" error handler.
The "old" error handler puts the entire response body in the message
field of the "error" object returned by the client.

There was only one place in the code where fields in the response body
was accessed - when attempting to complete a registration. Here we were
using the backend error message to display on the error page. We
shouldn't do this, so instead just display a generic error page.

Remove the "old" error handler so that the clients return an error
with a uniform signature when a request returns a non-success response.
This error contains the error message string extracted from the HTTP
response body.

### DIFF
--- a/app/services/clients/base-client/wrapper.js
+++ b/app/services/clients/base-client/wrapper.js
@@ -34,9 +34,7 @@ module.exports = function (method, verb) {
 
     // Set up post response and error handling method
     const transform = opts.transform || false
-    const handleError = opts.baseClientErrorHandler || 'new'
     opts.transform = undefined
-    opts.baseClientErrorHandler = undefined
 
     // start request
     requestLogger.logRequestStart(context)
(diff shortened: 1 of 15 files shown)


sfount (approved):

app/services/clients/base-client/wrapper.js
```diff
@@ -49,21 +47,13 @@ module.exports = function (method, verb) {
...
-          reject(err)
-        } else {
-          // eslint-disable-next-line
-          reject({
-            errorCode: response.statusCode,
-            message: response.body
-          })
-        }
+        let errors = lodash.get(body, 'message') || lodash.get(body, 'errors')
+        if (errors && errors.constructor.name === 'Array') errors = errors.join(', ')
```
sfount, line 51: This is old code that sat in a condition you've removed so entirely optional but this feels a bit janky
```suggestion
        if (errors && Array.isArray(errors)) errors = errors.join(', ')
```

stephencdaly: Thanks, will use your suggestion in an upcoming PR

PR #3569, opened 2022-10-21
PP-10240 Make Base client API uniform

stephencdaly:
Previously, the base client "wrapper" function took 3 arguments and then determined which was which be looking at the type. These were:
- `url`, which was optionally present. The URL was otherwise provided in the `opts` object
- `opts`, which was always present and contains the request options
- `cb`, a callback which was optionally present. This was only used by tests.

Make it so that it now only takes one argument, the request options object. To allow for this, made changes to `connector.client` to always pass in the URL in the request options object.

Removed `wrapper-arguments.test.js` as it was testing the argument parsing logic which is no longer required.

Renamed `wrapper-request-callback.test.js`, and altered tests to work without needing to pass in the callback.

Deleted `base.client.test.js`. This contained 2 tests:
- The first was to check the headers were set correctly. Testing this has been moved to `wrapper.test.js`.
- The second was to check that the same client was used to make multiple requests. This wasn't testing anything useful so just removed it.


sfount (approved):
Much more consistent and easy to understand from a distance

PR #3574, opened 2022-10-24
PP-10240 Introduce RESTClientError type

stephencdaly:
Introduce a new type which extends Error for representing errors thrown by the REST client. This makes the API of these errors clearer and makes it possible to handle these in a specific way.

Handle uncaught RESTClientErrors in the error handler middleware in order to log more detail about the error.

Rewrite `wrapper.test.js` to be easier to understand.

### DIFF
--- a/app/errors.js
+++ b/app/errors.js
@@ -1,5 +1,16 @@
 'use strict'
 
+class RESTClientError extends Error {
+  constructor (message, service, errorCode, errorIdentifier, reason) {
+    super(message)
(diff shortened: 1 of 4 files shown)


sfount (approved):
I've not looked at everything that consumes these clients to be 100% confident that the new error shape lines up with the error handling but I suspect that almost all errors aren't recoverable and are just passed to the default error handler. Tests also pass.

This should make handling these upstream a lot clearer

PR #3580, opened 2022-10-25
PP-10240 Merge wrapper.js into base.client.js

stephencdaly:
Moving the function in wrapper.js into base.client.js should make it easier to follow how the base client works.


app/services/clients/base-client/base.client.js
```diff
@@ -36,3 +35,68 @@ module.exports = {
 function retryStrategy (err, response, body) {
   return process.env.NODE_ENV === 'production' && (err && ['ECONNRESET'].includes(err.code))
 }
+
+function makeRequest (method, opts) {
+  return new Promise((resolve, reject) => {
+    opts.method = method.toUpperCase()
+
+    const loggingContext = {
+      correlationId: correlator.getId(),
```
sfount, line 44: 🤔 if we do this on a base client and have [middleware to set this up on every request](https://github.com/alphagov/pay-selfservice/blob/master/app/routes.js#L152) why do we have any correlation IDs being passed around through code? 

I don't think these apps do anything outside of the context of responding to HTTP requests.

sfount (approved):
Refactor looks sensible to me, this being across two files made the context harder to understand and it doesn't feel too big in its current form. 

One quick question on understanding the methodology for passing on correlation IDs (although I know this hasn't changed in this PR), I seem to recall a lot of explicit passing around of correlation ID through client methods.

app/services/clients/base-client/base.client.js
stephencdaly: Oh yeh good point, I'll look a bit more into what's going on here. I'll merge this one as this functionality hasn't changed, and raise a new PR if appropriate.

PR #3581, opened 2022-10-26
PP-9648 Remove link to 3d-secure-2 page

kbottla:
## WHAT
- Removed link to 3ds-secure-2 product which is out of date and about to be removed


alexbishop1 (approved):
🚮

PR #3589, opened 2022-11-03
PP-10198 Validate merchant account code for MOTO and non-MOTO accounts

kbottla:
## WHAT
- Added validation to allow merchant IDs ending with MOTO only for moto enabled accounts, and allow non-MOTO IDs for non-moto accounts

### DIFF
--- a/app/controllers/credentials/worldpay.controller.js
+++ b/app/controllers/credentials/worldpay.controller.js
@@ -34,6 +34,16 @@ async function updateWorldpayCredentials (req, res, next) {
     const credential = getCredentialByExternalId(req.account, req.params.credentialId)
     const results = credentialsForm.validate(req.body)
 
+    if (!results.errorSummaryList.length) {
+      const merchantId = req.body.merchantId
+      if (req.account.allow_moto && !merchantId.endsWith('MOTO')) {
+        results.errors.merchantId = 'Enter a MOTO merchant code. MOTO payments are enabled for the account'
+      } else if (!req.account.allow_moto && merchantId.endsWith('MOTO')) {
+        results.errors.merchantId = 'MOTO merchant code not allowed. Please contact support if you would like MOTO payments enabled'
+      }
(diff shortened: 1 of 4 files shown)


SandorArpa (dismissed):
LGTM

app/controllers/credentials/worldpay.controller.js
```diff
@@ -35,12 +35,26 @@ async function updateWorldpayCredentials (req, res, next) {
     const credential = getCredentialByExternalId(req.account, req.params.credentialId)
     const results = credentialsForm.validate(req.body)
 
+    if (!results.errorSummaryList.length) {
+      const merchantId = req.body.merchantId
+      if (req.account.allow_moto && !merchantId.endsWith('MOTO')) {
+        results.errors.merchantId = 'Enter a valid MOTO merchant account code. MOTO payments are enabled for the account'
```
alexbishop1, line 41: I think just ‘merchant code’ would be better because it matches the terminology we use on the page (and what Worldpay use). Not sure the word ‘valid’ is really adding much.

app/controllers/credentials/worldpay.controller.js
```diff
@@ -35,12 +35,26 @@ async function updateWorldpayCredentials (req, res, next) {
...
+      } else if (!req.account.allow_moto && merchantId.endsWith('MOTO')) {
+        results.errors.merchantId = 'Enter a valid merchant account code. Please contact support if you would like moto payments enabled'
```
alexbishop1, line 43: The acronym ‘moto’ should definitely be capitalised. But maybe this message could more direct? Something like “MOTO merchant code not allowed. Please contact support if you would like MOTO payments enabled”.

alexbishop1 (changes requested):

alexbishop1 (approved):

alexbishop1 (approved):

PR #3590, opened 2022-11-03
PP-10276 Remove unused correlation ids with base client

sfount:
The base client currently uses the `correlation-id` library to pull the ID from an async local thread storage, it never uses the `correlationId` param that is passed in by controllers/ services.

Remove all references and instances of this being manually fetched and passed in.


stephencdaly (approved):
🔥

PR #3591, opened 2022-11-03
PP-10276 Replace correlation id with async local storage

sfount:
Depends on https://github.com/alphagov/pay-selfservice/pull/3590

Our logging currently uses the Node core asyc hooks to maintain a number
of metadata properties about each request to be logged with each
message.

The base client then separately uses a library `correlation-id` which
uses similar technology underneath.

Have the base client use the Node core library and remove this dep.


stephencdaly (approved):
LGTM!

PR #3596, opened 2022-11-07
PP-10037 Run all analytics ids through PII method

sfount:
Set the redacted 'location' property for all analytics ids.

Fix setting the `anonymizeIp` property on the linker plugin rather than than the tracking profile.


stephencdaly (approved):
🤞

PR #3599, opened 2022-11-08
PP-10244 Render "My services" page for no services

stephencdaly:
A user can be a member of no services if they have been removed from all services they were a member of. Currently the controller for showing the "My services" page throws an error if the user is a member of no services and an error page is rendered.

Fix this so that the controller does not throw an error, and the page is rendered. The rendered page does not include the "Reports" section and has an additional paragraph instructing the user on what to do next.

Add some additional Cypress tests to check we're displaying the "My services" page correctly in different scenarios.

## Review note

Hiding whitespace changes in the diff should make it easier to see what's changed in the template.

## Page

Screenshot 2022-11-08 at 16 13 31

### DIFF
--- a/app/controllers/my-services/get-index.controller.js
+++ b/app/controllers/my-services/get-index.controller.js
@@ -56,7 +56,10 @@ module.exports = async function getServiceList (req, res) {
   const aggregatedGatewayAccountIds = servicesRoles
     .flatMap(servicesRole => servicesRole.service.gatewayAccountIds)
 
-  const aggregatedGatewayAccounts = await serviceService.getGatewayAccounts(aggregatedGatewayAccountIds)
+  let aggregatedGatewayAccounts = []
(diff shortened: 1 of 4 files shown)


alexbishop1:
We say to contact “manager on an existing service” rather than an “administrator of an existing service”?

stephencdaly:
> We say to contact “manager on an existing service” rather than an “administrator of an existing service”?

Hmm we were deciding between the two, and looks like we might have decided on administrator but the mockup on the story says manager so I'll double check. I think we thought "manager" might be good if it's a new user who doesn't really know who to contact, but "administrator" is more consistent.

stephencdaly:
> We say to contact “manager on an existing service” rather than an “administrator of an existing service”?

Turns out we did want administrator, thanks

alan-gds (approved):
Looks good. 👍

PR #3601, opened 2022-11-08
PP-10244 Handle service not created during registration

stephencdaly:
We are going to remove creating a service as part of a self-signup registration.

The service is currently created by adminusers when a request is made to the `/v1/api/invites/{code}/complete` endpoint, which also creates the user account. Selfservice then redirects the user to a page to name the created service.

We are going to stop creating a test service in adminusers when `/v1/api/invites/{code}/complete` is called. When we are no longer creating a service, the response from this endpoint will no longer include a `service_external_id` field.

Prepare for a service no longer being created by handling the case where `/v1/api/invites/{code}/complete` does not include a `service_external_id`. In this case, do not attempt to create a gateway account in connector. And instead of redirecting the user to the page to set the service name after logging them in, redirect the user to the "My services" page.

## Review note

Hiding whitespace changes in the diff will make this PR a lot easier to review.

### DIFF
--- a/app/controllers/register-service.controller.js
+++ b/app/controllers/register-service.controller.js
@@ -150,7 +150,7 @@ async function showOtpVerify (req, res, next) {
   }
 }
 
-async function createPopulatedService (req, res, next) {
+async function submitOtpCode (req, res, next) {
   const sessionData = req.register_invite
   if (!registrationSessionPresent(sessionData)) {
     return next(new RegistrationSessionMissingError())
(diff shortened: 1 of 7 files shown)


app/controllers/register-service.controller.js
```diff
@@ -150,7 +150,7 @@ async function showOtpVerify (req, res, next) {
   }
 }
 
-async function createPopulatedService (req, res, next) {
+async function submitOtpCode (req, res, next) {
```
alan-gds, line 153: This is `submitOtpVerify` in `register-user.controller` - maybe it would be worth harmonising the names?

alan-gds (approved):
Looks good to me.  One suggestion for renaming (see comment), but that's optional.

app/controllers/register-service.controller.js
stephencdaly: Good point. I would agree if these controllers were sticking around, but our aim is to kill them off in the near future in favour of our new controllers for the combined journey so this inconsistency won't affect people for very long. So I won't do it this time if that's ok, just so I don't have to rebuild the PR.

alan-gds: 👍

PR #3604, opened 2022-11-11
PP-10244 Remove code for naming service during registration

stephencdaly:
Adminusers no longer creates a test service when `/v1/api/invites/{code}/complete` is called for a self-registration invite. Selfservice already handles this by not displaying the page to name a service when adminusers doesn't include a `service_external_id` in the response which will now always be the case.

Remove the routes and controllers for naming a service that are no longer user.

Delete `service-registration.service.it.test.js` as `service-registration.service.js` is now just a thin wrapper around the adminusers client and contains no logic.


alan-gds (approved):
LGTM

PR #3612, opened 2022-11-14
PP-10100 Limit organisation name to 100 characters

kbottla:
## WHAT
- Limits organisation name to 100 characters as PSP (stripe) allows only up to 100 characters.


stephencdaly (approved):
LGTM!

PR #3616, opened 2022-11-15
PP-10055 Add routes and templates for new registration journey

stephencdaly:
Add the GET routes and the basic templates for the new registration journey that includes the ability to select an authenticator app to get 2FA codes.

Currently when the routes are visited, the templates are just rendered with dummy data and no other logic is performed.


alan-gds (approved):
LGTM

PR #3620, opened 2022-11-16
PP-10055 Add client method to update invite password

stephencdaly:
Add a method to the adminusers client to update the password for an invite. Add a corresponding pact test.


SandorArpa (approved):
LGTM

PR #3622, opened 2022-11-17
PP-10055 Add route to submit password page

stephencdaly:
Add a POST `/register/password` route that is POSTed to when the password page is submitted. This does validation of the "Enter a password" and "Re-type your password" fields.

If the password is valid, a PATCH request is made to adminusers to update the password and we redirect to the next page in the journey.

### DIFF
--- a/app/controllers/registration/registration.controller.js
+++ b/app/controllers/registration/registration.controller.js
@@ -1,10 +1,16 @@
 'use strict'
 
 const qrcode = require('qrcode')
+const lodash = require('lodash')
 
 const { RegistrationSessionMissingError } = require('../../errors')
 const adminusersClient = require('../../services/clients/adminusers.client')()
 const paths = require('../../paths')
+const { validatePassword } = require('../../utils/validation/server-side-form-validations')
+const { isEmpty } = require('../../utils/validation/field-validation-checks')
+
+const PASSWORD_INPUT_FIELD_NAME = 'password'
+const REPEAT_PASSWORD_INPUT_FIELD_NAME = 'repeat-password'
 
 const registrationSessionPresent = function registrationSessionPresent (sessionData) {
   return sessionData && sessionData.email && sessionData.code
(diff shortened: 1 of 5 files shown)


app/controllers/registration/registration.controller.js
```diff
@@ -28,6 +31,39 @@ async function showPasswordPage (req, res, next) {
...
+
+  const passwordValidationResult = validatePassword(password)
+  if (!passwordValidationResult.valid) {
+    errors.password = passwordValidationResult.message
+  }
+  if (isEmpty(repeatPassword)) {
+    errors['repeat-password'] = 'Re-type your password'
+  } else if (!errors.password && password !== repeatPassword) {
+    errors['password'] = errors['repeat-password'] = 'Enter same password in both fields'
+  }
```
sfount, line 53: All of this feels like it could be moved out to a separate function (at the least) or a service to help code readability at the controller level and also a little bit easier to test

I don't think this is a blocker.

Would be really good to get around to doing form validation in a uniform way which I know we've talked about a few times internally.

kbottla (approved):

app/controllers/registration/registration.controller.js
```diff
@@ -28,6 +31,39 @@ async function showPasswordPage (req, res, next) {
   }
 }
 
+async function submitPasswordPage (req, res, next) {
+  const sessionData = req.register_invite
+  if (!registrationSessionPresent(sessionData)) {
+    return next(new RegistrationSessionMissingError())
+  }
+
+  const password = req.body['password']
```
sfount, line 40: 'password' and 'repeat-password' feel like they could be constants to 
* describe what they actually are
* make sure they're the same everywhere they're used

(these could optionally be passed in and used in the template but that might be overkill)

Something like?
```js 
const PASSWORD_INPUT_FIELD_NAME = 'password'
const REPEAT_PASSWORD_INPUT_FIELD_NAME = 'repeat-password'
```

app/controllers/registration/registration.controller.test.js
```diff
@@ -69,6 +70,114 @@ describe('Registration', () => {
...
+    })
+
+    it('should render the password page with an error when both password fields are empty', async () => {
+      req.body = {
+        password: '',
+        'repeat-password': ''
+      }
+      await registrationController.submitPasswordPage(req, res, next)
+
+      sinon.assert.calledWith(res.render, 'registration/password', {
```
sfount, line 91: Test coverage looks great but as the comment above I wonder if these should be covering a simple service and then the controller just checks once that the appropriate error object is passed through

sfount (dismissed):

app/controllers/registration/registration.controller.js
stephencdaly: good point, have updated to use constants

sfount (approved):

app/controllers/registration/registration.controller.js
sfount: After discussion we don't think a half-way implementation of this would be helpful or more readable, this looks like other validation code and we should only significantly differ if we've got guidance on how to do this pattern consistently

app/controllers/registration/registration.controller.test.js
sfount: https://github.com/alphagov/pay-selfservice/pull/3622#discussion_r1025298260

PR #3625, opened 2022-11-18
PP-10054 Add choose second factor method routes

stephencdaly:
Add a POST `/register/get-security-codes` route that validates that a radio button option has been selected and redirects to either the page to enter a phone number, or to set up an authenticator app depending on the option chosen.

## Review notes

Hiding whitespace changes from the diff will help with review, as indentation had changed in the cypress test.


SandorArpa (approved):
LGTM

PR #3633, opened 2022-11-21
PP-10052 Improve validation messages for OTP codes

stephencdaly:
Improve validation messages for OTP codes to be more specific about why a code is invalid if it does not have the correct format.

Uses the error messages from the design system page for this pattern https://design-system.service.gov.uk/patterns/confirm-a-phone-number/

Also, use this OTP validation when an OTP code is entered for changing the sign-in method. Previously we only checked whether the code was empty.


james-peacock-gds (commented):
LGTM!

james-peacock-gds (approved):

PR #3634, opened 2022-11-21
PP-10052 Use constants for invalid OTP error messages

stephencdaly:
To ensure that we are using the same error message in all places that validate OTP codes, move the validation messages we display when adminusers returns a 401 response when verifying the code to shared constants.

Also, display an improved error message if the OTP method is to use an authenticator app. This message is taken from the GOV.UK Sign-in login/registration pages.

### DIFF
--- a/app/controllers/user/two-factor-auth/post-configure.controller.test.js
+++ b/app/controllers/user/two-factor-auth/post-configure.controller.test.js
@@ -86,7 +86,7 @@ describe('Configure new second factor method post controller', () => {
     })
 
     describe('Adminusers returns a 401 response', () => {
-      it('should call redirect with errors in session', async () => {
+      it('should call redirect with errors in session for SMS method', async () => {
         const error = new RESTClientError('An error', 'adminusers', 401)
         const adminusersRejectsStub = () => Promise.reject(error)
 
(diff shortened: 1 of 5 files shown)


app/controllers/user/two-factor-auth/post-configure.controller.test.js
```diff
@@ -101,6 +101,24 @@ describe('Configure new second factor method post controller', () => {
...
+        const adminusersRejectsStub = () => Promise.reject(error)
+
+        const controllerWithAdminusersError = getController(adminusersRejectsStub)
+        await controllerWithAdminusersError(req, res, next)
+
+        sinon.assert.calledWith(res.redirect, paths.user.profile.twoFactorAuth.configure)
+        expect(req.session.pageData).to.have.property('configureTwoFactorAuthMethodRecovered')
+        expect(req.session.pageData.configureTwoFactorAuthMethodRecovered).to.deep.equal({
+          errors: {
+            securityCode: 'The security code you entered is not correct, try entering it again or wait for for your authenticator app to give you a new code'
```
kbottla, line 118:
```suggestion
            securityCode: 'The security code you entered is not correct, try entering it again or wait for your authenticator app to give you a new code'
```

iqbalgds (dismissed):

PR #3642, opened 2022-11-22
BAU Move pact tests

stephencdaly:
The pact tests previously lived under `/test/unit/clients`.

As we have moved other unit tests out of the `/test/unit` directory, move the pact tests to be under `/test/pact` so they are easier to find.


kbottla (approved):

SandorArpa (approved):
LGTM

PR #3643, opened 2022-11-22
PP-10052 Add logic for route to display authenticator app page

stephencdaly:
A route already existed for GET `/register/authenticator-app` which just displayed dummy data. Update the controller for this to retrieve the invite from adminusers and then render the page, displaying the OTP secret key from the invite and the QR code generated from this.


SandorArpa (approved):
LOKTM

PR #3645, opened 2022-11-22
PP-10199 Update pact test for new field `takes_payments_over_phone`

kbottla:
## WHAT
- Updated pact tests for the new field `takes_payments_over_phone` added to service in adminusers


SandorArpa (approved):
LGTM

PR #3646, opened 2022-11-22
PP-10052 Add route to configure authenticator app

stephencdaly:
Add POST `/register/authenticator-app` route.

This route:
- validates the OTP code, and redirected to the GET route with errors in the session if it is invalid
- calls the adminusers endpoint to verify the OTP code
- if verification is successful, calls the adminusers endpoint to complete the endpoint
- redirects to the route to log the user in


app/controllers/registration/registration.controller.js
```diff
@@ -102,15 +106,55 @@ async function showAuthenticatorAppPage (req, res, next) {
...
+    errors[OTP_CODE_FIELD_NAME] = validationResult.message
+    sessionData.recovered = { errors }
+    return res.redirect(paths.register.authenticatorApp)
+  }
+
+  try {
+    await adminusersClient.verifyOtpForInvite(sessionData.code, otpCode)
+    const completeInviteResponse = await adminusersClient.completeInvite(sessionData.code)
+    const userExternalId = completeInviteResponse.user_external_id
+    loginController.setupDirectLoginAfterRegister(req, res, userExternalId)
```
kbottla, line 142: Any reason controller is calling another controller? Can it be in a service method ?

Looks like the only statement required from `setupDirectLoginAfterRegister` is `req.register_invite.userExternalId = userExternalId` as userExternalId will never be null when calling from `registration.controller.js`

stephencdaly: Yes I copied this code from the existing controllers and was meaning to change this in another PR, but actually probably makes sense to change that as part of this PR so its not forgotten about. Will make the change.

stephencdaly: updated

kbottla (approved):

PR #3651, opened 2022-11-23
PP-10199 Set latest go live stage in the Zendesk email

kbottla:
## WHAT
- Currently, zendesk emails for go-live requests include the last stage service is in (which is usually CHOOSEN_PSP..) rather than the new stage (TERMS_AGREED_..) that will be set when the terms are agreed.
- Updated so that the email will include TERMS_AGREED.. stage which will match with Toolbox.


SandorArpa (approved):
LGTM

PR #3660, opened 2022-11-25
PP-10056 Add POST route for resend code page

stephencdaly:
Add a POST route for resending the OTP code. The page asks for the user to confirm their phone number before resending the code. When the submit button is clicked:
- Validate the phone number
- Make a PATCH request to adminusers to update the phone number on the invite
- Make a POST request to adminusers to send an OTP code to the entered phone number
- Redirect to the page for the user to enter their security code.


alan-gds (approved):
LGTM

PR #3667, opened 2022-11-29
PP-10395 Show success page when registration is complete

stephencdaly:
After submitting a valid OTP code on either the page to configure SMS as the second factor method or Authenticator app as the second factor method, redirect to the success page. This page informs the user that registration was successful and has a button to continue to the "My services" page.

When the `/register/success` route is visited, call the middleware that will log the user in using the `userExternalId` from the `register_invite` cookie. Previously this was done as part of the redirect to `/proceed-to-login` which logged the user in then called the controller to render the "My services" page.


alan-gds (approved):
LGTM

PR #3673, opened 2022-11-30
PP-10395 Handle revisiting the success page better

stephencdaly:
- Previously, if the /register/success page was refreshed or visited later, an error page was shown because the registration cookie is no longer present.
- Make it so that the route doesn't require the registration cookie to be present. If it is present, the `login-after-register` middleware will log the user in and then the success page will be shown. If it is not present, the user will either be redirected to the "My services" page if they are logged in, or the log in page if they are not.

- Rename login-after-register.controller.js to login-after-register.middleware.js


app/routes.js
```diff
@@ -21,6 +21,7 @@ const restrictToStripeAccountContext = require('./middleware/stripe-setup/restri
 const restrictToSwitchingAccount = require('./middleware/restrict-to-switching-account')
 const uploadGovernmentEntityDocument = require('./middleware/multer-middleware')
 const inviteCookieIsPresent = require('./middleware/invite-cookie-is-present')
+const loginAfterRegister = require('./middleware/login-after-register.middleware')
```
sfount, line 24: Agree with the middleware naming -- as it's a one line middelware I wonder if it's a little clearer to just inline here in routes 
```javascript
app.get(
  registerUser.logUserIn,
  passport.authenticate('localDirect', { failureRedirect: user.logIn }),
  userIsAuthorised,
  rootController.get
)
```

I'm not entirely convinced by this and could see pros and cons either way.

app/services/auth.service.js
```diff
@@ -90,19 +90,23 @@ async function localStrategy2Fa (req, done) {
...
-    })
+async function localDirectStrategy (req, done) {
+  const registrationSession = req[INVITE_SESSION_COOKIE_NAME]
+  try {
+    if (!(registrationSession && registrationSession.userExternalId)) {
+      return done(null, false)
+    }
+
+    const user = await userService.findByExternalId(registrationSession.userExternalId)
+    lodash.set(req, 'gateway_account.currentGatewayAccountId', lodash.get(user, 'serviceRoles[0].service.gatewayAccountIds[0]'))
```
sfount, line 101: Do we need to do this anymore now that users are landing on the services page? I'm guessing this would then be set by the user following a link into an individual account. I might be missing something!

test/cypress/integration/registration/register.cy.test.js
```diff
@@ -256,4 +256,29 @@ describe('Register', () => {
...
+
+        cy.visit('/register/success')
+        cy.title().should('eq', 'Choose service - GOV.UK Pay')
+      })
+    })
+
+    describe('There is no logged in user', () => {
+      it('should redirect to the login page', () => {
+        cy.visit('/register/success')
+        cy.title().should('eq', 'Sign in to GOV.UK Pay')
```
sfount, line 280: 👍

sfount (approved):
Change looks good! I think there's a conflict that needs resolving

app/routes.js
stephencdaly: The main advantage I can see of it living in its own file is it has a bit more of a descriptive name of what it does. Although maybe the strategy could be something named better than "localDirect" and that would fix it?

The disadvantage of having its own middleware file is you get to that file and it's just calling another middleware - which is a bit confusing.

Do you think inlining and renaming the strategy to "loginAfterRegistration" would be a good way to do it?

app/services/auth.service.js
stephencdaly: Yes I don't think we need this anymore! I think this was possibly left around from when we got rid of the gateway account cookie as a temporary thing but never removed? I can look into this quickly and then maybe raise another PR

app/routes.js
sfount: oh yeah sounds like a clean middle ground 👍 I think the 'local' part has some semantic meaning in the context of strategies so maybe we could keep it, something like `localStrategyLoginDirectAfterRegistration`? Bit of a mouthful but I think it keeps the context and hopefully is much clearer as to what the strategy is for.

app/routes.js
```diff
@@ -169,7 +170,10 @@ module.exports.bind = function (app) {
   app.post(registerUser.otpVerify, registerController.submitOtpVerify)
   app.get(registerUser.reVerifyPhone, registerController.showReVerifyPhone)
   app.post(registerUser.reVerifyPhone, registerController.submitReVerifyPhone)
-  app.get(registerUser.logUserIn, loginController.loginAfterRegister, userIsAuthorised, rootController.get)
+  app.get(
+    registerUser.logUserIn,
+    passport.authenticate('localStrategyLoginDirectAfterRegistration', { failureRedirect: '/login' }),
```
sfount, line 175:
```suggestion
    passport.authenticate('localStrategyLoginDirectAfterRegistration', { failureRedirect: users.logIn }),
```

app/services/auth.service.test.js
```diff
@@ -317,23 +290,67 @@ describe('auth service', function () {
...
-          expect(registerInviteCookie.destroy.called).to.equal(true)
-          expect(req.session.secondFactor).to.equal('totp')
-          expect(doneSpy.calledWithExactly(null, user)).to.equal(true)
-          expect(req.session.version).to.equal(1)
-          done()
-        })
-        .catch(done)
+      await getServiceWithMockedUserService(userServiceMock).localStrategyLoginDirectAfterRegistration(req, doneSpy)
+
+      sinon.assert.calledWith(findByExternalIdSpy, userExternalId)
```
sfount, line 301: 🙌

sfount (approved):
Looks great 👍

app/routes.js
stephencdaly: Thanks, just pushed another commit to fix this

sfount (approved):

PR #3677, opened 2022-12-01
PP-10052 Fix QR code for configuring authenticator app

stephencdaly:
- When scanning the QR code with an authenticator app, the app was suggesting to create an account with name `GOV.UK Pay: username` rather than with the actual username. Fix this to use the email from the invite.


alan-gds (approved):
Looks good to me - I ran the service locally to reproduce the behaviour on master and see it working correctly on the branch.

PR #3685, opened 2022-12-06
BAU Log when a user completes registration

stephencdaly:
Log a line when a user completes a registration. Include in the log line the second factor method chosen and whether the user was an internal GDS user or not.

Include the `internal_user` field for all lines logged for the registration pages, determining this from the email address used for the registration.

### DIFF
--- a/app/utils/email-tools.js
+++ b/app/utils/email-tools.js
@@ -2,11 +2,20 @@
 
 const rfc822Validator = require('rfc822-validate')
 
-module.exports = (email) => {
+function isValidEmail (email) {
   if (!rfc822Validator(email)) {
     return false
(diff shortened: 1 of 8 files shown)


app/utils/email-tools.js
```diff
@@ -2,11 +2,20 @@
 
 const rfc822Validator = require('rfc822-validate')
 
-module.exports = (email) => {
+function isValidEmail (email) {
```
kbottla, line 5: Do we have unit tests for this method anywhere?

stephencdaly: I've added some now

sfount (approved):

PR #3701, opened 2022-12-12
Pp 10313 switch psp flex tasklist

iqbalgds:
- `Switch PSP` page - add new task when switching to Worldpay
  - Users must provide 3dS flex credentials.
  - This task links to the existing `3DS flex page`.
- Add Cypress tests to check this functionality.
  - Make sure it displays the tasks in the right state.
  - Should only enable the `verify with live payment` when all
    the right tasks have been completed.
  - Test it correctly goes to the 3DS flex page from the task.
- Update the `Cypress > gateway account stubs > post 3DS flex settings stub`
  - Clean up code.
  - Make it more flexible by passing down the options parameter to calling functions.
  - Update the `your psp` Cypress test - to work with the latest updates to this method.

### DIFF
--- a/app/controllers/switch-psp/switch-tasks.service.test.js
+++ b/app/controllers/switch-psp/switch-tasks.service.test.js
@@ -15,22 +15,30 @@ describe('Switching PSP service', () => {
         })
         const targetCredential = getSwitchingCredential(account)
         const taskList = getTaskList(targetCredential, account)
-        expect(Object.keys(taskList)).to.have.length(2)
+        expect(Object.keys(taskList)).to.have.length(3)
         expect(taskList.LINK_CREDENTIALS.enabled).to.equal(true)
         expect(taskList.LINK_CREDENTIALS.complete).to.equal(false)
+        expect(taskList.LINK_FLEX_CREDENTIALS.enabled).to.equal(true)
+        expect(taskList.LINK_FLEX_CREDENTIALS.complete).to.equal(false)
+        expect(taskList.VERIFY_PSP_INTEGRATION.enabled).to.equal(false)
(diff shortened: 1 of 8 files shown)


app/controllers/switch-psp/switch-tasks.service.test.js
```diff
@@ -62,6 +72,34 @@ describe('Switching PSP service', () => {
       expect(isComplete(taskList)).to.equal(false)
     })
 
+    it('correctly calculates progress required for Worldpay when 3ds is not enabled', () => {
```
kbottla, line 75:
```suggestion
    it('should correctly calculate progress required for Worldpay when 3ds is not enabled', () => {
```

app/controllers/switch-psp/switch-tasks.service.test.js
```diff
@@ -62,6 +72,34 @@ describe('Switching PSP service', () => {
...
+        ],
+        requires3ds: false,
+        integrationVersion3ds: 2
+      })
+      const targetCredential = getSwitchingCredential(account)
+      const taskList = getTaskList(targetCredential, account)
+      expect(isComplete(taskList)).to.equal(false)
+    })
+
+    it('correctly calculates progress required for Worldpay when it only has 3ds version 1', () => {
```
kbottla, line 89:
```suggestion
    it('should correctly calculate progress required for Worldpay when account 3ds version is 1', () => {
```

kbottla (commented):
Looks good. Just a few minor comments.

Would be good to explain in the commit about `WHY` the change/ describe WHAT functionality is changing/added.

It is clear from the code changes what files are updated so don't think we need to list every file changed in the commit (unless they are not obvious to comprehend from code changes).

kbottla (approved):

PR #3704, opened 2022-12-15
PP-10099 Extend URL validation used for organisation website address

sfount:
Our current method of URL validation has allowed a number of invalid URLs to pass. This results in the URL being passed to downstream services which can reject it in an unpredictable way where the error isn't well handled.

The two outliers:
- spaces in the URL allowing multiple URLs to be enterered (neither expected nor wanted)
- incomplete URLs without a TLD specified

The `validatorjs/validator` library includes some fairly gnarly and well tested regex to ensure that TLD can be required and are well formed. This library is also actively maintained, quite popular and has no dependencies.

Include the validator library and use this to catch the exceptional cases provided above.

### DIFF
--- a/app/utils/validation/server-side-form-validations.js
+++ b/app/utils/validation/server-side-form-validations.js
@@ -4,6 +4,7 @@ const moment = require('moment-timezone')
 const ukPostcode = require('uk-postcode')
 const commonPassword = require('common-password')
 const lodash = require('lodash')
+const validator = require('validator')
 const { URL } = require('url')
 
 const {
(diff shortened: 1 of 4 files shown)


app/controllers/kyc/organisation-url/post.controller.js
```diff
@@ -22,6 +22,9 @@ function validateOrgUrl (organisationUrl) {
 
   const organisationUrlValidationResult = validateUrl(organisationUrl)
   if (!organisationUrlValidationResult.valid) {
+    logger.info('Blocked invalid organisation URL', {
```
kbottla, line 25: This functionality is not used anymore and will be removed soon so we don't need it here.

kbottla: We validate here https://github.com/alphagov/pay-selfservice/blob/38266a49c7f75bb70cd637328aab6912fc0bb7f5/app/controllers/request-to-go-live/organisation-address/post.controller.js#L79 as part of go-live flow

kbottla (approved):

PR #3706, opened 2022-12-15
PP-10338: Remove references to invite type

alan-gds:
The `type` field in invites returned from adminusers is going to be removed; use the `invite_to_join_service`
 flag instead.

N.b. The field is `isInviteToJoinService` in adminusers, but is getting serialised without the 'is' because it's a boolean.  I think this looks a bit confusing at the selfservice end - would it be worth me changing it in adminusers?  I presume I would be able to override the renaming there.

### DIFF
--- a/app/controllers/invite-validation.controller.js
+++ b/app/controllers/invite-validation.controller.js
@@ -30,12 +30,10 @@ async function validateInvite (req, res, next) {
     }
 
     if (invite.user_exist) {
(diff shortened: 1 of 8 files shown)


stephencdaly (commented):
Looks good! Agreed we would change the field returned by adminusers to be `is_invite_to_join_service` to make it clearer.

stephencdaly (approved):
LGTM!

PR #3707, opened 2022-12-15
PP-10465 Change where we re-provision the OTP key

stephencdaly:
Previously we were reprovisioning the OTP key for an invite when:
- The user visited the configure authenticator app page
- The user submitted their phone number

The problem with this was that when the configure authenticator app page was reloaded due to an invalid security code being submitted, the OTP key would be regenerated. This would mean the user would have to configure a new account in their app using the new key, which is not obvious or user friendly.

Instead, reprovision the code when the user submits the page to choose their OTP method. This means that the OTP key would be reprovisioned if the user follows the in-page links to change their method.

There is a small risk that the OTP key wouldn't be provisioned if the user clicks back in their browser a few times after choosing one method and then a different method - or if the user visits the URL for configuring their authenticator app or phone number directly. However, the repercussions and risk if they were to do this are low.

### DIFF
--- a/app/controllers/registration/registration.controller.js
+++ b/app/controllers/registration/registration.controller.js
@@ -125,7 +125,9 @@ function showChooseSignInMethodPage (req, res) {
   res.render('registration/get-security-codes')
 }
 
-function submitChooseSignInMethodPage (req, res) {
+async function submitChooseSignInMethodPage (req, res, next) {
+  const sessionData = req[INVITE_SESSION_COOKIE_NAME]
+
   const signInMethod = req.body['sign-in-method']
   if (!signInMethod) {
     return res.render('registration/get-security-codes', {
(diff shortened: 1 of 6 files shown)


app/controllers/registration/registration.controller.js
```diff
@@ -135,18 +137,23 @@ function submitChooseSignInMethodPage (req, res) {
     })
   }
 
-  if (signInMethod === 'SMS') {
-    res.redirect(paths.register.phoneNumber)
-  } else {
-    res.redirect(paths.register.authenticatorApp)
+  try {
+    await adminusersClient.reprovisionOtp(sessionData.code)
+    if (signInMethod === 'SMS') {
```
sfount, line 142: Another mechanism here to protect against some of the edge cases described in the commit message might be to keep track of what the last method that an OTP was provisioned against was (set anytime that we call this admin users endpoint) and then only re-provision if it's different, something along the lines of

```javascript
if (signInMethod != sessionData.signInMethod) {
  await adminusersClient.reprovisionOtp(sessionData.code)
}
```

sfount (approved):

app/controllers/registration/registration.controller.js
stephencdaly: Ah yes good idea - that might be preferable. I guess we could go back to re-provisioning in the old places but use this cookie to decide.

PR #3714, opened 2022-12-19
PP-10465 Only reprovision OTP key if method has changed

stephencdaly:
We want to avoid the situation where the OTP key is re-provisioned when the page to configure an authenticator APP is reloaded - particularly if it is reloaded due to the security code entered being invalid. If the OTP key is re-provisioned, the user would have to set up a new account in their app with the new OTP, which they are likely to miss - meaning their invite will be disabled.

Store the last selected second factor method in the registration cookie, and only re-provision the OTP key if the method currently being configured does not match the method last time we re-provisioned the OTP key. This means we will only re-provision if the user goes back and selects a different method.


alan-gds (approved):
LGTM

PR #3719, opened 2022-12-20
PP-10349 Hyperlink stripe onboarding tasks

AbdirizakIdris:
- Updated your PSP task list test to check each task is hyperlinked with the correct url and the state  of each task is updated


stephencdaly (approved):
LGTM!

PR #3732, opened 2023-01-05
PP-10397 Add missing stubs to cypress tests

stephencdaly:
We are intending to change the default response when Mountebank does not match on a stub for a request from returning a 200 to returning a 404.

In order to do this, we need to setup some missing stubs for Cypress tests where we currently rely on the default behaviour of getting a 200 when there is no stub, as these tests would fail when the default is 404.

This PR adds missing stubs to several (but not all) Cypress tests.


test/cypress/integration/payment-links/delete-payment-links.cy.test.js
```diff
@@ -1,24 +1,37 @@
...
 const userExternalId = 'a-user-id'
 const gatewayAccountExternalId = 'a-valid-account-id'
 const gatewayAccountId = 42
 const productExternalId = 'a-product-id'
+const apiToken = 'an-api-token'
 
 const product = {
-  external_id: productExternalId
+  external_id: productExternalId,
+  payApiToke: apiToken
```
james-peacock-gds, line 17: typo - payAPIToken

james-peacock-gds (changes requested):
just the typo to fix

test/cypress/integration/payment-links/delete-payment-links.cy.test.js
stephencdaly: Thanks, seems I got interrupted part way through adding the missing stub to this test and it wouldn't have worked. Have pushed e6759f5fb6610ec2188dfb9bfcd8fd97e4c45a08 which fixes this test.

Will squash all the commits in this PR before merging

james-peacock-gds (approved):

PR #3742, opened 2023-01-09
PP-10397 Change mountebank imposter default response

stephencdaly:
Return a 404 when no stub matches a predicate, rather than a 200.

This means that all requests to external services will need to be stubbed when previously we didn't have to when we just expected a success response and didn't rely on the response having a non-empty body.

Add missing stubs to the final tests where we were previously relying on the 200 default response from Mountebank where stubs weren't set up.

### DIFF
--- a/test/cypress/stubs/gateway-account-stubs.js
+++ b/test/cypress/stubs/gateway-account-stubs.js
@@ -212,7 +212,19 @@ function patchAccountEmailCollectionModeSuccess (opts) {
     request: gatewayAccountFixtures.validGatewayAccountEmailCollectionModeRequest('MANDATORY')
   })
 }
+function patchAccountUpdateApplePaySuccess (gatewayAccountId, allowApplePay) {
+  const path = `/v1/api/accounts/${gatewayAccountId}`
+  return stubBuilder('PATCH', path, 200, {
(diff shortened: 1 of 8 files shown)


test/cypress/stubs/gateway-account-stubs.js
```diff
@@ -212,7 +212,19 @@ function patchAccountEmailCollectionModeSuccess (opts) {
...
   })
 }
+function patchAccountUpdateApplePaySuccess (gatewayAccountId, allowApplePay) {
+  const path = `/v1/api/accounts/${gatewayAccountId}`
+  return stubBuilder('PATCH', path, 200, {
+    request: gatewayAccountFixtures.validUpdateToggleApplePayRequest(allowApplePay)
+  })
+}
 
+function patchAccountUpdateGooglePayPaySuccess (gatewayAccountId, allowGooglePay) {
```
kbottla, line 222:
```suggestion
function patchAccountUpdateGooglePaySuccess (gatewayAccountId, allowGooglePay) {
```

kbottla (approved):

test/cypress/stubs/gateway-account-stubs.js
stephencdaly: Good spot. I've got another small PR coming so will fix this there.

PR #3743, opened 2023-01-09
PP-10397 Remove verification of Cypress stubs

stephencdaly:
This was an attempt to verify that stubs were called as part of Cypress testing, but was not really used. Part of the reason for this being added was to ensure that the stub setup was used for requests rather than Mountebank not matching on a stub and returning the default of 200. Now that Mountebank will return a 404 as the default response, this is less necessary.

Remove this verification, as it is probably more confusing than it is helpful.

Also remove the cypress_testing.md document as the information in this now exists in the team manual.


test/cypress/stubs/gateway-account-stubs.js
```diff
@@ -335,5 +334,5 @@ module.exports = {
   postUpdateNotificationCredentialsSuccess,
   postSwitchPspSuccess,
   patchAccountUpdateApplePaySuccess,
-  patchAccountUpdateGooglePayPaySuccess
+  patchAccountUpdateGooglePaySuccess
```
iqbalgds, line 337: Nice - 👍

iqbalgds (approved):

james-peacock-gds (approved):
LGTM!

PR #3744, opened 2023-01-09
PP-10485 Show page not found when ledger returns 404

stephencdaly:
Ledger returns a 404 when the search agreements API endpoint is called with a page number that does not exist. Handle this by showing a not found error page to the user with status code 404.


SandorArpa (approved):
LGTM

PR #3751, opened 2023-01-16
PP-10242 Privacy page - fix content bugs

iqbalgds:


stephencdaly (approved):
Haven't checked the content against the doc, but the HTML looks fine

PR #3752, opened 2023-01-17
PP-10352 redirect user when VAT complete

AbdirizakIdris:
- Updated VAT's post controller so when ENABLE_STRIPE_ONBOARDING_TASK_LIST is set to true it redirects to task-list page
- Updated VAT's cypress test


iqbalgds (approved):

iqbalgds (approved):

PR #3754, opened 2023-01-18
PP-10356 redirect user when company registration complete

AbdirizakIdris:
-  Updated Company number's post controller so when ENABLE_STRIPE_ONBOARDING_TASK_LIST flag is set to true it redirects to task-list page

### DIFF
--- a/app/controllers/stripe-setup/company-number/post.controller.js
+++ b/app/controllers/stripe-setup/company-number/post.controller.js
@@ -19,6 +19,7 @@ const COMPANY_NUMBER_FIELD = 'company-number'
 
 module.exports = async (req, res, next) => {
   const isSwitchingCredentials = isSwitchingCredentialsRoute(req)
+  const enabledStripeOnboardingTaskList = (process.env.ENABLE_STRIPE_ONBOARDING_TASK_LIST === 'true')
   const stripeAccountSetup = req.account.connectorGatewayAccountStripeProgress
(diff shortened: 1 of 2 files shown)


app/controllers/stripe-setup/company-number/post.controller.js
```diff
@@ -62,6 +63,8 @@ module.exports = async (req, res, next) => {
       })
       if (isSwitchingCredentials) {
         return res.redirect(303, formatAccountPathsFor(paths.account.switchPSP.index, req.account.external_id))
+      } else if (enabledStripeOnboardingTaskList) {
+        return res.redirect(303, formatAccountPathsFor(paths.account.yourPsp.index, req.account && req.account.external_id, req.account.gateway_account_credentials[0].external_id))
```
stephencdaly, line 67: We should have the credential ID in the path params for the request. 
You should be able to access it from `req.params.credentialId` - this is because it's specified as a path param in the paths file `/your-psp/:credentialId/company-number` with the key "credentialsId".

I think this is better than assuming it will be the only credential id (even though this should always be the case)

stephencdaly (commented):
Looks good, just a small comment. Happy to discuss.

stephencdaly (approved):
Looks good. We should make sure we update the redirect for the other controllers to use `req.params.credentialId` but that is fine in a separate PR.

PR #3757, opened 2023-01-19
PP-10353 return user when service director task complete

AbdirizakIdris:
- Updated post controller so that when ENABLE_STRIPE_ONBOARDING_TASK_LIST flag is on and service director task is  completed correctly it redirects to the task list page
- Updated unit tests accordingly

### DIFF
--- a/app/controllers/stripe-setup/director/post.controller.test.js
+++ b/app/controllers/stripe-setup/director/post.controller.test.js
@@ -278,4 +278,39 @@ describe('Director POST controller', () => {
     const expectedError = sinon.match.instanceOf(Error)
     sinon.assert.calledWith(next, expectedError)
   })
+
+  it('should redirect to the task list page when ENABLE_STRIPE_ONBOARDING_TASK_LIST is set to true ', async function () {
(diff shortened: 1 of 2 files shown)


app/controllers/stripe-setup/director/post.controller.test.js
```diff
@@ -278,4 +278,42 @@ describe('Director POST controller', () => {
...
+    process.env.ENABLE_STRIPE_ONBOARDING_TASK_LIST = 'false'
+
+    updateCompanyMock = sinon.spy(() => Promise.resolve())
+    setStripeAccountSetupFlagMock = sinon.spy(() => Promise.resolve())
+    const controller = getControllerWithMocks()
+
+    req.body = postBody
+    req.params = {
+      credentialId: 'a-valid-credential-external-id'
+    }
```
stephencdaly, line 311: We can leave out setting the `req.params` for this particular test as it won't be used by the code path taken when the environment variable is set to false.

stephencdaly (dismissed):

stephencdaly (approved):

PR #3768, opened 2023-01-25
PP-10351 upgrade Cypress version to 11.2.0

AbdirizakIdris:
- Context: upgrading cypress version to 11.2.0. Currently we have version 5.0.0 which does not allow us to easily test uploading files when we need to.

- Added new cypress config file as recommended by Cypress docs
- Updated failing Cypress tests


iqbalgds (approved):
Amazing work - going from v5 to v11 !!

package.json
```diff
@@ -119,7 +119,7 @@
     "cheerio": "1.0.0-rc.12",
     "chokidar-cli": "latest",
     "csrf": "^3.1.0",
-    "cypress": "5.0.0",
+    "cypress": "^11.2.0",
```
stephencdaly, line 122: I think we should change this to use an exact version number otherwise GitHub actions might get confused - as we cache the Cypress version used for CI builds using the version number. With the `^`, if a new minor version of Cypress is released, we will install the new minor version locally, but we will still be trying to use the old cached version in CI builds, which might cause some issues.

Semantic versioning used by node is explained here https://docs.npmjs.com/about-semantic-versioning

You can see in the logs for the "Cypress tests" github action run where we restore the cache `Cache restored from key: Linux-cypress-^11.2.0`.

```suggestion
    "cypress": "11.2.0",
```

iqbalgds: Is this something we should do for all NPM modules in `package.json`?
I noticed that Mountebanc has `^`

stephencdaly: I'm not entirely sure on that. I think we started using exact versions for all NPM modules at some point because we have dependabot updating versions for us, but on some projects I've seen that we do use `^`.

Possibly something we need to discuss as a team if we're not quite sure - and write it down. Steven F might have more knowledge about our reasoning for using it/not using it.

For Cypress this is a bit of a special case as the GitHub action that runs Cypress tests extracts this version number from the package.json file and uses it as the key for cacheing Cypress - so the CI build doesn't have to download and unzip Cypress every time which can take a minute or two. You can see this in the GitHub actions workflow file here https://github.com/alphagov/pay-selfservice/blob/master/.github/workflows/run-tests.yml#L121

stephencdaly (approved):

stephencdaly (approved):

PR #3771, opened 2023-01-27
PP-10413 remove gateway account cookies

SandorArpa:
## WHAT
We are including an account external ID in URLs for gateway account-specific pages We now send users to the “My services” page on login.
- Retire the gateway_account cookie so we don't need to maintain special code to handle users who still have the old URLs bookmarked forever. Redirect any user visiting an old account URL that does not contain the account ID to the "My services" page.


stephencdaly (approved):
LGTM!

This PR makes it so that we no longer set the `gateway_account` cookie. We already have logic in routes.js that will redirect to "/my-services" when the gateway_account cookie isn't present and a user attempts to visit an account URL without the external ID.

We should make sure we update the logic in `routes.js` to never use the `gateway_account` cookie, but that can be in the next PR.

PR #3776, opened 2023-01-30
PP-10413 remove legacy url reformatting

SandorArpa:
## WHAT
Since we no longer use gateway account cookie we don't reformat legacy urls. Instead we redirect the user to the `My services` page.
- removing reformatting, leaving redirect only.


stephencdaly (approved):
LGTM!

PR #3779, opened 2023-01-31
Pp 10351 removing no longer needed test scripts

AbdirizakIdris:
- Context: upgraded Cypress from version 5.0.0 to 11.2.0 and so the naming for the Cypress files have been changed as recommended by Cypress. Therefore, the scripts also needed to be updated to reflect the changes
- Updated package json to remove unnecessary test scripts


stephencdaly (approved):
LGTM!

PR #3780, opened 2023-02-02
PP-10413 add new endpoint for demo payments

SandorArpa:
## WHAT
Removing gateway cookies broke the user journey for demo payments.
- add a new end point that finds the gateway account external id from the product external id, and redirects the user to the transactions page.


stephencdaly (approved):
Changes look good - just a suggestion about how we handle errors

stephencdaly (approved):

PR #3790, opened 2023-02-15
PP-10350 redirect user when address name complete cypress test

AbdirizakIdris:
- Updated cypress test to redirect to task-list when organisation details are correct

### DIFF
--- a/test/cypress/integration/settings/your-psp-stripe-tasklist.cy.js
+++ b/test/cypress/integration/settings/your-psp-stripe-tasklist.cy.js
@@ -246,6 +246,28 @@ describe('Your PSP Stripe page', () => {
     })
   })
 
+  describe('Confirm Organisation details task', () => {
+    it('should click on Confirm organisation details task and redirect to tasklist page when organisation details are correct', () => {
+      setupYourPspStubs()
+
+      cy.get('span').contains('Confirm your organisation’s name and address match your government entity document').click()
+      cy.get('h1').contains('Check your organisation’s details')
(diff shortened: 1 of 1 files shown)


test/cypress/integration/settings/your-psp-stripe-tasklist.cy.js
```diff
@@ -246,6 +246,28 @@ describe('Your PSP Stripe page', () => {
     })
   })
 
+  describe('Confirm Organisation details task', () => {
+    it('should click on Confirm organisation details task and redirect to tasklist page when organisation details are correct', () => {
+      setupYourPspStubs()
+
+      cy.get('span').contains('Confirm your organisation’s name and address match your government entity document').click()
+      cy.get('h1').contains('Check your organisation’s details')
+      cy.get(':nth-child(1) > .govuk-label').click()
```
stephencdaly, line 255: I think it would be nicer if we had a more specific selector for this - e.g. selecting by text or using a `data-cy` attribute if possible. This is because if this test breaks it might due if we make some HTML/CCS changes, it could be difficult to see what we were trying to click.

See https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

stephencdaly (commented):
Tests look good - small comment about the selector

test/cypress/integration/settings/your-psp-stripe-tasklist.cy.js
AbdirizakIdris: Good spot, updated now

stephencdaly (approved):
LGTM!

PR #3791, opened 2023-02-16
PP-10675:replaced all usages of set-output in run-tests.yml and post-merge.yml workflows.

Rkotyank:
Located and replaced all usages of set-output commands in run-tests.yml and post-merge.yml workflows.

### DIFF
--- a/.github/workflows/post-merge.yml
+++ b/.github/workflows/post-merge.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
       - name: Parse Node version
-        run: echo NVMRC=`cat .nvmrc` >> $GITHUB_OUTPUT
+        run: echo "NVMRC=$(cat .nvmrc)" >> $GITHUB_OUTPUT
         id: parse-node-version
       - name: Setup
         uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c
--- a/.github/workflows/post-merge.yml
+++ b/.github/workflows/post-merge.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
       - name: Parse Node version
-        run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
+        run: echo NVMRC=`cat .nvmrc` >> $GITHUB_OUTPUT
         id: parse-node-version
       - name: Setup
         uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c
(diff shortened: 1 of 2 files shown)


.github/workflows/post-merge.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
       - name: Parse Node version
-        run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
+        run: echo NVMRC=`cat .nvmrc` >> $GITHUB_OUTPUT
```
oswaldquek, line 28: For consistency sake can we make this line same as [this line](https://github.com/alphagov/pay-selfservice/pull/3791/files#diff-7314d0ebbd2e9537ae4889316745b4fd2fa43cb86275c9caae18a86ba228b642R29)? So 
```
echo "NVMRC=$(cat .nvmrc)" >> $GITHUB_OUTPUT
```
This is also the format suggested in https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/.

Rkotyank: I tried that format as suggested prior to this in another repo https://github.com/alphagov/pay-products-ui/pull/2317 which caused the tests to fail, hence had to figure out a different way. This is not failing the test.

oswaldquek: Hmm I'm not convinced as per the reasons I gave before:
* you're using [this format in this same PR](https://github.com/alphagov/pay-selfservice/pull/3791/files#diff-7314d0ebbd2e9537ae4889316745b4fd2fa43cb86275c9caae18a86ba228b642R29)
* `echo "{name}={value}" >> $GITHUB_OUTPUT` is the format recommended by github - https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

katstevens: I agree with Oz - the curly braces format should work here (if it doesn't there's probably something else wrong). Using backticks for evaluating a statement instead of `${ ... }` is not recommended (the backticks are deprecated in bash).

katstevens (approved):

.github/workflows/run-tests.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           restore-keys: ${{ runner.os }}-node-
       - name: Parse Cypress version
         id: parse-cypress-version
-        run: echo "::set-output name=CYPRESS_VERSION::$(jq -r '.devDependencies.cypress' package.json)"
+        run: echo "CYPRESS_VERSION=$(jq -r '.devDependencies.cypress' package.json)" >> $GITHUB_STATE
```
stephencdaly, line 58: I just noticed that it doesn't look like this is working. Prior to this change, in the "Cache Cypress" step the following was logged:
> Cache restored from key: Linux-cypress-11.2.0

But now it is
> Cache restored from key: Linux-cypress-

so doesn't seem to be getting the version

Is it because we're putting it in `GITHUB_STATE` but trying to get it out of the outputs?
```key: ${{ runner.os }}-cypress-${{ steps.parse-cypress-version.outputs.CYPRESS_VERSION }}```

katstevens: Good spot @stephencdaly - I'll raise a PR to fix

katstevens: https://github.com/alphagov/pay-selfservice/pull/3797

PR #3795, opened 2023-02-20
PP-10342 Add back-link to redirect user from each task to the task list

AbdirizakIdris:
-  Added function to look for ENABLE_STRIPE_ONBOARDING_TASK_LIST and yourPsp route
- Added necessary unit tets
-  Updated controller logic for all seven tasks to have a back-link to task-list when EnableStribeOnboardingTaskList flag is true
- Updated nunjucks logic to render back-link when flag is 'true'

### DIFF
--- a/app/controllers/stripe-setup/bank-details/get.controller.js
+++ b/app/controllers/stripe-setup/bank-details/get.controller.js
@@ -1,11 +1,13 @@
 'use strict'
 
 const { response } = require('../../../utils/response')
-const { isSwitchingCredentialsRoute } = require('../../../utils/credentials')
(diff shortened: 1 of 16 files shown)


iqbalgds (dismissed):
Looks good - just not sure if we should be checking the FLAG before adding the back the link.
Otherwise, this will appear on all instances of the task pages.

AbdirizakIdris:
> Looks good - just not sure if we should be checking the FLAG before adding the back the link. Otherwise, this will appear on all instances of the task pages.

Good spot, updated now to check the flag

kbottla (approved):

PR #3796, opened 2023-02-20
PP-10692: Upgrade base image

oswaldquek:
Without the `--force-overwrite` flag we get the following when running in a local docker node 16.19.1 image:
```
/ # apk add --no-cache glibc-2.28-r0.apk
fetch https://dl-cdn.alpinelinux.org/alpine/v3.17/main/aarch64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.17/community/aarch64/APKINDEX.tar.gz
(1/1) Installing glibc (2.28-r0)
ERROR: glibc-2.28-r0: trying to overwrite etc/nsswitch.conf owned by alpine-baselayout-data-3.4.0-r0.
1 error; 14 MiB in 18 packages
```
The fix to add the `--force-overwrite` is found at https://github.com/sgerrand/alpine-pkg-glibc/issues/185.


nlsteers (approved):
👍

PR #3805, opened 2023-02-28
Pp 10343 add progress indicator for task-list cypress

AbdirizakIdris:
- Added logic in nunjucks file to show progress indicator when ENABLE_STRIPE_ONBOARDING_TASK_LIST is true
- Updated cypress test to check this logic
- Updated cypress test to check the completeness of the tasks


stephencdaly (approved):
Looks good to me, just a stray console.log we should make sure to remove

PR #3806, opened 2023-02-28
 PP-10343 Add progress indicator for tasklist cypress tests

AbdirizakIdris:
- Added logic in nunjucks file to show progress indicator when ENABLE_STRIPE_ONBOARDING_TASK_LIST is true
- Updated cypress test to check progress of tasks

### DIFF
--- a/app/views/your-psp/_stripe.njk
+++ b/app/views/your-psp/_stripe.njk
@@ -3,8 +3,8 @@
     <span class="app-task-list__task-name">
       {% if not item.complete %}
         <a class="govuk-link" href="{{ url }}" aria-describedby="{{ id }}-status">
-        {{ label }}
-      </a>
+          {{ label }}
+        </a>
       {% else %}
         <span>{{ label }}</span>
       {% endif %}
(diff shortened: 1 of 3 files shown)


app/views/your-psp/_stripe.njk
```diff
@@ -94,6 +94,12 @@
  {% else %}
  {% if  enableStripeOnboardingTaskList %}
     <h1 class="govuk-heading-m govuk-!-margin-top-8">Information for Stripe</h1>
+    <h3 class="govuk-body"> {{progressIndicator.no_of_tasks_completed}} out of {{progressIndicator.total_number_of_tasks}} steps completed</h3>
```
stephencdaly, line 97: I don't think this should be a h3.
Are we following the guidance in the task list pattern? https://design-system.service.gov.uk/patterns/task-list-pages/

If so, looks like the progress indicator should be below the h2 in a `<p> `- as in this example https://prototype-kit.service.gov.uk/v12/docs/templates/task-list

stephencdaly (commented):
Spotted one thing

app/views/your-psp/_stripe.njk
AbdirizakIdris: You're right, good spot

app/views/your-psp/_stripe.njk
```diff
@@ -91,57 +92,67 @@
     </div>
   </div>
 
- {% else %}
- {% if  enableStripeOnboardingTaskList %}
+{% else %}
+  {% if enableStripeOnboardingTaskList %}
     <h1 class="govuk-heading-m govuk-!-margin-top-8">Information for Stripe</h1>
```
stephencdaly, line 97: We already have a `h1` on this page, and we should only have 1. If we're keeping this header, it should be a h2 and the current h2s should become h3.

app/views/your-psp/_stripe.njk
```diff
@@ -91,57 +92,67 @@
...
+
+    {% if progressIndicator.no_of_tasks_completed < progressIndicator.total_number_of_tasks %}
+      <h2 class="govuk-heading-m govuk-!-margin-top-8">Information incomplete</h2>
```
stephencdaly, line 100: Screenshot 2023-02-28 at 17 05 15

The styling doesn't look quite right here. The h2 should have the following classes `<h2 class="govuk-heading-s govuk-!-margin-bottom-2">` according to the design system guidance.

Are we working from a mocked up design from this page? Is it worth checking with Johnny whether we want all the headers? And if so what styling they should have?

stephencdaly (approved):
LGTM! The two h1s on the page will be addressed in a different PR.

PR #3807, opened 2023-03-02
PP-10343 stripe progress indicator: fix heading styling

AbdirizakIdris:
- Updated heading styling according to the task list design pattern


iqbalgds (approved):

PR #3816, opened 2023-03-07
PP-10347 Add page heading to task list

AbdirizakIdris:
- Updated heading for Stripe onboarding tasks
- Updated necessary cypress tests

### DIFF
--- a/app/views/your-psp/index.njk
+++ b/app/views/your-psp/index.njk
@@ -26,7 +26,12 @@
       type: 'success'
     }) }}
   {% endif %}
-  <h1 class="govuk-heading-l page-title">Your payment service provider (PSP) - {{ credential.payment_provider | formatPSPname }}</h1>
+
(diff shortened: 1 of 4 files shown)


app/views/your-psp/index.njk
```diff
@@ -26,7 +26,13 @@
       type: 'success'
     }) }}
   {% endif %}
+
+  {% if activeCredential.payment_provider == 'stripe' %}
```
iqbalgds, line 30: We need to check for the flag as well

iqbalgds: Also, can we change it to 3 equals.

iqbalgds: ===

iqbalgds: Also the spacing as well.

iqbalgds (approved):

iqbalgds (approved):

PR #3830, opened 2023-03-16
PP-10358 fixing content for Stripe restricted banner

AbdirizakIdris:
- For the dashboard, fix minor content bug to the Stripe restricted banner


iqbalgds (approved):

PR #3835, opened 2023-03-20
 PP-10358 fixing content for the new Stripe dashboard

AbdirizakIdris:
- For the dashboard, update minor content for the Stripe banner


iqbalgds (approved):

PR #3844, opened 2023-03-21
PP-10840 Stripe task list - fix side nav error when updating org address

iqbalgds:
- Pass the flag down that check if the Stripe tasklist is enabled and if on a `your-psp` route to the the error page.
- This flag is used to show the side nav.


AbdirizakIdris (approved):
Looks good to me!

PR #3848, opened 2023-03-22
 PP-10358 updating stripe banners

AbdirizakIdris:
- Before a service could take £2000 of payments before their account was restricted if they hadn’t filled in all the Stripe information.
- Now the account gets restricted striaght away.
- This means that currently the service would always see the 'Stripe has restricted your account. You need to submit additional information to Stripe to take payments.' if they haven’t completed the tasks.
- They will never see 'You need to submit additional information to Stripe to be able to take payments.'
- Therefore removing the restricted scenario when setting up a new Stripe account so that they see the standard Stripe task list banner.
- Also removing the new Stripe account > deadline scenario as in reality this will not occur anymore when the service hasn’t completed the task list.
- However, we are keeping the restricted scenario when a Stripe account has already been set up.


stephencdaly (approved):
LGTM!

PR #3857, opened 2023-03-27
PP-10862 Combine dashboard statistics Cypress tests

stephencdaly:


test/cypress/integration/dashboard/dashboard-statistics.cy.js
```diff
@@ -59,9 +55,8 @@ describe('Account dashboard', () => {
       cy.get('.dashboard-total-group__count').should('not.exist')
       cy.get('.dashboard-total-group__amount').should('have.text', '£97.00')
     })
-  })
 
-  it('should update when a different time range is selected', () => {
+    cy.log('Select a different time range and check statistics are updated')
```
kbottla, line 59: Do we need this log?

PR #3862, opened 2023-03-27
PP-10862 Combine organisation details Cypress tests

stephencdaly:

Remove the `.within(() => {})` blocks from the tests as they made it difficult to read once the tests were combined.

Add in `{ delay: 0 }` to the options when calling `.type()` on an element to speed up text input to make runtime less.

## Review note

Hiding whitespace changes from the diff will make this a lot easier to review

### DIFF
--- a/test/cypress/integration/my-services/organisation-details.cy.js
+++ b/test/cypress/integration/my-services/organisation-details.cy.js
@@ -17,191 +17,156 @@ const invalidTelephoneNumber = 'abd'
 const validUrl = 'https://www.example.com'
 const invalidUrl = 'invalid.url'
 
-describe('Dashboard', () => {
-  beforeEach(() => {
-    // keep the same session for entire describe block
-    Cypress.Cookies.preserveOnce('session')
-  })
+const merchantDetails = {
+  name: validName,
+  address_line1: validLine1,
+  address_line2: validLine2,
+  address_city: validCity,
+  address_country: countryGb,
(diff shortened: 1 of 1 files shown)


sfount (approved):
Great refactor and nice optimisations 👍 I wonder if its worth doing something like (from a Cypress question thread): 

```javascript
Cypress.Commands.overwrite('type', (originalFn, subject, text, options = {}) => {
  options.delay = options.delay || 0;

  return originalFn(subject, text, options)
})
```

And then explicitly slow it down if it needs to simulate typing speed (testing client side js behaviour or something). It does feel a little brittle (assuming the command `type` doesn't change name!) but I don't think there's a global config -- specifying per command looks fine to me as well

stephencdaly:
> Great refactor and nice optimisations 👍 I wonder if its worth doing something like (from a Cypress question thread):
> 
> ```js
> Cypress.Commands.overwrite('type', (originalFn, subject, text, options = {}) => {
>   options.delay = options.delay || 0;
> 
>   return originalFn(subject, text, options)
> })
> ```
> 
> And then explicitly slow it down if it needs to simulate typing speed (testing client side js behaviour or something). It does feel a little brittle (assuming the command `type` doesn't change name!) but I don't think there's a global config -- specifying per command looks fine to me as well

That seems like a great suggestion, thanks! I was looking for a global config.
As we don't use much client-side javascript in selfservice, it should probably be fine in most cases never having a delay?

PR #3865, opened 2023-03-27
PP-9033 Remove director task from addtional KYC data functionality

kbottla:
## WHAT
- Removed director task from the additional KYC data task list


SandorArpa (approved):
LGTM

PR #3866, opened 2023-03-27
Pp 9033 remove remaining kyc functionality

kbottla:
## WHAT
- Removed remaining additional KYC data functionality which includes governtment_entity_task and organisation URL task


nlsteers (approved):
Lots of changes here so hard to completely confident but it looks okay to me. 

Someone with more `selfservice` context may be better placed to know if there's anything else KYC related lurking around but looks like you've got it all.

🗑️ 🔥

PR #3867, opened 2023-03-28
PP-19862 Combine Cypress tests for creating payment links

stephencdaly:

Remove some of the .within(() => {}) blocks from the tests as they made it difficult to read once the tests were combined.

## Review notes

Hiding whitespace changes from the diff will help with reviewing


SandorArpa (approved):
LOKTM

PR #3869, opened 2023-03-28
PP-10862 Combine edit payment link Cypress tests

stephencdaly:

For these tests, I've split them up into an `it` per edit page (information, reference, amount). As these can be tested independently without the session being stored between the tests.

## Review notes


SandorArpa (approved):
LGTM

PR #3872, opened 2023-03-29
PP-10862 Make request to go live tests standalone

stephencdaly:

## Review notes


SandorArpa (approved):
LGTM

PR #3874, opened 2023-03-29
PP-10862 Make settings Cypress tests standalone

stephencdaly:

Simplify payment-types.cy.js to not attempt to check what the new selected card types are after updating. This was only checking that the page has the enabled card types returned by connector selected and not any selfservice logic. Instead, just check that the page is successfully reloaded with a success message.

## Review notes

Hiding whitespace changes will make this slightly easier to review


SandorArpa (approved):
LGTM

PR #3875, opened 2023-03-29
PP-10862 Make switch PSP tests standalone

stephencdaly:

Restructured the tests slightly so that the tests for the individual pages for submitting the switch details are separate from tests for the task list logic. I Did this because the task list logic relies purely on what is returned from connector rather than any client session saved after a step is completed. This means we don't need to set up new stubs part way through a test to simulate the task list being updated after a step is completed.

## Review notes


SandorArpa (approved):
LOKTM

PR #3881, opened 2023-03-29
PP-10862 Make update service name Cypress tests standalone

stephencdaly:

### DIFF
--- a/test/cypress/integration/my-services/update-service-name.cy.js
+++ b/test/cypress/integration/my-services/update-service-name.cy.js
@@ -26,34 +26,30 @@ function setupStubs (serviceName, stubs = []) {
 
 describe('Update service name', () => {
   beforeEach(() => {
-    // keep the same session for entire describe block
-    Cypress.Cookies.preserveOnce('session')
+    cy.setEncryptedCookies(authenticatedUserId)
   })
 
   describe('Edit a service name without a Welsh name', () => {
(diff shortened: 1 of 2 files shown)


test/cypress/integration/my-services/update-service-name.cy.js
```diff
@@ -26,34 +26,30 @@ function setupStubs (serviceName, stubs = []) {
...
-    // keep the same session for entire describe block
-    Cypress.Cookies.preserveOnce('session')
+    cy.setEncryptedCookies(authenticatedUserId)
   })
 
   describe('Edit a service name without a Welsh name', () => {
-    it('should display the my services page', () => {
-      cy.setEncryptedCookies(authenticatedUserId)
-      setupStubs(serviceName)
+    it('should allow updating service name', () => {
```
iqbalgds, line 33: Can we get rid of the function defined on line 19 (setupStubs)?

stephencdaly: It's still used in 2 places, but I guess we could inline it?

iqbalgds: That's fine, I didn't realise it is still being used.
I will approve this PR - as this PR is to make them standalone.

iqbalgds (approved):

PR #3882, opened 2023-03-30
PP-10863 Better mark-up for telephone number inputs

alexbishop1:
Use better mark-up for telephone number fields to help browsers display an appropriate on-screen keyboard for entering phone numbers and autofill appropriate phone numbers.

Specifically, when the input expects a user’s phone number for receiving text messages, use `type="tel"` and `autocomplete="mobile tel"`.

When the input expects a responsible person’s phone number, use `type="text"` and `autocomplete="work tel"`. We use `type="text"` as opposed to `type="text"` because a responsible person’s work phone number may have an extension and an on-screen keyboard triggered by `type="tel"` may not offer a way to enter this (though it’s unclear if Stripe accept extensions in phone numbers).

When the input expects an organisation’s phone number, use `type="tel"` and `autocomplete="work tel"`.

### DIFF
--- a/app/views/registration/phone-number.njk
+++ b/app/views/registration/phone-number.njk
@@ -39,6 +39,8 @@
         },
         id: "phone",
         name: "phone",
+        type: "tel",
+        autocomplete: "mobile tel",
         errorMessage: { text: errors['phone'] } if errors['phone'] else false,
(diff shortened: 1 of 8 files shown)


SandorArpa (approved):
LGTM

PR #3883, opened 2023-03-30
PP-10862 Upgrade to Cypress 12

stephencdaly:
Make changes to make tests pass:
- Cannot call assertions on multiple elements - so when `cy.get` finds multiple matches need to call `first()` or `eq()` before doing assertions
- A blank page is now loaded at the start of each `it`, rather then the previous behaviour of starting on the page the last test loaded. Needed to make sure all tests call `cy.visit`. Combined some tests to deal with this.


iqbalgds (approved):

SandorArpa (approved):
LGTM

PR #3893, opened 2023-04-05
PP-10863 Appropriate attributes for 2FA OTP inputs

alexbishop1:
For inputs where we expect the user to provide a second-factor authentication one-time password (security code), consistently use `type="text"`, `inputmode="numeric"` and `pattern="[0+9]+"` (as recommended by the GOV.UK Design System for non-decimal numbers), along with `autocomplete="one-time-code"`.

Previously we used `autocomplete="off"` in some places. Switching to `autocomplete="one-time-code"` may cause browsers to suggest old OTPs where they did not before but this is a fairly minor issue.

### DIFF
--- a/app/views/login/otp-login.njk
+++ b/app/views/login/otp-login.njk
@@ -49,11 +49,11 @@
         name: "code",
         classes: "govuk-!-width-one-third",
         type: "text",
+        inputmode: "numeric",
+        pattern: "[0-9]*",
+        autocomplete: "one-time-code",
         attributes: {
-          "autofocus": "true",
-          "autocomplete": "off",
-          "inputmode": "numeric",
-          "pattern": "[0-9]*"
+          "autofocus": "true"
         }
       })
     }}
--- a/app/views/registration/authenticator-app.njk
+++ b/app/views/registration/authenticator-app.njk
@@ -42,6 +42,9 @@
         },
         id: "code",
         name: "code",
+        inputmode: "numeric",
+        pattern: "[0-9]*",
+        autocomplete: "one-time-code",
         classes: "govuk-input--width-10",
         errorMessage: { text: errors['code'] } if errors['code'] else false
       }) }}


app/views/login/otp-login.njk
```diff
@@ -49,11 +49,11 @@
         name: "code",
         classes: "govuk-!-width-one-third",
         type: "text",
+        inputmode: "numeric",
+        pattern: "[0-9]*",
+        autocomplete: "one-time-code",
```
kbottla, line 54: Given that the one-time code changes every time, is there any need to set auto-complete for this instead of `off`?

kbottla: Do we know how many users this will benefit? Just thinking if it is going to cause issues if users start to select code based on browser suggestion

alexbishop1: Not sure how many users will benefit.

I did some playing around on my computer (running macOS Ventura 13.3) and it seems the state of affairs is as follows:

- Chrome 112.0.5615.49 seems to suggest previous values for `autocomplete="one-time-code"` but a [fix for this was committed literally a week ago](https://bugs.chromium.org/p/chromium/issues/detail?id=1385892) and will presumably make its way into future versions of Chrome (and other Chromium-based browsers like Microsoft Edge, Opera, Brave etc.)
- Safari 16.4 seems to be smart enough to not to suggest previous values for `autocomplete="one-time-code"`
- Firefox 111.0.1 seems to be smart enough to not suggest previous values for `autocomplete="one-time-code"` and there’s a [test in their source tree that suggests this is intentional](https://searchfox.org/mozilla-central/source/toolkit/components/satchel/test/test_form_submission.html#187-191)

So maybe we should leave it as `autocomplete="off"` for now and reassess in a few months when the Chromium fix has had some time to propagate.

alexbishop1:
Closing in favour of #3901.

PR #3931, opened 2023-04-25
BAU — Do not autofocus OTP input for configuring authenticator app

alexbishop1:
When we ask the user to scan a QR code to set up their authenticator app, do not autofocus the input where they enter the security code generated by their app because there’s loads of content on the page before that.


app/views/two-factor-auth/configure.njk
```diff
@@ -71,12 +71,9 @@
...
         type: "text",
-        attributes: {
-          "autocomplete": "off",
-          "autofocus": true,
-          "inputmode": "numeric",
-          "pattern": "[0-9]*"
-        }
+        autocomplete: "off",
+        inputmode: "numeric",
+        pattern: "[0-9]*"
```
katstevens, line 76: Do these no longer need to be inside the `attributes` block?

alexbishop1: Correct, all three are now top-level options of the Nunjucks macro — this can be seen visiting https://design-system.service.gov.uk/components/text-input/ then selecting ‘Nujucks’ under the first example, followed by ‘Nunjucks macro options’.

katstevens (approved):

PR #3934, opened 2023-04-27
PP-10894 Update Zendesk client - Remove node-zendesk and use axios

iqbalgds:
- Removing `node-zendesk` NPM and replacing with `axios`.
  - In order to use the Zendesk API key, we need to do the following: - Use basic auth as follows: - Username - use the API username with `/token` at the end. - Password - set as API key
- `axios` will automatically detect if the `http_proxy` ENV variable is set and use that to proxy requests.
  - The `http_proxy` ENV is set on the AWS environments.
  - When running it locally, no proxy is used.
- No unit tests required for the Zendesk client as other tests cover this functionality.


app/services/clients/zendesk.client.js
```diff
@@ -1,33 +1,42 @@
...
+        username: process.env.ZENDESK_USER + '/token',
+        password: process.env.ZENDESK_API_KEY
+      }
     }
-  })
+
+    await axios(zendeskOptions)
+  } catch (error) {
+    throw error
+  }
```
sfount, line 39: If we're just rethrowing the error it's fine to let this error go without catching it

app/services/clients/zendesk.client.js
```diff
@@ -1,33 +1,42 @@
...
+      url: process.env.ZENDESK_URL + '/tickets',
+      data: zendeskTicketData,
+      auth: {
+        username: process.env.ZENDESK_USER + '/token',
+        password: process.env.ZENDESK_API_KEY
+      }
     }
-  })
+
+    await axios(zendeskOptions)
```
sfount, line 36: The previous method this is replacing returns the created zendesk ticket once it is created, this changes the functionality by returning nothing (I'm guessing the created ticket is then not used by the rest of the code, just something to be aware of)

app/services/clients/zendesk.client.js
```diff
@@ -1,33 +1,42 @@
...
-      comment: {
-        body: opts.message
-      },
-      group_id: zendeskConfig.GROUP_ID,
-      organization_id: zendeskConfig.ORG_ID,
-      ticket_form_id: zendeskConfig.FORM_ID,
-      tags: opts.tags
+    const zendeskOptions = {
+      method: 'post',
+      url: process.env.ZENDESK_URL + '/tickets',
```
sfount, line 28: Although we do this elsewhere in self service we might want to use url join here as we can't strictly guarantee the format of the `ZENDESK_URL` provided by the environment, it should be roughly the same lines of code

```
const urlJoin = require('url-join')

const zendeskOptions = {
  url: urlJoin(process.env.ZENDESK_USER, '/token')
}
```

This will avoid any issues with trailing or not present `/` in the URL

app/services/clients/zendesk.client.js
```diff
@@ -1,33 +1,42 @@
...
+const axios = require('axios')
 const zendeskConfig = require('../../../config/zendesk')
 
-const zendeskClient = zendesk.createClient({
-  username: process.env.ZENDESK_USER,
-  token: process.env.ZENDESK_API_KEY,
-  remoteUri: process.env.ZENDESK_URL,
-  proxy: process.env.http_proxy
-})
+async function createTicket (opts) {
```
sfount, line 6: As this method isn't currently waiting for anything or handling any errors it could be simplified a bit
```js
function createTicket(opts) {
  const zendeskTicketData = {
    // ...
  }
  return axios({
    method: 'post',
    data: zendeskTicketData,
    // ...
  })
}
```

sfount (commented):
Change looks good, we use axios on other node repositories and so developers should have experience with this interface.

A few fairly minor suggestions that might help the code read more clearly and anticipate some corner cases

app/services/clients/zendesk.client.js
iqbalgds: Thanks - just removed this

iqbalgds:
I have made the requested changes - please take a look.

sfount (approved):
Change looks clean, great work 👍 This looks like something we'll want to review in pre-prod environments to make sure the proxy settings are all picked up correctly by axios, advise we pin the CD release in staging to review

iqbalgds:
Good idea - I will pin the release in staging

PR #3942, opened 2023-05-02
PP-10951: Update qs to 6.11.1 and @cypress/request to 2.88.11

katstevens:
Cypress has its own fork of request as a dependency - we don't use this in our tests so it should be safe to override.

~TODO: figure out why the npm has mangled the package-lock version for `qs` in one place (`"version": "https://registry.npmjs.org/qs/-/qs-6.5.3.tgz",`). [See equivalent in products-ui](https://github.com/alphagov/pay-products-ui/pull/2417/files).~ Fixed by refreshing the package-lock.json.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1771,9 +1771,9 @@
       }
     },
     "node_modules/@cypress/request": {
-      "version": "2.88.10",
-      "resolved": "https://registry.npmjs.org/@cypress/request/-/request-2.88.10.tgz",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -28523,8 +28532,7 @@
           }
         },
         "qs": {
-          "version": "6.5.3",
-          "resolved": "https://registry.npmjs.org/qs/-/qs-6.5.3.tgz",
+          "version": "https://registry.npmjs.org/qs/-/qs-6.5.3.tgz",
```
katstevens, line 28535: 🤔

katstevens: Refreshing the package-lock removes this.

jfharden (approved):

PR #3964, opened 2023-05-19
PP-11020 Update webhook calls to pass `gateway account id`

iqbalgds:
- We now require the gateway account ID when calling webhook functions.
- This PR is to update calls to webhooks.
- The backend changes will be done once this is merged in.

### DIFF
--- a/app/controllers/webhooks/webhooks.controller.js
+++ b/app/controllers/webhooks/webhooks.controller.js
@@ -17,7 +17,7 @@ async function webhookDetailPage (req, res, next) {
   const page = req.query.page || 1
 
   try {
-    const webhook = await webhooksService.getWebhook(req.params.webhookId, req.service.externalId)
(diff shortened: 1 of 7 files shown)


SandorArpa (approved):
LGTM

app/controllers/webhooks/webhooks.controller.js
```diff
@@ -17,7 +17,7 @@ async function webhookDetailPage (req, res, next) {
   const page = req.query.page || 1
 
   try {
-    const webhook = await webhooksService.getWebhook(req.params.webhookId, req.service.externalId)
+    const webhook = await webhooksService.getWebhook(req.params.webhookId, req.service.externalId, req.account.external_id)
```
sfount, line 20: This would ideally be what all of the different services would use but unfortunately they talk in internal account identifiers. Everywhere it's specified here should be `req.account.gateway_account_id`

PR #3982, opened 2023-06-02
PP-10601 Cancel agreement - add cancel functionality to agreeement controller

iqbalgds:
- Add cancel agreement method to the agreement controller.
  - This calls the `agreement service > cancel agreement` function
  - Then calls the `agreement details` function.
    - As we want to to show agreement details page after someone has cancelled an agreement.
  - Also pass a `isCancel` flag to the `agreement details` function that will be used to show a notification that the agreement was cancelled.


SandorArpa (approved):
LGTM

PR #3991, opened 2023-06-06
PP-10601 Cancel agreement - showing cancel functionality

iqbalgds:
- Update agreement controller
  - Set flag to only show cancel functionality when agreement is active and user has `agreements:update` functionality
  - Rename existing `isCancel` flag to `isCancelAgreementSuccess` so it is more descriptive.


SandorArpa (approved):
LGTM

PR #3993, opened 2023-06-07
PP-10601 Cancel agreement - update agreement controller

iqbalgds:
- Refactor when cancelling an agreement
  - So that it redirects to the agreement detail page rather than calling the agreement detail page directly.
- Use the req.flash method to set the `cancel success` notification


SandorArpa (approved):
LGTM

PR #3994, opened 2023-06-07
PP-10601 Cancel agreement - add Cypress test

iqbalgds:
- Create new Cypress test for the agreement detail page.
- Make one big Cypress test as per best practises from the team manual: https://pay-team-manual.cloudapps.digital/manual/how-to/write-cypress-tests.html#having-tests-rely-on-the-state-of-previous-tests
  - Load the agreement detail page.
  - Cancel the agreement
  - Reload the agreement detail page and make sure the cancel functionality is no longer available.

### DIFF
--- a/test/cypress/integration/agreements/agreement.cy.js
+++ b/test/cypress/integration/agreements/agreement.cy.js
@@ -1,0 +1,78 @@
+const userStubs = require('../../stubs/user-stubs')
+const gatewayAccountStubs = require('../../stubs/gateway-account-stubs')
+const agreementStubs = require('../../stubs/agreement-stubs')
+const transactionStubs = require('../../stubs/transaction-stubs')
+
+const userExternalId = 'some-user-id'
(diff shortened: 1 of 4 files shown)


test/cypress/integration/agreements/agreement.cy.js
```diff
@@ -0,0 +1,78 @@
...
+    { reference: 'payment-reference', amount: 1000, type: 'payment' },
+    { reference: 'second-reference', amount: 20000, type: 'payment' }
+  ],
+  filters: {
+    agreement_id: 'a-valid-agreement-id',
+    display_size: 5
+  }
+})
+
+describe('Agreements', () => {
```
kbottla, line 33: Could increase test coverage here. ex:
- the user does not have permission to cancel the agreement
- when cancel fails ( connector returns error) 

Can be in a separate PR

kbottla (approved):

PR #3995, opened 2023-06-08
BAU: Resolve ejs/mountebank-formatters conflict

katstevens:
## WHAT
Resolves a long running issue with `ejs` (see https://github.com/alphagov/pay-selfservice/security/dependabot/34). Solving this should reduce noise in our dependency alerts.

See the equivalent PR in products-ui: https://github.com/alphagov/pay-products-ui/pull/2471

## HOW 
Check if the cypress tests pass on this PR!


SandorArpa (approved):
LGTM

PR #3998, opened 2023-06-08
PP-10601 Cancel agreement - add more Cypress test coverage

iqbalgds:
- Add tests for:
  - If the user does not have correct permissions.
  - If the cancel agreement call to connector fails.
- Some minor code cleanup

### DIFF
--- a/test/cypress/integration/agreements/agreement.cy.js
+++ b/test/cypress/integration/agreements/agreement.cy.js
@@ -8,15 +8,22 @@ const gatewayAccountId = 10
 const gatewayAccountExternalId = 'gateway-account-id'
 const serviceExternalId = 'service-id'
 
-const userAndGatewayAccountStubs = [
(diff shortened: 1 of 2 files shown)


test/cypress/integration/agreements/agreement.cy.js
```diff
@@ -75,4 +82,56 @@ describe('Agreements', () => {
...
+        external_id: 'a-valid-agreement-id'
+      }),
+      transactionsStub
+    ])
+
+    cy.visit('/test/service/service-id/account/gateway-account-id/agreements/a-valid-agreement-id', { failOnStatusCode: false })
+
+    cy.get('h1').contains('Agreement detail')
+
+    cy.get('h1').contains('Agreement detail')
```
kbottla, line 134: Looks same as line 132

iqbalgds: Good spot - just updated it.

PR #3999, opened 2023-06-09
PP-10875 Remove request library usage from tests

kbottla:
## WHAT
- Replaced request library usage in tests with Axios. This is to eventually remove the request library dependency which is not maintained anymore

### DIFF
--- a/test/cypress/plugins/index.js
+++ b/test/cypress/plugins/index.js
@@ -7,7 +7,7 @@
 
 'use strict'
 
-const request = require('request-promise-native')
+const axios = require('axios')
 
 const cookieMonster = require('./cookie-monster')
 
@@ -33,11 +33,8 @@ module.exports = (on, config) => {
      * the same call.
      */
     setupStubs (stubs) {
-      return request({
-        method: 'POST',
-        url: mountebankImpostersUrl,
-        json: true,
-        body: {
+      return axios.post(mountebankImpostersUrl,
+        {
           port: config.env.MOUNTEBANK_IMPOSTERS_PORT,
           protocol: 'http',
           defaultResponse: {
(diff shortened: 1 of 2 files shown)


SandorArpa (dismissed):
LOKTM

test/cypress/plugins/index.js
```diff
@@ -33,27 +33,35 @@ module.exports = (on, config) => {
...
-          protocol: 'http',
-          defaultResponse: {
-            statusCode: 404,
-            body: 'No stub predicate matches the request',
-            headers: {}
-          },
-          stubs
-        }
+      return new Promise((resolve, reject) => {
+        axios.post(mountebankImpostersUrl,
```
iqbalgds, line 37: Is it possible to change this to await/async?
As we prefer using that syntax as per our Node style guide:
https://gds-way.cloudapps.digital/manuals/programming-languages/nodejs/#use-of-async-await

iqbalgds (changes requested):
Looks good - left a comment

test/cypress/plugins/index.js
kbottla: Returning promise is fine in this case as this is expected by Cypress https://github.com/cypress-io/cypress/pull/6920/files

iqbalgds: Not sure if this is possible - but can we return a promise but change the axios code to use Async/Await?

E.g.

return new Promise(async (resolve, reject) => {
  try {
    // Axios code
    resolve()
  }catch (error){
    reject()
  }
})

kbottla: updated

kbottla: actually, it is not recommended to use async when returning promise https://eslint.org/docs/latest/rules/no-async-promise-executor

kbottla: I have removed returning Promise but it is not possible to change Axios usage to async/await without changing setupStubs to async

iqbalgds: Thanks for looking into that - I will approve

iqbalgds (approved):

PR #4011, opened 2023-06-15
PP-10520 display cancelled date and cancelled by user email

SandorArpa:
## WHAT
- Display Date cancelled and Cancelled by fields if available when agreement is cancelled


james-peacock-gds (approved):
LGTM

PR #4015, opened 2023-06-19
PP-10598 refactor agreement pact test

SandorArpa:
## WHAT
- refactor cancel agreement pact test to use the new state from connector

with @james-peacock-gds


james-peacock-gds (approved):
I'm confident that this change does not require further review and meets the secure coding requirements of [GOV.UK](http://gov.uk/) Pay

PR #4030, opened 2023-07-06
PP-11244 use /api/ endpoint for getting gateway accounts

stephencdaly:
Use the GET `/v1/api/accounts` and GET `/v1/api/accounts/{accountId}` endpoints rather than the `/v1/frontend/` variants.

The behaviour of the `/v1/api/` and `/v1/frontend/` endpoints is now identical and we intend to remove the `/v1/frontend` variants.


SandorArpa (approved):
LGTM

PR #4038, opened 2023-07-11
PP-10627 add agreement details page cy test

SandorArpa:
## WHAT
- add a new test that checks the rendering of the agreement details page in specifics

### DIFF
--- a/test/cypress/integration/agreements/agreement.cy.js
+++ b/test/cypress/integration/agreements/agreement.cy.js
@@ -42,7 +42,74 @@ describe('Agreement detail page', () => {
     cy.setEncryptedCookies(userExternalId)
   })
 
-  it('should display agreement detail page and allow cancelling agreement', () => {
+  it('should display agreement detail page', () => {
+    cy.task('setupStubs', [
+      ...userAndGatewayAccountStubs(),
+      agreementStubs.getLedgerAgreementSuccess({
+        service_id: serviceExternalId,
+        live: false,
+        gatewayAccountId,
+        external_id: 'a-valid-agreement-id'
+      }),
+      transactionsStub
+    ])
+
+    cy.visit('/test/service/service-id/account/gateway-account-id/agreements/a-valid-agreement-id')
+
+    cy.get('h1').contains('Agreement detail')
+
+    cy.get('.govuk-heading-l').should('have.text', 'Agreement detail')
+
+    cy.get('[data-cy=agreement-detail]').find('dt').eq(0).contains('ID')
+    cy.get('[data-cy=agreement-detail]').find('dd').eq(0).contains('a-valid-agreement-id')
+    cy.get('[data-cy=agreement-detail]').find('dt').eq(1).contains('Reference')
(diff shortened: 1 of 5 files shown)


test/cypress/integration/agreements/agreement.cy.js
```diff
@@ -42,7 +42,49 @@ describe('Agreement detail page', () => {
...
+      transactionsStub
+    ])
+
+    cy.visit('/test/service/service-id/account/gateway-account-id/agreements/a-valid-agreement-id')
+
+    cy.get('h1').contains('Agreement detail')
+
+    cy.get('.govuk-heading-l').should('have.text', 'Agreement detail')
+
+    cy.get('.govuk-summary-list__value').contains('Test User')
```
stephencdaly, line 63: It's not clear which `.govuk-summary-list__value` this is checking, which might make it difficult to debug if it starts failiing.

To be more deliberate about which elements we are checking the contents of, we could add a `data-cy` attribute on the elements in the Nunjucks and use that in the test to get the element we're checking.

See: https://pay-team-manual.cloudapps.digital/manual/how-to/write-cypress-tests.html#select-elements-with-the-data-cy-attribute

Also, are we checking all the attributes on the agreement are displayed as expected? Id, reference, status etc.

SandorArpa: 👍

test/cypress/integration/agreements/agreement.cy.js
```diff
@@ -42,7 +42,75 @@ describe('Agreement detail page', () => {
...
+      transactionsStub
+    ])
+
+    cy.visit('/test/service/service-id/account/gateway-account-id/agreements/a-valid-agreement-id')
+
+    cy.get('h1').contains('Agreement detail')
+
+    cy.get('.govuk-heading-l').should('have.text', 'Agreement detail')
+
+    cy.get('[data-cy=agreement-detail]').get('h2').contains('Payment instrument')
```
stephencdaly, line 63: This assertion is ever so slightly confusing as we're also asserting on the "Payment instrument" title further down when we have assertions for the payment instrument section. It makes it seem like the assertions below this might be looking at the payment instrument section.

Maybe we can remove this one?

test/cypress/integration/agreements/agreements.cy.js
```diff
@@ -57,7 +57,7 @@ describe('Agreements', () => {
     cy.setEncryptedCookies(userExternalId)
   })
 
-  it('should correctly display agreements for a given service', () => {
+  it.only('should correctly display agreements for a given service', () => {
```
stephencdaly, line 60: `.only`

test/cypress/integration/agreements/agreement.cy.js
SandorArpa: 👍

test/cypress/integration/agreements/agreements.cy.js
SandorArpa: 🤦

stephencdaly (approved):
LGTM!

PR #4039, opened 2023-07-14
PP-9858 Add classNames to agreements elements

james-peacock-gds:
Context: End-to-end tests require that certain elements on the agreements pages can be identified
- Add 'agreements-list' and 'agreements-list--items' class names to relevant elements so they can be identified
- Correct misleading ids for agreement list headers


alexbishop1 (approved):
Classy!

iqbalgds (approved):

PR #4052, opened 2023-07-24
PP-11251 Update `Your PSP > Worldpay` with new styling

iqbalgds:
- Update the page to use the new `Summary list card` styling from design system v4.
  - Update the `account credentials` and `3DS flex` sections.
  - No longer have a seperate change link for each entry in the `summary list`. Just one change link in the heading.
  - Remove the `?change=**` query parameter from the change link.  As this is no longer required as we only have 1 change link per section.
    - Update the `3DS flex` credentials nunjucks file to longer use the `change` query       parameter.
- Update the 3DS flex controller - Remove references to the `change` query parameter as  this is no longer used.

image

### DIFF
--- a/app/views/your-psp/_worldpay.njk
+++ b/app/views/your-psp/_worldpay.njk
@@ -2,32 +2,35 @@
 
 <p class="govuk-body">Go to your <a class="govuk-link" href="https://secure.worldpay.com/sso/public/auth/login.html">Worldpay account</a> to get the details you need to enter here, or <a class="govuk-link" href="https://docs.payments.service.gov.uk/switching_to_live/set_up_a_live_worldpay_account/#connect-your-live-account-to-worldpay">read more in our documentation.</a></p>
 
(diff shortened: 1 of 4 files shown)


app/views/your-psp/_worldpay.njk
```diff
@@ -2,32 +2,35 @@
...
+      }
+    },
     rows: [
       {
         key: {
           text: "Merchant code"
         },
         value: {
-          text: credential.credentials.one_off_customer_initiated.merchant_code if isAccountCredentialsConfigured else "Not configured",
+        text: credential.credentials.one_off_customer_initiated.merchant_code if isAccountCredentialsConfigured else "Not configured",
```
stephencdaly, line 32: I think the indentation got messed up slightly by conflict resolution. Can be fixed in next PR.

stephencdaly (approved):
LGTM!

app/views/your-psp/_worldpay.njk
iqbalgds: Thanks - will fix this in the next PR.

PR #4054, opened 2023-07-25
PP-11251 Your PSP > Worldpay creds - add functionality for recurring …

iqbalgds:
- Allow users to enter `customer initiated` and `merchant initiated` credentials for
  gateway account with recurring enabled.
- Uses the design system v4 Summary card styling.

- Cypress test will be added in a future PR.

image

### DIFF
--- a/app/views/your-psp/_worldpay.njk
+++ b/app/views/your-psp/_worldpay.njk
@@ -2,55 +2,169 @@
 
 <p class="govuk-body">Go to your <a class="govuk-link" href="https://secure.worldpay.com/sso/public/auth/login.html">Worldpay account</a> to get the details you need to enter here, or <a class="govuk-link" href="https://docs.payments.service.gov.uk/switching_to_live/set_up_a_live_worldpay_account/#connect-your-live-account-to-worldpay">read more in our documentation.</a></p>
 
-{% set isAccountCredentialsConfigured = (credential.credentials and credential.credentials.one_off_customer_initiated|length) %}
+{% if isRecurringEnabled %}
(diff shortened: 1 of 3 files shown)


app/views/your-psp/_worldpay.njk
```diff
@@ -2,55 +2,169 @@
...
-            attributes: {
-              id: "credentials-change-link"
+      card: {
+        title: {
+          text: "Recurring customer initiated transaction (CIT) credentials"
+        },
+        actions: {
+          items: [
+            {
+              href: '#',
```
alexbishop1, line 21: Should this be a real URL?

alexbishop1 (changes requested):

app/views/your-psp/_worldpay.njk
iqbalgds: Good question - planning to update this later when working on this ticket.
https://payments-platform.atlassian.net/browse/PP-11252 - which adds the functionality to edit credentials.

As I'm not sure yet if we are using the existing `route / path` or `query parameter`.
May need to create a new route.

sfount: I tried out https://github.com/alphagov/pay-selfservice/pull/4058/commits/a5ac7fe9d347f7475d3e1e397d0a7f0902fe406e which has some ideas about how that might work and depends on this

stephencdaly (approved):

stephencdaly:
Links will be added by this upcoming PR https://github.com/alphagov/pay-selfservice/pull/4058

PR #4058, opened 2023-07-27
PP-11252 Worldpay update credentials page for recurring merchant details

sfount:
Depends on https://github.com/alphagov/pay-selfservice/pull/4054

Adds a central place to keep track of the nested merchant details
Worldpay accounts can have to enable different payment channels. This
shifts over to explicitly specifying default credentials are for one off
customer initiated payments.

The mechanism for verifying the merchant details with Worldpay remains
unchanged.

Update the page template to give different context and label guidance
based on which is being updated.

Assert existing behaviour works for:
- updating your PSP Worldpay details
- switching PSP to Worldpay


stephencdaly (dismissed):
LGTM!

sfount:
We'll wait for the additional cypress coverage added after https://github.com/alphagov/pay-selfservice/pull/4054 and then extend that for the update pages in this PR, we can the resolve conflicts and get this merged

PR #4060, opened 2023-07-27
PP-11283 re-enable pact tests

SandorArpa:
## WHAT

In a previous commit we disabled some pact tests due to adminusers being refactored.

- refactor tests for adminusers changes
- re-enable tests


stephencdaly (approved):
LGTM!

PR #4061, opened 2023-07-28
PP-11251 Your PSP > Worldpay creds for test accounts - update design

iqbalgds:
- Update the 3DS section for test Worldpay accounts.
- Move the `toggle 3DS` button higher.
- General clean up of code.

## Test Worldpay accounts - move the toggle button higher
image

### DIFF
--- a/app/views/your-psp/_worldpay-flex.njk
+++ b/app/views/your-psp/_worldpay-flex.njk
@@ -2,23 +2,47 @@
 {% set isTest = currentGatewayAccount.type === "test" %}
 
 {% if isTest %}
-  <h2 class="govuk-heading-m govuk-!-margin-top-9">3DS Flex</h2>
+  <div data-cy="3ds-flex-section-for-test-worldpay-accounts">
+    <h2 class="govuk-heading-m govuk-!-margin-top-9">3DS Flex</h2>
 
-  {{ govukInsetText({
-    "attributes": {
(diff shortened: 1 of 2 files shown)


app/views/your-psp/_worldpay-flex.njk
```diff
@@ -2,23 +2,47 @@
...
+    {{ govukInsetText({
+      "attributes": {
+        "id": "worldpay-3ds-flex-is-on" if isWorldpay3dsFlexEnabled else "worldpay-3ds-flex-is-off"
+      },
+      text: "3DS Flex is turned on." if isWorldpay3dsFlexEnabled else "3DS Flex is turned off."
+    }) }}
+
+    <p class="govuk-body">
+      If you have set up 3DS Flex on your Worldpay account, you need to enter the following details to turn 3DS Flex on.
+      You’ll find these details in your <a class="govuk-link" href="https://secure.worldpay.com/sso/public/auth/login.html">Worldpay account</a>.
```
stephencdaly, line 17: I'm not sure this text makes sense for test accounts, as you can't "set up" flex in Worldpay, I think you just have to use the test 3DS flex credentials from the Worldpay docs. I think this text remains from when you could enable/disable flex for live accounts.

But doesn't look like this PR has changed this text at all.

stephencdaly (approved):
LGTM!

app/views/your-psp/_worldpay-flex.njk
iqbalgds: Thanks for that.  I will merge this in and see if Alan has any thoughts.

PR #4070, opened 2023-08-02
PP-11251 Your PSP > Worldpay test accounts - Toggle 3DS flex content

iqbalgds:
Your PSP > Worldpay Test account
  - When 3DS flex crendentials already exist in a Worldpay test account, hide the content that tells the user to enter 3DS flex crendentials in order to enable 3DS flex.
  - As it is not relevant if 3DS flex settings are already configured.


stephencdaly (approved):
LGTM!

PR #4079, opened 2023-08-07
PP-10823 Check for PAN in reference field - add util methods

iqbalgds:
- Add util methods to check if value is a potential PAN.
- These methods were taken from an older version of `products-ui` which use to check if a reference was a PAN using these methods.
- These methods will be used in a future PR to check if the user is searching for transactions with a potential PAN.

### DIFF
--- a/app/utils/is-a-potential-pan.test.js
+++ b/app/utils/is-a-potential-pan.test.js
@@ -1,0 +1,31 @@
+'use strict'
+
+const isAPotentialPAN = require('./is-a-potential-pan')
+const { expect } = require('chai')
+
+describe('Is A Potential card number', () => {
+  it('should return true if reference is 15 digits passes luhn check', () => {
+    expect(isAPotentialPAN('424242424242424')).to.equal(true)
+  })
+  it('should return true if reference is 16 digits passes luhn check', () => {
+    expect(isAPotentialPAN('4242424242424242')).to.equal(true)
+  })
+  it('should return true if reference has characters [- or space] and passes luhn check', () => {
+    expect(isAPotentialPAN('4242 4242 4242-42-42')).to.equal(true)
+  })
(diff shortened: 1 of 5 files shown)


app/utils/is-a-potential-pan.test.js
```diff
@@ -0,0 +1,28 @@
+'use strict'
+
+const isAPotentialPAN = require('./is-a-potential-pan')
+const { expect } = require('chai')
+
+describe('Is A Potential card number', () => {
+  it('should return true if reference passes luhn check', () => {
+    expect(isAPotentialPAN('4242424242424242')).to.equal(true)
```
stephencdaly, line 8: Should we add another test case for a 15 digit card number that passes the luhn check? Seeing as it's only 15/16 digits we care about it's probably worthwhile.

app/utils/is-a-potential-pan.test.js
```diff
@@ -0,0 +1,28 @@
...
+    expect(isAPotentialPAN('4242 4242 4242 42')).to.equal(false)
+  })
+  it('should return false if reference fails luhn check', () => {
+    expect(isAPotentialPAN('42424242424211')).to.equal(false)
+  })
+  it('should return false if reference has characters [- or space] and but fails luhn check', () => {
+    expect(isAPotentialPAN('4242-4242-4242-11')).to.equal(false)
+  })
+  it('should return false for reference with characters', () => {
+    expect(isAPotentialPAN('REF123456')).to.equal(false)
```
stephencdaly, line 26: Could probably improve this test by having a valid 15 digit card number proceeded by an alphabetic character, that way we can be sure it's returning false due to having an alphabetic character, rather than for other reasons

app/utils/is-a-potential-pan.test.js
iqbalgds: Good spot

PR #4082, opened 2023-08-08
PP-10823 - Check reference for PAN - add JS validation

iqbalgds:
- Adds client side JS validation to check if a reference used during a search might be a PAN.
- Displays an inline error only when the user moves focus away from the field.

image


stephencdaly (approved):
LGTM!

PR #4124, opened 2023-09-13
BAU Log details for AxiosError

stephencdaly:
We use axios to make requests to the zendesk API. Update `error-handler.js` to log available data from errors returned by axios.

### DIFF
--- a/app/middleware/error-handler.js
+++ b/app/middleware/error-handler.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const Sentry = require('@sentry/node')
+const { AxiosError } = require('axios')
 
 const logger = require('../utils/logger')(__filename)
 const {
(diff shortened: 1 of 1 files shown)


app/middleware/error-handler.js
```diff
@@ -84,6 +85,11 @@ module.exports = function errorHandler (err, req, res, next) {
       service: err.service,
       status: err.statusCode
     })
+  } else if (err instanceof AxiosError) {
+    logger.info(`Unhandled AxiosError caught: ${err.message}`, {
```
alexbishop1, line 89: Curious that we log this at info level…

alexbishop1 (approved):

app/middleware/error-handler.js
stephencdaly: I think it's just because we also send the error to sentry using `Sentry.captureException(err)`. If we logged at error level, we would create 2 events in sentry

PR #4125, opened 2023-09-13
BAU Fix creating zendesk ticket

stephencdaly:
When creating a Zendesk ticket, we need to send the name of the requester. We were getting this from the `username` field on a User, but this field has been removed. Zendesk was returning a 422 when the name was empty in some cases.

Instead, get the name from the `email` field, which was the same as the `username` anyway when the username existed.

Remove/replace other places where we were still trying to access the `username` on the User.


AbdirizakIdris (approved):
Looks good to me

jfharden (approved):

PR #4142, opened 2023-10-12
PP-11120 Spike replacing request with axios

stephencdaly:
Add axios-base-client.js to replace base.client.js. Replace usage in connector.client.js as proof of it working.

Added some tests in axios-base-client.test.js, but possibly will need to add a few more when this is properly implemented.

Have implemented custom retry logic to handle ECONNRESET errors. We could use the `axios-retry` library for this instead if we don't trust this.

The intention is that `axios-base-client.js` could be moved to pay-js-commons. When the client is configure, it accepts an options object which can provide hook functions to hook into setting the headers for a request and performing actions on the request starting/ending, which can be used to do logging per app.


app/services/clients/connector.client.js
```diff
@@ -103,12 +103,8 @@ ConnectorClient.prototype = {
    */
   getAccount: function (params) {
     const url = _accountUrlFor(params.gatewayAccountId)
-    return baseClient.get({
-      baseUrl: this.connectorUrl,
-      url,
-      description: 'get an account',
-      service: SERVICE_NAME
-    })
+    return this.client._axios.get(url, { description: 'get an account' })
```
kbottla, line 106: is it possible to make it loo like `this.client.get(url)` for simplicity?

stephencdaly: Good idea, probably makes sense to do that - could also make `description` a parameter to a `get` method too as we'll always want to set it

katstevens:
> The intention is that axios-base-client.js could be moved to pay-js-commons.

The dream 😻 I like the sound of being able to configure different logging settings per app (eg if we are paranoid about not logging card data in Frontend but need to debug everything in Toolbox).

nlsteers:
We'll still need to handle the requirement to route requests through a proxy.

Does Axios pick up proxy settings via the environment automatically?

stephencdaly:
> We'll still need to handle the requirement to route requests through a proxy (for `frontend` at least).
> 
> Does Axios pick up proxy settings via the environment automatically?

Yeh I think so - will have to double check this works as expected but not sure how to test that locally 🤔. We already use Axios for request to the Zendesk API and we don't set any special configuration there. It should just pick up the HTTP_PROXY environment variable and use that.

app/services/clients/connector.client.js
stephencdaly: Have done this https://github.com/alphagov/pay-selfservice/pull/4142/commits/8788f01e16eed3e5af489602f7c209144c7e8711

PR #4146, opened 2023-10-18
PP-9916 Change test card number in Stripe test accounts

JFSGDS:
## WHAT
Change test card number in Stripe test accounts to UK issued test card

https://payments-platform.atlassian.net/browse/PP-9916

## HOW 
Update ‘Mock card numbers’ template to test for ‘stripe’ account and display UK issued test card.

### DIFF
--- a/test/cypress/integration/demo-payment/mock-cards.cy.js
+++ b/test/cypress/integration/demo-payment/mock-cards.cy.js
@@ -1,0 +1,36 @@
+const userStubs = require('../../stubs/user-stubs')
+const gatewayAccountStubs = require('../../stubs/gateway-account-stubs')
+const transactionStubs = require('../../stubs/transaction-stubs')
+
+const userExternalId = 'cd0fa54cf3b7408a80ae2f1b93e7c16e'
+const gatewayAccountId = '42'
+const gatewayAccountExternalId = 'a-valid-external-id'
+
+describe('Show Mock cards screen', () => {
+  beforeEach(() => {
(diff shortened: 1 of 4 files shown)


test/cypress/integration/demo-payment/mock-cards.cy.js
```diff
@@ -0,0 +1,36 @@
...
+      gatewayAccountStubs.getGatewayAccountByExternalIdSuccess({
+        gatewayAccountId,
+        gatewayAccountExternalId,
+        paymentProvider: 'sandbox'
+      }),
+      transactionStubs.getTransactionsSummarySuccess()
+    ])
+  })
+
+  it('should load the mock cards page and show non stripe card', () => {
```
alexbishop1, line 22: Presumably we can have a version of this test that sets `paymentProvider` to `stripe` and checks we output the right card number for Stripe too?

JFSGDS: Agreed. Test to check correct card number is displayed for Stripe accounts now added. Note, unable to land directly on the Dashboard page for a Stripe account  as this route returned a ‘403 Forbidden’ status. Landing on the Stripe account’s Settings page and then navigating to the Dashboard pages worked but this probably needs refactoring.

alexbishop1: I think it’s fine — the amount of extra code to do the navigation is not huge and it’s not worth investing too much more time into this.

PR #4148, opened 2023-10-19
PP-7843 Stop selecting default option on permission level radios

hjvoid:
## WHAT

- Removed the `checked: true` property from the role-input checkboxes.
- Stopped the error being passed to the try/catch block and implemented a more explicit flash error message for the user.
- Added a test to check that the correct error message is displayed if the role-id is not recognised.
- Changed the property on the invite-user ui test to check that none of the role inputs should be checked when the page is loaded.

## HOW 

- When a user invites a new team member, no user role should be selected. 
- If the user does not check on of the checkboxes before they submit the form they are presented with an error which reads 'Select the team member's permission level'.


app/controllers/invite-user.controller.js
```diff
@@ -60,7 +60,9 @@ async function invite (req, res, next) {
     lodash.set(req, 'session.pageData', { invitee })
     res.redirect(303, formatServicePathsFor(paths.service.teamMembers.invite, externalServiceId))
   } else if (!role) {
-    next(new Error(`Cannot identify role from user input ${roleId}`))
+    req.flash('genericError', 'Select the team member\'s permission level')
```
alexbishop1, line 63: If you use a right single quotation mark (that is, `’` instead of `'`) then it looks nicer and does not need to be escaped.

package-lock.json
```diff
@@ -64,7 +64,7 @@
         "chai-arrays": "2.2.0",
         "chai-as-promised": "7.1.1",
         "cheerio": "1.0.0-rc.12",
-        "chokidar-cli": "*",
```
alexbishop1, line 67: I think this should be excluded unless we’re actually trying to update dependencies.

alexbishop1 (changes requested):
Looks good. Just one thing about an apostrophe.

package-lock.json
nlsteers: any reason why we can't update chokidar? I see this change every time I `npm i` so it'd be nice to merge this.

app/controllers/invite-user.controller.js
hjvoid: Gotcha! Have updated this as above.

package-lock.json
hjvoid: I've reverted this back to `latest` for the time being as I guess it's not part of the story?

alexbishop1 (changes requested):
One final thing with the `package-lock.json` file and this is good to go.

package.json
```diff
@@ -128,7 +128,7 @@
     "chai-arrays": "2.2.0",
     "chai-as-promised": "7.1.1",
     "cheerio": "1.0.0-rc.12",
-    "chokidar-cli": "latest",
```
alexbishop1, line 131: This should be `*` (what it is on master), not `latest`.

alexbishop1 (changes requested):

package.json
hjvoid: Doh...Missed that!

alexbishop1 (approved):

PR #4157, opened 2023-11-24
PP-11591 Set 3ds Flex by default for Worldpay accounts

james-peacock-gds:
Currently 3ds integration version is set to 2 for live Worldpay accounts
- Set 3ds integration version to 2 for all Worldpay accounts, including test, when flex credentials are entered
- Remove the 3ds flex toggle buttons as 3ds flex is on by default when credentials are entered


oswaldquek (approved):
lovely

PR #4158, opened 2023-11-27
PP-11701 Remove wallets feature flag

stephencdaly:
Remove feature flag for showing the setting to enable digital wallets for sandbox accounts, and always show the setting.

Fix some Cypress test that started breaking because the wallet setting is now showing for sandbox accounts, and the tests were accessing the settings by element index. Use the `data-cy` attribute to select the relevant section in the settings before using element indices.


james-peacock-gds (approved):
LGTM!

PR #4173, opened 2024-02-09
PP-12167 Update joi to resolve hoek vulnerability

james-peacock-gds:
Context: Need to update joi version to one which does not use hoek versions before 8.5.1 because previous versions of hoek have a [vulnerability](https://github.com/alphagov/pay-selfservice/security/dependabot/71)

- update to current joi version 17.12.1
- update uses of joi as Joi.validate() function is no longer supported


stephencdaly (approved):
LGTM!

PR #4178, opened 2024-02-19
PP-10608: Correcting the number of days in transaction queries.

nataliecarey:
## WHAT
As described in PP-10608 the number of days searched on the dashboard is currently out by one.  It was intentionally written (with comments describing it) but having looked deeper into it because of the use of midnight as the start of the period the extra day wasn't needed

## HOW 
On an account with transactions every day visit the dashboard and select "Previous 7 days" or "Previous 30 days", the query was previously showing a date range one day longer than requested for each of these categories.

### DIFF
--- a/app/controllers/all-service-transactions/get.controller.js
+++ b/app/controllers/all-service-transactions/get.controller.js
@@ -40,7 +40,7 @@ module.exports = async function getTransactionsForAllServices (req, res, next) {
     const searchResultOutput = await transactionService.search(userPermittedAccountsSummary.gatewayAccountIds, filters.result)
     const cardTypes = await client.getAllCardTypes()
     const downloadRoute = filterLiveAccounts ? paths.allServiceTransactions.download : paths.formattedPathFor(paths.allServiceTransactions.downloadStatusFilter, 'test')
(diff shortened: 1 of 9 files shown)


nlsteers:
Code change looks fine, there's a small accordion on the dashboard service view that explains how the values are calculated. Is this still correct or does it need updating in line with the changes in this PR? 

See example image: 
Screenshot 2024-02-22 at 16 44 11

nataliecarey:
> Code change looks fine, there's a small accordion on the dashboard service view that explains how the values are calculated. Is this still correct or does it need updating in line with the changes in this PR?
> 
> See example image: Screenshot 2024-02-22 at 16 44 11

Good spot.  I'll check, we should be updating it and I expect it's updated as a side-effect of my change but I'll confirm.

nataliecarey:
> Code change looks fine, there's a small accordion on the dashboard service view that explains how the values are calculated. Is this still correct or does it need updating in line with the changes in this PR?
> 
> See example image: Screenshot 2024-02-22 at 16 44 11

I've tested, it shows correctly in the accordion 🎉

nlsteers (approved):
🚀

PR #4181, opened 2024-02-26
PP-11681 Replace ‘base-client’ with ‘axios-base-client’

JFSGDS:
https://payments-platform.atlassian.net/browse/PP-11681

## WHAT
- Change  ‘products.client’ methods to use ‘axios-base-client’ library.
- Refactor call parameters initialisation.
- Use ‘async/await’ syntax.

## HOW 
All automated tests should pass without the need to be amended.
Existing functionality should remain unaffected.


iqbalgds (approved):
Looks good to me.

PR #4182, opened 2024-02-27
PP-12295 Add new Stripe terms and conditions page

iqbalgds:
- This page will contain the latest Stripe terms and conditons.
- This page will be on a brand new URL: /policy/stripe-terms-and-conditions
- Intend to put this live but hidden from real users.
  - This will allow the team to review the content and request updates.
  - We expect there to be some late edits as we are working with the legal team to finalise the content.
- This page is implemented differently from the other /policy/ pages:
  - The other pages generate a 1 hour link to a policy PDF file and follow a specific page layout.
  - This new page will look different and we need to release it quickly.
  - Making this work with the existing code would have required too much effort.
  - Therefore, this was implemented differently as a static page - which is quicker to implement without affecting the existing policy pages.

### DIFF
--- a/app/controllers/stripeTermsAndConditions.controller.js
+++ b/app/controllers/stripeTermsAndConditions.controller.js
@@ -1,0 +1,7 @@
+'use strict'
+
+const { response } = require('../utils/response')
+
+module.exports.get = function (req, res) {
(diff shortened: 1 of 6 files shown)


JFSGDS (approved):
Looks OK to me.

PR #4190, opened 2024-03-06
PP-12237 Fix permissive regex

stephencdaly:
Code QL flagged that the `A-z` in the regex was a "Suspicious character range that is equivalent to [A-Z\[\\]^_`a-z]."


alexbishop1 (approved):
Learnt something new today.

PR #4194, opened 2024-03-13
PP-11681 Replace 'base-client' with 'axios-base-client' in 'adminusers.client'

JFSGDS:
https://payments-platform.atlassian.net/browse/PP-11681

## WHAT
- Change  ‘adminusers.client’ methods to use ‘axios-base-client’.
- Refactor call parameters initialisation.
- Use ‘async/await’ syntax.
- Extend ‘instanceof RESTClientError’ test in registration controller to include a test for constructor name. This is in order to identify an error thrown by the new 'axios-base-client'. Possibly needs refactoring.

## HOW 
All automated tests should pass without the need to be amended.
Existing functionality should remain unaffected.


app/controllers/registration/registration.controller.js
```diff
@@ -53,7 +53,7 @@ async function submitEmailPage (req, res, next) {
 
     res.redirect(paths.register.checkEmail)
   } catch (err) {
-    if (err instanceof RESTClientError) {
+    if (err instanceof RESTClientError | err.constructor.name === 'RESTClientError') {
```
iqbalgds, line 56: Interesting - I am wondering why this is needed?
We can chat about this.

iqbalgds (changes requested):
Looks great - just the test that needs fixing.

app/services/clients/adminusers.client.js
```diff
@@ -32,17 +33,12 @@ module.exports = function (clientOptions = {}) {
...
-        baseUrl,
-        url: `${userResource}/${externalId}`,
-        json: true,
-        description: 'find a user',
-        service: SERVICE_NAME,
-        transform: responseBodyToUserTransformer
-      }
-    )
+  async function getUserByExternalId (externalId) {
+    const url = `${baseUrl}${userResource}/${externalId}`
```
nataliecarey, line 37: Ideally this should use `encodeURIComponent` or a URL builder to avoid injection vulnerabilities or encoding issues if special chars are included.

app/services/clients/adminusers.client.js
```diff
@@ -51,101 +47,69 @@ module.exports = function (clientOptions = {}) {
...
-        },
-        json: true,
-        description: 'find a user',
-        service: SERVICE_NAME,
-        transform: responseBodyToUserListTransformer
-      }
-    )
+  async function getUsersByExternalIds (externalIds = []) {
+    const url = `${baseUrl}${userResource}?ids=${externalIds.join()}`
+    this.client = new Client(SERVICE_NAME)
```
nataliecarey, line 52: Using `this` in this context seems to be unhelpful.  I'd suggest using `const client =` instead.

app/services/clients/adminusers.client.js
```diff
@@ -276,116 +199,84 @@ module.exports = function (clientOptions = {}) {
...
-      {
-        baseUrl,
-        url: `/v1/api/invites/${inviteCode}/reprovision-otp`,
-        json: true,
-        description: 're-provision OTP key',
-        service: SERVICE_NAME
-      }
-    )
+  async function getValidatedInvite (inviteCode) {
+    const url = `${baseUrl}/v1/api/invites/${inviteCode}`
```
nataliecarey, line 233: I'm assuming that `inviteCode` is from user input (or something the user can manipulate) so it makes my previous comment about encoding more important in this instance.

app/services/clients/adminusers.client.js
nataliecarey: Having got further through the code I'm wondering whether you'd prefer to use a shared client.  I'm not sure from just reading the code whether that's a suitable solution or not but it seems like it might be a good match as the client seems to need configuring before use and the configuration seems to be generic.

app/services/clients/adminusers.client.js
```diff
@@ -2,7 +2,8 @@
 
 const lodash = require('lodash')
 
-const baseClient = require('./base-client/base.client')
+const { Client } = require('@govuk-pay/pay-js-commons/lib/utils/axios-base-client/axios-base-client')
```
nataliecarey, line 5: I know it's not related to this PR but I'd suggest avoiding being so specific about the path inside the other project.  I'd personally go with:

```
const { AxiosClient } = require('@govuk-pay/pay-js-commons')
```

Or if you want it kept separate there are ways to have a shorter path e.g.

```
const { Client } = require('@govuk-pay/pay-js-commons/axios')
```

To be really clear:  This would need to be changed in `pay-js-commons` and it's not really part of this PR but it's related to this wider piece of work.

app/services/clients/adminusers.client.js
JFSGDS: Agreed, this approach seems better. Code changed as suggested.

app/services/clients/adminusers.client.js
JFSGDS: Good suggestion. I agree it would be much neater.
I also agree that it's not really part of this PR so I propose creating a new ticket to cover this. It should be applied to other repositories already amended for the new version of `pay-js-commons` containing the new `axios-base-client` namely `pay-products-ui` and `pay-frontend`.

app/services/clients/adminusers.client.js
JFSGDS: Good point. Presumably this should be done for for all constructs of url with query string parameters (or at least those though to be provided by user input)?
@iqbalgds  and I thought this may be better done as a separate PR.

app/services/clients/adminusers.client.js
JFSGDS: Good point. Please see earlier comment.

app/services/clients/adminusers.client.js
nataliecarey: So I tend to do it with everything that's not hard-coded but it's particularly important for user input.  In some projects I end up creating a URL model that would be something like:

```
const url = buildUrl({
  base: baseUrl,
  uriParts: [userResource, externalId],
  query: {
    theKey: 'theValue'
  }
})
```

For this ticket I think it's reasonable to just use `encodeURIComponent` on each piece of user input in URLs, but personally if it were me doing it I'd rather use it for all URL parts that aren't hard-coded as  I'd find that to be less effort than figuring out which ones are or might be from user input.

app/services/clients/adminusers.client.js
iqbalgds: I have created a new ticket for this:
https://payments-platform.atlassian.net/browse/PP-12468

nataliecarey (approved):

app/controllers/registration/registration.controller.js
iqbalgds: This has been updated - resolving.

iqbalgds (approved):

PR #4197, opened 2024-03-25
PP-11681 Encode URL when making axios call

JFSGDS:
## WHAT
When configuring `axios-base-client` call, encode the url passed in.
## HOW

- Amend `configureClient` function to encode url passed in from client code issuing axios call.
- Amend unit test to cover change.


iqbalgds (approved):
Looks good to me

PR #4203, opened 2024-04-03
PP-12100 Allow users to apply filters to their transaction search if it times out

james-peacock-gds:
- Add new error for a ledger timeout on all services transactions search
- Link from the error page to the all services transaction search page,  without the automatic search, to avoid a perpetual timeout
- Refactor to share the code that populates the model
- Update unit tests
- Update cypress tests

with @DomBelcher and @nataliecarey

### DIFF
--- a/app/middleware/error-handler.js
+++ b/app/middleware/error-handler.js
@@ -16,7 +16,8 @@ const {
   InvalidConfigurationError,
   ExpiredInviteError,
   RESTClientError,
-  GatewayTimeoutError, GatewayTimeoutForAllServicesSearchError
+  GatewayTimeoutError,
+  GatewayTimeoutForAllServicesSearchError
 } = require('../errors')
 const paths = require('../paths')
 const { renderErrorView, response } = require('../utils/response')
--- a/app/middleware/error-handler.js
+++ b/app/middleware/error-handler.js
@@ -16,7 +16,7 @@ const {
   InvalidConfigurationError,
   ExpiredInviteError,
   RESTClientError,
-  GatewayTimeoutError
+  GatewayTimeoutError, GatewayTimeoutForAllServicesSearchError
 } = require('../errors')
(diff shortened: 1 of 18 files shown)


app/services/transaction.service.js
```diff
@@ -21,12 +21,16 @@ const connectorRefundFailureReasons = {
...
 }
 
-const searchLedger = async function searchLedger (gatewayAccountIds = [], filters) {
+const searchLedger = async function searchLedger (gatewayAccountIds = [], filters, allServices = false) {
   try {
     return await Ledger.transactions(gatewayAccountIds, filters)
   } catch (error) {
     if (error.errorCode === 504) {
-      throw new GatewayTimeoutError('Your request has timed out. Please apply more filters and try again.')
+      if (allServices === true) {
```
nlsteers, line 29: could be simplified to `if (allServices) {`

nlsteers:
can you add a few screenshots to the PR demonstrating the new behaviour?

DomBelcher:
This is the error shown when a search times out:
timeout-error png

And this is the view it links to when navigating back to the search page, with no search results
nosearch png (1)

app/services/transaction.service.js
DomBelcher: 👍

app/middleware/error-handler.js
```diff
@@ -16,7 +16,7 @@ const {
   InvalidConfigurationError,
   ExpiredInviteError,
   RESTClientError,
-  GatewayTimeoutError
+  GatewayTimeoutError, GatewayTimeoutForAllServicesSearchError
```
nataliecarey, line 19: Optional: This could go on a new line to be consistent.

james-peacock-gds: done!

nataliecarey (approved):

PR #4212, opened 2024-04-19
PP-11820 Upgrade Selfservice to Cypress 13

marcotranchino:
With this change, we are upgrading pay-selfservice to the new version of Cypress: 13.8.0

We are also removing the package.json override for cypress/request, which is no longer needed.

Further information in Jira.

https://payments-platform.atlassian.net/browse/PP-11820

## Feedback request

Is there any unnecessary config/deprecated code still to be removed?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1890,9 +1890,9 @@
       }
     },
     "node_modules/@cypress/request": {
-      "version": "2.88.11",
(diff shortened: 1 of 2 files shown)


iqbalgds (approved):
Looks good to me, well done

PR #4217, opened 2024-05-07
PP-12593 Explain exclusion of Maestro in sandbox accounts

marcotranchino:
With this change, we are explaining in the UI that Maestro is not available in
**Sandbox accounts**.

**In test accounts** and in **live accounts**, we are showing copy to explain that 
3DS needs to be enabled, if it isn't and if it is required.

This is needed because the current hint displayed on production for Maestro,
would link users to a page that is no longer available.

To summarise the scenarios:

- IF (Account is LIVE or TEST) AND (Card requires 3D Secure) AND  (3D Secure is not enabled)  

  - THEN The card option is disabled and the hint states: “{card type} cannot be used because 3D Secure is not available. Please contact support.“ 

- IF (Account is SANDBOX)  AND (Card requires 3D Secure) AND  (3D Secure is not enabled) 

  - THEN The card option is disabled and there is a hint stating that “{card type} is not available on sandbox test accounts”.

- OTHERWISE The card option is enabled and there is no hint.

Note that with the first commit, we are also removing a broken link from the README page.

The link was pointing to test/cypress/cypress_testing.md, containing obsolete
Cypress testing information deleted some time ago[1].

Further information in JIRA[2].

[1]
https://github.com/alphagov/pay-selfservice/commit/6d8a142649f8302e253313d5c1bb77b9e4f1941c

[2]
https://payments-platform.atlassian.net/browse/PP-12593

## SCENARIOS

SCENARIOS VERSION 3

## BEFORE

BEFORE

## AFTER

AFTER - LIVE or TEST

AFTER - ENABLED

AFTER SANDBOX

Note: the copy needs to be amended with Alan:

- **“{card type} cannot be used because 3D Secure is not available. Please contact support.“**
- **“{card type} is not available on sandbox test accounts”**

### DIFF
--- a/test/cypress/integration/settings/payment-types.cy.js
+++ b/test/cypress/integration/settings/payment-types.cy.js
@@ -56,7 +56,7 @@ describe('Payment types', () => {
   context('when the account is test', () => {
     context('when 3DS is not enabled', () => {
       beforeEach(() => {
-        setupStubs(userExternalId, gatewayAccountId, gatewayAccountExternalId, serviceName)
+        setupStubs(userExternalId, gatewayAccountId, gatewayAccountExternalId, serviceName, 'test')
         cy.visit(`/account/${gatewayAccountExternalId}/payment-types`)
       })
 
--- a/app/controllers/payment-types/get-index.controller.js
+++ b/app/controllers/payment-types/get-index.controller.js
@@ -28,7 +28,8 @@ function formatCardsForTemplate (allCards, acceptedCards, account) {
 
       if (card.requires3ds && !threeDSEnabled) {
         formatted.disabled = true
-        formatted.hint = { html: account.type === 'sandbox' ? `${card.label} is not available on sandbox test accounts` : `${card.label} cannot be used because 3D Secure is not available. Please contact support`
+        formatted.hint = {
+          html: account.type === 'sandbox' ? `${card.label} is not available on sandbox test accounts` : `${card.label} cannot be used because 3D Secure is not available. Please contact support`
         }
       }
       return formatted


alexbishop1 (changes requested):
Looks good overall but the logic for when to show the label is not quite right.

test/cypress/integration/settings/payment-types.cy.js
```diff
@@ -53,6 +53,104 @@ describe('Payment types', () => {
       .should('contain', 'Accepted card types have been updated')
   })
 
+  context('when the account is test', () => {
```
alexbishop1, line 56: I suppose this is test but not sandbox, really?

test/cypress/integration/settings/payment-types.cy.js
```diff
@@ -53,6 +53,104 @@ describe('Payment types', () => {
...
+    context('when 3DS is not enabled', () => {
+      beforeEach(() => {
+        setupStubs(userExternalId, gatewayAccountId, gatewayAccountExternalId, serviceName)
```
alexbishop1, line 59: Might be good to explicitly specify `test` here

test/cypress/integration/settings/payment-types.cy.js
```diff
@@ -53,6 +53,104 @@ describe('Payment types', () => {
...
+        cy.visit(`/account/${gatewayAccountExternalId}/payment-types`)
+      })
+
+      it('should show debit cards that do not require 3DS as enabled and without hint', () => {
+        cy.get('#debit-2').should('be.checked')
+        cy.get('#debit-2-item-hint').should('not.exist')
+      })
+
+      it('should show debit cards requiring 3DS as disabled and with hint', () => {
+        cy.get('#debit-3').should('be.disabled')
```
alexbishop1, line 69: `#debit-3` might be a bit brittle but I cannot immediately think of a better way of doing this

alexbishop1 (changes requested):
Just a couple of stylistic things.

alexbishop1 (approved):

PR #4220, opened 2024-05-17
PP-12632 - Improve the agreement page by sanitising the description field

marcotranchino:
With this change, we are stripping tags from the description field in the Agreement page.

This is needed because the content is otherwise displayed without being sanitised first.

The first commit adds a failing test to reproduce the issue.

The second commit fixes the problem.

Further information in Jira.

https://payments-platform.atlassian.net/browse/PP-12632

## FAILING TEST AFTER THE FIRST COMMIT

FAILING TEST

## Request for feedback

Among any other considerations, could you please also let me know if we could/should sanitise the description
in the same way as the reference is sanitised, which is with `text` instead of `html`.
See alternative fix (which also passes the newly created failing test) below.

### current fix

CURRENT FIX

### alternative fix

ALTERNATIVE FIX

### DIFF
--- a/app/views/agreements/detail.njk
+++ b/app/views/agreements/detail.njk
@@ -62,7 +62,7 @@
       value: { html: agreementStatusTag(agreement.status) }
     },{
       key: { text: "Description" },
-      value: { html: agreement.description },
+      value: { text: agreement.description },
       attributes: {
         id: "data-description"
       }


app/views/agreements/detail.njk
```diff
@@ -62,7 +62,7 @@
       value: { html: agreementStatusTag(agreement.status) }
     },{
       key: { text: "Description" },
-      value: { html: agreement.description },
+      value: { html: agreement.description | striptags(true) | escape | nl2br },
```
alexbishop1, line 65: I think the `text` approach is better. I can’t really think why this was `html` in the first place. Possibly it was just copied from the one above.

alexbishop1 (changes requested):
Please rename the branch and this PR so it has “PP-12632”.

alexbishop1:
Can you rename the branch as well? You might have to open a new PR.

PR #4221, opened 2024-05-17
PP-12632 - Improve the agreement page by sanitising the description field

marcotranchino:
With this change, we are stripping tags from the description field in the Agreement page.

This is needed because the content is otherwise displayed without being sanitised first.

The first commit adds a failing test to reproduce the issue.

The second commit fixes the problem.

Further information in Jira.

https://payments-platform.atlassian.net/browse/PP-12632

Note: review for this PR was provided in a previous PR: https://github.com/alphagov/pay-selfservice/pull/4220

## FAILING TEST AFTER THE FIRST COMMIT

FAILING TEST

### DIFF
--- a/app/views/agreements/detail.njk
+++ b/app/views/agreements/detail.njk
@@ -62,7 +62,7 @@
       value: { html: agreementStatusTag(agreement.status) }
     },{
       key: { text: "Description" },
-      value: { html: agreement.description },
+      value: { text: agreement.description },
       attributes: {
         id: "data-description"
       }


test/cypress/integration/agreements/agreement.cy.js
```diff
@@ -68,6 +68,7 @@ describe('Agreement detail page', () => {
     cy.get('[data-cy=agreement-detail]').find('dd').eq(2).contains('active')
     cy.get('[data-cy=agreement-detail]').find('dt').eq(3).contains('Description')
     cy.get('[data-cy=agreement-detail]').find('dd').eq(3).contains('Reason shown to paying user for taking agreement')
+    cy.get('[data-cy=agreement-detail]').find('dd').eq(3).contains("console.log('AGREEMENT')")
```
alexbishop1, line 71: Might be better if this contains the full `Reason shown to paying user for taking agreement <script>console.log('AGREEMENT')</script>` word so we know we’re sanitising the HTML tags rather than removing them.

alexbishop1 (changes requested):
One final thing…

alexbishop1 (approved):

alexbishop1 (approved):

PR #4223, opened 2024-05-29
PP-12620 degateway feature flag

nlsteers:
### WHAT

- add new degateway preference setting to user profile options
- add new convenience method to User class for determining account 'degatewayification'
- add new patch method to adminusers client for updating user features
- repurpose user 'features' as possible database flag
- add degateway banner include
- add feature flag to environment:  DEGATEWAY_FLAG


app/models/User.class.js
```diff
@@ -133,6 +134,10 @@ class User {
         serviceRole.service && serviceRole.service.externalId === serviceExternalId)
       .length > 0
   }
+
+  isDegatewayed () {
+    return (this.features.includes('degatewayaccountification') && DEGATEWAY_FLAG)
```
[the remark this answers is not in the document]
nlsteers: done :D

app/views/includes/degateway-banner.njk
```diff
@@ -1,10 +1,11 @@
...
-  <p class="govuk-phase-banner__content">
-    <strong class="govuk-tag govuk-phase-banner__content__tag govuk-tag--orange">
-      Experimental
-    </strong>
-    <span class="govuk-phase-banner__text">
-      Your account is opted in to the experimental degateway view – <a class="govuk-link" href="{{ routes.user.profile.degateway }}">opt out</a>.
-    </span>
-  </p>
-</div>
+{% if isDegatewayed %}
```
nlsteers, line 1: it's clearer from a developers perspective if the logic block that conditionally renders this include is in the template it's imported in rather than the the include itself

nlsteers:
these are all sensible changes, however I'd reconsider hiding conditional render blocks inside of includes

app/views/includes/degateway-banner.njk
```diff
@@ -1,10 +1,11 @@
...
+  {% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
+
+  {{ govukPhaseBanner({
+    tag: {
+      text: "Experimental",
+      classes: "govuk-tag--orange"
+    },
+    html: 'You are seeing an experimental version of this page, it is only intended for the GOV.UK Pay Team.  If you think you should not be seeing this please <a class="govuk-link" href="https://www.payments.service.gov.uk/support/">contact the GOV.UK Pay Team</a>.'
```
nlsteers, line 9: I'd say something like:

```html
You are seeing an experimental version of this page, some features may not work as expected. Please <a class="govuk-link" href="https://www.payments.service.gov.uk/support/">contact support</a> if you did not opt in to this experiment.
```

app/views/includes/degateway-banner.njk
```diff
@@ -0,0 +1,9 @@
+{% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
+{{ govukPhaseBanner({
+  attributes: {id: "degateway-phase-banner"},
+  tag: {
+    text: "Experimental",
+    classes: "govuk-tag--orange"
+  },
+  html: 'This is a test version of this page, intended only for use by the GOV.UK Pay team in GDS. You currently have \'Degatewayaccountification\' enabled in \'My profile\'.<br />If you think you should not be seeing this, please <a class="govuk-link" href="https://www.payments.service.gov.uk/support/">contact the GOV.UK Pay Team</a>.'
```
nlsteers, line 8: this banner will show when the user is 'degatewayed' even if the ability to see the setting in 'My Profile' is disabled. 

this could end up being confusing

app/controllers/user/degateway/post.controller.js
```diff
@@ -0,0 +1,28 @@
...
+  const degatewayPreference = req.body['degateway-preference']
+
+  try {
+    if (!SHOW_DEGATEWAY_SETTINGS) {
+      return res.redirect(paths.user.profile.index)
+    }
+    const { features } = await userService.findByExternalId(req.user.externalId)
+    // todo: should probably fix adminusers to take empty string for features rather than always having a value for subsequent patch requests
+    let featureSet = new Set(['default'])
+    features.forEach(feature => {
```
DomBelcher, line 17: Is `features` not going to be returned as a string, and then needs to be split into an array?

app/models/User.class.js
```diff
@@ -3,6 +3,7 @@
 const _ = require('lodash')
 const ServiceRole = require('./ServiceRole.class')
 const { isInternalGDSEmail } = require('../utils/email-tools')
+const DEGATEWAY_FLAG = process.env.DEGATEWAY_FLAG === 'true'
```
DomBelcher, line 6: I think this could be better named - I assume this is the global enable/disable option? I think something like `ENABLE_DEGATEWAY_VIEW` would convey the purpose a bit better

app/models/User.class.js
```diff
@@ -133,6 +134,10 @@ class User {
         serviceRole.service && serviceRole.service.externalId === serviceExternalId)
       .length > 0
   }
+
+  isDegatewayed () {
+    return DEGATEWAY_FLAG && this.features.includes('degatewayaccountification')
```
DomBelcher, line 139: There's a `hasFeature` method which does the same as `this.features.includes(...)`, only minor but could make use of it

app/views/services/index.njk
```diff
@@ -5,7 +5,11 @@
   Choose service - GOV.UK Pay
 {% endblock %}
 
-{% block beforeContent %}{% endblock %}
+{% block beforeContent %}
+  {% if isDegatewayed %}
+    {% include "includes/degateway-banner.njk" %}
+  {% endif %}
+{% endblock %}
```
DomBelcher, line 12: If `index.njk` extends `layout.njk`, will the degateway banner not already be included because it's in `layout.njk`?

app/models/User.class.js
nlsteers: it's just a temporary convenience method that saves someone passing `degatewayaccountification` as a string and checking that the env var is set.

app/controllers/user/degateway/post.controller.js
nlsteers: see the [User constructor](https://github.com/alphagov/pay-selfservice/blob/bb00443e838cd218e32236d0db8d12465bfd7dc6/app/models/User.class.js#L47)

app/views/includes/degateway-banner.njk
nataliecarey: That's an intentional decision - it allows us to each create an account opted in to this functionality, then turn off the option so that test more closely resembles prod for all other users.

nataliecarey (approved):

nataliecarey:
FYI - I just checked (with @katstevens's help) and the flags are set up correctly on Test, Staging and Production.

james-peacock-gds (approved):
I've checked that we now have an alert in place for the unlikely event that the degateway feature is erroneously deployed into production.

I've reviewed the code which has been added by several people including myself and I'm satisfied that this doesn't require a further review.

PR #4226, opened 2024-06-13
PP-12659 Bump pay-js-commons

iqbalgds:
- Pay-js-commons has been updated with the latest changes of Axios base client.
- The new changes do not affect this repository directly, but want update the version so that it is aligned with the other reporistories.


hjvoid (approved):
Looks good!

PR #4232, opened 2024-06-24
PP-12624 Fix bug in timeout error page

james-peacock-gds:
Context: When there is a database timeout from a transactions search across all services, the resulting error page sometimes has a malformed url, because it relies on a parameter stored in the session, which is not present when a user of live services clicks on the 'View transactions for all my services' link on the My Services page.

- update the formation of the url in the link to the all services transactions 'no-search' page so that if the `live|test` parameter is not stored in the session, `live` is added to the url.
- update Cypress tests to cover all permutations of the all-services-transactions url.


DomBelcher (approved):
Looks good 👍

PR #4237, opened 2024-07-10
PP-12735 Change RESTClientError import to pay-js-commons class

hjvoid:
## WHAT
Updating the post-configure.controller to use the RESTClientError from pay-js-commons/axios-client so that correct error message is shown to users as opposed to a 500 internal server error message.

## HOW 
_Steps to test or reproduce:_

1. Sign in to an existing user account in the admin tool
2. Follow the steps to set up a new authenticator app until you reach the page where you have to enter the code from the app.
3. Deliberately enter an incorrect code.
4. You should now see the following error message:
> The security code you entered is not correct, try entering it again or wait for your authenticator app to give you a new code
Screenshot 2024-07-11 at 17 02 15
Screenshot 2024-07-11 at 17 02 26

_Note: There are other instances of the class that was causing this bug which are being addressed in another ticket_ [here](https://payments-platform.atlassian.net/browse/PP-12868).


app/controllers/user/two-factor-auth/post-configure.controller.js
```diff
@@ -5,7 +5,7 @@ const lodash = require('lodash')
 const paths = require('../../../paths')
 const userService = require('../../../services/user.service.js')
 const secondFactorMethod = require('../../../models/second-factor-method')
-const { RESTClientError } = require('../../../errors')
+const { RESTClientError } = require('@govuk-pay/pay-js-commons/lib/utils/axios-base-client/errors.js')
```
iqbalgds, line 8: Can we please get rid of the .js file extension - thanks

app/controllers/user/two-factor-auth/post-configure.controller.test.js
```diff
@@ -5,7 +5,7 @@ const { expect } = require('chai')
 const userFixtures = require('../../../../test/fixtures/user.fixtures')
 const User = require('../../../models/User.class')
 const paths = require('../../../paths')
-const { RESTClientError } = require('../../../errors')
+const { RESTClientError } = require('@govuk-pay/pay-js-commons/lib/utils/axios-base-client/errors.js')
```
iqbalgds, line 8: Same again - thanks!

iqbalgds (changes requested):
Looks great - tiny teeny change requested

app/controllers/user/two-factor-auth/post-configure.controller.js
hjvoid: Done.

alexbishop1 (approved):

iqbalgds (approved):

PR #4243, opened 2024-07-26
PP-13022: Updated the content on the webhooks page to match the implementation.

nataliecarey:
## WHAT
There was a mismatch between the time we display webhooks for and the time we claim to hold them for.  This corrects that mismatch.


richblake:
Can approve from the angle that this change makes the admin tool reflect what actually happens. LGTM.  Could a developer review the actual code change.

oswaldquek (approved):

hjvoid (approved):
Cool 😎

PR #4246, opened 2024-08-13
PP-12413: Hardcode connector urls

oswaldquek:
It makes for a very unpleasant developer experience to be unable to easily see what connector urls are being called. Rectify this by hardcoding URLs.

### DIFF
--- a/app/services/clients/connector.client.js
+++ b/app/services/clients/connector.client.js
@@ -8,85 +8,11 @@ const StripeAccount = require('../../models/StripeAccount.class')
 
 // Constants
 const SERVICE_NAME = 'connector'
-const ACCOUNTS_API_PATH = '/v1/api/accounts'
-const ACCOUNT_API_PATH = ACCOUNTS_API_PATH + '/{accountId}'
-const CHARGES_API_PATH = ACCOUNT_API_PATH + '/charges'
-const CHARGE_API_PATH = CHARGES_API_PATH + '/{chargeId}'
-const CHARGE_REFUNDS_API_PATH = CHARGE_API_PATH + '/refunds'
-const CARD_TYPES_API_PATH = '/v1/api/card-types'
-const STRIPE_ACCOUNT_SETUP_PATH = ACCOUNT_API_PATH + '/stripe-setup'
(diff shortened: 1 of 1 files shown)


app/services/clients/connector.client.js
```diff
@@ -97,6 +23,15 @@ function ConnectorClient (connectorUrl) {
 }
 
 ConnectorClient.prototype = {
+  getAccountByServiceIdAndAccountType: async function (params) {
+    const url = `${this.connectorUrl}/v1/api/service/{serviceId}/account/{accountType}`
+      .replace('{serviceId}', params.serviceId)
+      .replace('{accountId}', params.accountType)
```
nataliecarey, line 29: The replaces here should be wrapped in `encodeURIComponent` to escape special charactors.

nataliecarey:
I think this is a good change, it brings the replaceable value names and their replacement close to eachother which enhances readability.

I'm concerned that we don't use URL Encoding as a default everywhere and this seems like a good moment to add it here, if you think there's a reason not to introduce it in this PR then I'll raise a ticket for it more generally.

app/services/clients/connector.client.js
oswaldquek: I'll do this

app/services/clients/connector.client.js
```diff
@@ -132,7 +69,7 @@ ConnectorClient.prototype = {
    *@return {Promise}
    */
   getAccounts: async function (params) {
-    const url = `${this.connectorUrl}${_accountsUrlFor(params.gatewayAccountIds)}`
+    const url = `${this.connectorUrl}/v1/api/accounts?accountIds=` + params.gatewayAccountIds.join(',')
```
nataliecarey, line 72: This `params.gatewayAccountIds.join(',')` should probably also be encoded (`encodeURIComponent(params.gatewayAccountIds.join(','))`), or if we want the comma character left unencoded it can be `params.gatewayAccountIds.map(encodeURIComponent).join(',')`.

nataliecarey (approved):
LGTM :)

app/services/clients/connector.client.js
oswaldquek: Oops, I missed this. Will fix this in a next PR.

PR #4247, opened 2024-08-13
PP-12876-Enable PACT test for create gateway account

james-peacock-gds:
Context: In a subsequent PR, serviceId will be made mandatory for create gateway account requests to connector. Selfservice already sends the serviceId but this isn't reflected in the PACT tests

- Enable PACT tests for create gateway account
- Update PACT tests for create gateway account to include serviceId
- Amend PACT test for failed create gateway account request to reflect that connector sends 422 for invalid parameter values


DomBelcher (approved):
LGTM

PR #4254, opened 2024-08-19
PP-13096: Refactored URL generation.

nataliecarey:
## WHAT
Refactoring URL generation for readability

I've gone through a number of files, enough to see a range of URLs that I believe are better handled by the new approach.

### DIFF
--- a/app/services/clients/utils/generateUrl.js
+++ b/app/services/clients/utils/generateUrl.js
@@ -1,0 +1,30 @@
+function generateUrl (template, urlParams = {}, queryParams = {}) {
+  const matches = template.match(/\{(\w+)}/g)
+  if (matches) {
+    const urlParamKeys = Object.keys(urlParams)
+    const expectedUrlParams = matches.map(x => x.substring(1, x.length - 1))
+    const missingUrlParams = expectedUrlParams.filter(x => !urlParamKeys.includes(x))
+    const additionalUrlParams = urlParamKeys.filter(x => !expectedUrlParams.includes(x))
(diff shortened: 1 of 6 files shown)


app/services/clients/utils/generateUrl.js
```diff
@@ -0,0 +1,30 @@
+function generateUrl (template, urlParams = {}, queryParams = {}) {
```
oswaldquek, line 1: I take it it doesn't matter what the path params are in the template? For e.g. `/v1/{example}` would be treated the same way as `/v1/{gatewayId}`?
Also although there are comprehensive tests for this function I'm finding it really hard to understand how it works.

oswaldquek: Can we use's [axios to encode query params](https://axios-http.com/docs/urlencoded)? Then for the normal url we can revert to 
```
`${this.connectorUrl}/v1/api/accounts/{accountId}/stripe-account`
.replace('{accountId}', encodeURIComponent(gatewayAccountId))
```

nataliecarey: You can choose any name for the template variables and they all need to be provided when using the template.

I think it's good to put the full URL generation together in one place rather than splitting some of it out to Axios.

The Axios implementation also does some quite specific things with arrays that I wouldn't expect to be compatible with our Java backends - they remind me of PHP's handling of query strings.

Even if we want to use Axios's approach I'd build it into the generator so it's a centralised change if we want to change it.

PR #4256, opened 2024-08-20
PP-13091 select organisation type view/controller

nlsteers:
## WHAT

- add new view and controller to enable user to select what type of organisation they work for when creating a new service
- create Stripe test account by default when selecting 'local' government, sandbox when selecting 'central'
- redirect user to new service on creation (was previously the 'My Services' overview)
- update psp test account stage when creating new stripe service
- introduce new 'SystemMessages' macro for displaying simple notifications to users
- show 'Service created successfully' message on successful creation of a new service
- layout bug fixes

### DIFF
--- a/app/controllers/create-service/create-service.controller.js
+++ b/app/controllers/create-service/create-service.controller.js
@@ -7,9 +7,7 @@ const paths = require('../../paths')
 const logger = require('../../utils/logger')(__filename)
 const serviceService = require('../../services/service.service')
 const userService = require('../../services/user.service')
-const { ConnectorClient } = require('../../services/clients/connector.client')
 const formatAccountPathsFor = require('../../utils/format-account-paths-for')
-const connectorClient = new ConnectorClient(process.env.CONNECTOR_URL)
 
 function get (req, res) {
   const createServiceState = _.get(req, 'session.pageData.createService', {})
@@ -37,7 +35,7 @@ async function post (req, res, next) {
       const { service, externalAccountId } = await serviceService.createService(serviceName, serviceNameCy, organisationType)
       await userService.assignServiceRole(req.user.externalId, service.externalId, 'admin')
       _.unset(req, 'session.pageData.createService')
-      req.flash('messages', { icon: '&check;', content: 'We\'ve created your service.' })
+      req.flash('messages', { state: 'success', icon: '&check;', content: 'We\'ve created your service.' })
       res.redirect(formatAccountPathsFor(paths.account.dashboard.index, externalAccountId))
     } catch (err) {
       next(err)
--- a/app/controllers/create-service/create-service.controller.js
+++ b/app/controllers/create-service/create-service.controller.js
@@ -24,21 +24,21 @@ async function post (req, res, next) {
   const serviceName = createServiceState.current_name.trim()
   const serviceNameCy = createServiceState.service_selected_cy && createServiceState.current_name_cy ? createServiceState.current_name_cy.trim() : ''
   const organisationType = req.body['select-org-type']
-  if (organisationType && (organisationType === 'central' || organisationType === 'local')) {
-    try {
-      const { service, externalAccountId } = await serviceService.createService(serviceName, serviceNameCy, organisationType)
-      await userService.assignServiceRole(req.user.externalId, service.externalId, 'admin')
-      _.unset(req, 'session.pageData.createService')
-      req.flash('messages', { state: 'success', icon: '&check;', content: 'We\'ve created your service.' })
-      res.redirect(formatAccountPathsFor(paths.account.dashboard.index, externalAccountId))
-    } catch (err) {
-      next(err)
-    }
-  } else {
+  if (!organisationType || organisationType !== 'central' || organisationType !== 'local') {
     _.set(req, 'session.pageData.createService.errors', {
       organisation_type: 'Organisation type is required'
     })
     return res.redirect(paths.serviceSwitcher.create.selectOrgType)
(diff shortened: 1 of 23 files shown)


app/controllers/create-service/select-organisation-type/select-organisation-type.controller.js
```diff
@@ -0,0 +1,62 @@
...
+    back_link: paths.serviceSwitcher.create.index,
+    submit_link: paths.serviceSwitcher.create.index
+  }
+  _.unset(req, 'session.pageData.createService.errors')
+  return response(req, res, 'services/select-org-type', context)
+}
+
+function post (req, res) {
+  _.set(req, 'session.pageData.createService', {
+    current_name: req.body['service-name'],
```
DomBelcher, line 28: might be simpler if both of these are `'service-name'`/`'service-name-cy'` vs mixing in `current_name`/`current_name_cy`

app/controllers/create-service/create-service.controller.js
```diff
@@ -1,57 +1,50 @@
...
 function get (req, res) {
-  const pageData = lodash.get(req, 'session.pageData.createServiceName', {})
-  lodash.unset(req, 'session.pageData.createServiceName')
-
-  pageData.submit_link = paths.serviceSwitcher.create
-  pageData.my_services = paths.serviceSwitcher.index
-
-  return response(req, res, 'services/add-service', pageData)
+  const createServiceState = _.get(req, 'session.pageData.createService', {})
+  const context = {
```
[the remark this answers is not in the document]
DomBelcher: Maybe something to think about re. improving the codebase though

app/controllers/create-service/select-organisation-type/select-organisation-type.controller.js
```diff
@@ -0,0 +1,62 @@
...
+  return response(req, res, 'services/select-org-type', context)
+}
+
+// --- PRIVATE
+
+function validateServiceName (serviceName, serviceNameCy) {
+  const errors = {}
+  const nameValidationResult = validateMandatoryField(serviceName, SERVICE_NAME_MAX_LENGTH, 'service name')
+  if (!nameValidationResult.valid) {
+    errors['service_name'] = nameValidationResult.message
```
DomBelcher, line 50: also best not to mix and match snake case and kebab case if possible

app/services/service.service.js
```diff
@@ -55,12 +56,26 @@ async function createService (serviceName, serviceNameCy) {
   const gatewayAccount = await connectorClient.createGatewayAccount('sandbox', 'test', serviceName, null, service.externalId)
```
DomBelcher, line 56: This could be renamed `sandboxGatewayAccount` to avoid any ambiguity

app/views/dashboard/index.njk
```diff
@@ -15,11 +16,20 @@ Dashboard - {{currentService.name}} {{currentGatewayAccount.full_type}} - GOV.UK
...
+      }) }}
+    </div>
   {% endif %}
+
+  <div class="govuk-grid-column-full">
+    {% if messages is defined and messages is iterable and messages|length > 0 %}
+      {{ systemMessages({ messages: messages }) }}
+    {% endif %}
+  </div>
+
```
DomBelcher, line 32: Is there any way both of these banners can show at the same time? May be a bit out of scope but ideally the system messages banner could encompass both use cases

app/views/team-members/team-member-profile.njk
```diff
@@ -36,7 +36,7 @@ My profile - GOV.UK Pay
         Email
       </dt>
       <dd class="profile-settings__value govuk-table__cell" id="email">
-        {{email}}
+        {{email | truncate(10, true)}}
```
DomBelcher, line 39: If we do want to truncate emails I think we should do it in a more careful way

app/controllers/create-service/create-service.controller.js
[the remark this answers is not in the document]
nlsteers: they are different objects though

we use `context` for the response render elsewhere, this was a (possibly in vain) attempt to start standardising

app/views/dashboard/index.njk
nlsteers: possibly, however the non-systemMessages banner is only temporary afaik (shown when s stripe test account is requested from the dashboard) so am not inclined to put in the effort to change it if it's being removed

DomBelcher (approved):
Nice

app/controllers/create-service/select-organisation-type/select-organisation-type.controller.js
```diff
@@ -0,0 +1,62 @@
...
+  const errors = validateServiceName(req.body['service-name'], req.body['service-name-cy'])
+  if (!_.isEmpty(errors)) {
+    _.set(req, 'session.pageData.createService.errors', errors)
+    return res.redirect(paths.serviceSwitcher.create.index)
+  }
+  const context = {
+    back_link: paths.serviceSwitcher.create.index,
+    submit_link: paths.serviceSwitcher.create.index
+  }
+  return response(req, res, 'services/select-org-type', context)
```
oswaldquek, line 41: Don't we want to create the service in this `post` method? This `return response(req, res, 'services/select-org-type', context)` is also the final line in the `get` method... shouldn't they be different?

nlsteers: no, we post back to the create service controller to finish up the service creation

PR #4260, opened 2024-08-29
PP-13111 Show spinner when creating a Stripe test account

nlsteers:
### WHAT
 - Add a loading spinner which shows when creating a Stripe test account. This shows in two scenarios:
   - When creating a new service for a local government organisation
   - When requesting a Stripe test account on a service with a Sandbox test account
     - The spinner does not show when creating a central government service, as this should not take long enough to warrant use of the spinner, and may cause issues for tools such as screen readers

### HOW TO TEST
 - There are some Cypress tests for this, but they aren't comprehensive as it doesn't seem possible to check both that the spinner shows and the form submits in one test. 
 - This can be tested locally by modifying a local instance of Connector to enforce a delay on the `/request-stripe-test-account` endpoint - see comment below for more info

### DIFF
--- a/app/views/includes/spinner.njk
+++ b/app/views/includes/spinner.njk
@@ -1,0 +1,7 @@
+{% macro spinner(params) %}
+  <div class="spinner"></div>
+  {% if params.text %}
+    <h3 class="govuk-heading-m" style="text-align: center">{{ params.text }}</h3>
+  {% endif %}
+{% endmacro %}
+
--- a/app/views/includes/spinner.njk
+++ b/app/views/includes/spinner.njk
@@ -1,7 +1,7 @@
 {% macro spinner(params) %}
   <div class="spinner"></div>
   {% if params.text %}
-    <h3 class="govuk-heading-m" style="text-align: center">{{ params.text }}</h3>
+    <h3 class="govuk-heading-m govuk-!-text-align-centre">{{ params.text }}</h3>
   {% endif %}
 {% endmacro %}
 
--- a/app/views/request-psp-test-account/index.njk
+++ b/app/views/request-psp-test-account/index.njk
@@ -42,6 +42,9 @@
           <p class="govuk-body">
             You will need to create new API keys and add them to your integration as your old keys will stop working.
           </p>
+          <p class="govuk-body">
+            Creating your Stripe account may take around ten seconds.
+          </p>
 
           {{ govukButton({
             html: "Get a Stripe test account",
--- a/app/views/request-psp-test-account/index.njk
+++ b/app/views/request-psp-test-account/index.njk
@@ -94,10 +94,12 @@
   </div>
 
   <script>
-    document.getElementById('submit-button').addEventListener('click', function () {
-      disableButtonAndShowSpinner()
-      document.getElementById('submit-request-for-psp-test-account-form').requestSubmit()
-    })
+    if (document.getElementById('submit-button')) {
+      document.getElementById('submit-button').addEventListener('click', function () {
+        disableButtonAndShowSpinner()
+        document.getElementById('submit-request-for-psp-test-account-form').requestSubmit()
+      })
+    }
 
     function disableButtonAndShowSpinner () {
       document.getElementById('submit-request-for-psp-test-account-form').hidden = true
--- a/app/views/request-psp-test-account/index.njk
+++ b/app/views/request-psp-test-account/index.njk
@@ -53,7 +53,7 @@
           }) }}
         </form>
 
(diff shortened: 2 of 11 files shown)


DomBelcher:
This can be tested locally by checking out [this branch](https://github.com/alphagov/pay-connector/tree/PP-13111/delay-on-request-stripe-account-for-testing) in Connector, and running `pay local --local selfservice connector`. The connector branch replaces the API calls to Stripe for creating the Stripe account with an 8 second delay, bypassing the need for a Stripe API key

app/views/services/select-org-type.njk
```diff
@@ -78,7 +78,9 @@
 
   <script>
     document.getElementById('submit-button').addEventListener('click', function () {
-      disableButtonAndShowSpinner()
+      if (document.getElementById('org-type-local').checked) {
```
nlsteers, line 81: is it necessary to perform this check? If so, you probably also want to account for the user not selecting anything as that will also result in a page load

test/cypress/integration/my-services/add-new-service.cy.js
```diff
@@ -141,5 +141,42 @@ describe('Add a new service', () => {
...
+          serviceExternalId: newServiceId,
+          gatewayAccountId: newGatewayAccountId,
+          serviceName: { en: newServiceName }
+        }),
+        serviceStubs.patchUpdateServiceGatewayAccounts({ serviceExternalId: newServiceId }),
+        gatewayAccountStubs.getGatewayAccountByExternalIdSuccess({
+          gatewayAccountExternalId: 'a-valid-external-id',
+          gatewayAccountId: '1'
+        }),
+        transactionsSummaryStubs.getDashboardStatistics()
```
nlsteers: are all these stubs needed if we're not actually submitting the form?

app/views/includes/spinner.njk
```diff
@@ -0,0 +1,7 @@
+{% macro spinner(params) %}
+  <div class="spinner"></div>
+  {% if params.text %}
+    <h3 class="govuk-heading-m" style="text-align: center">{{ params.text }}</h3>
```
nlsteers, line 4: remove this in-line style (I know I added this but that was back when this was a WIP)

app/views/request-psp-test-account/index.njk
```diff
@@ -9,77 +10,108 @@
...
+          </p>
+
+          {{ govukButton({
+            html: "Get a Stripe test account",
+            id: "submit-button",
+            preventDoubleClick: true
+          }) }}
+        </form>
+
+        <div id="spinner" hidden="hidden" aria-hidden="true">
```
nlsteers, line 56: this id should probably be more accurately named to something like `spinner-container`, alternatively consider moving the containing div into the nunjucks macro so that it doesn't need to be duplicated

app/views/services/select-org-type.njk
```diff
@@ -60,8 +62,36 @@
...
 
-      {{ govukButton({ text: "Create service" })}}
+      {{ govukButton({
+        html: "Create service <span class=\"spinner\"></span>",
+        id: "submit-button",
+        preventDoubleClick: true
+      }) }}
     </form>
 
+    <div id="spinner" hidden="hidden" aria-hidden="true">
```
nlsteers, line 72: see above re macro improvements

nlsteers (commented):
nice job finishing this off, some small comments

app/views/services/select-org-type.njk
DomBelcher: The button doesn't necessarily show e.g. because you can't request a Stripe test account if your service is live or you've already submitted a request (under the old system where it was done manually). The code for manual creation can probably be removed though

app/views/includes/spinner.njk
DomBelcher: Fixed

app/views/services/select-org-type.njk
```diff
@@ -60,8 +62,36 @@
...
     </form>
 
+    <div id="spinner-container" hidden="hidden" aria-hidden="true">
+      {{ spinner({
+        text: "We're creating your service, please wait..."
+      }) }}
+    </div>
   </div>
+
+  <script>
```
nataliecarey, line 79: Can we move the JS out of this page and into a JS file please?  It can either:

 - Be a specific JS file that's loaded into this page only
 - Be included in a more general JS bundle and only trigger on this page

DomBelcher: This should be fixed now

PR #4263, opened 2024-09-06
PP-13117 fix correlation id issue

SandorArpa:
## WHAT

The Node apps were updated to use middleware that used the async_hooks - to help with logging. However, there is an issue where the correlation is not being set correctly and not being passed to the backend call.

## HOW 

- fix issue (add crypto generated id if missing)
- remove `app/utils/request-context.js` and refactor code to use `app/services/clients/base/request-context.js`


iqbalgds (approved):
Looks good!

PR #4267, opened 2024-09-10
PP-13121-Hide API Keys link

james-peacock-gds:
- Do not show API Keys link in Settings for a disabled gateway account


DomBelcher (approved):
Nice work

PR #4274, opened 2024-09-20
PP-13030: Show WORLDPAY TEST SERVICE label

oswaldquek:
A service is a worldpay test service if it only has one test gateway account which has payment provider of "worldpay".

This will not break any current services because it used to be that when a service was created, a sandbox gateway account would be created. We would then create another worldpay test gateway account (via toolbox) on request. In this case there would be two test gateway accounts on the service.

A screenshot to show some local testing:
Screenshot 2024-09-20 at 16 14 41


nlsteers (approved):
LGTM, nice job

PR #4280, opened 2024-10-03
Spike - Add formatting help for payment link description

stephencdaly:
Add a details section with information about how to add markdown to the payment link description.

This also needs to be added in `app/views/payment-links/edit-information.njk`. Although it's recommended that the create and edit views for payment links are merged to avoid duplication.

Screenshot 2024-10-03 at 11 11 52

### DIFF
--- a/app/views/payment-links/information.njk
+++ b/app/views/payment-links/information.njk
@@ -2,7 +2,7 @@
 {% from "../macro/error-summary.njk" import errorSummary %}
 
 {% block pageTitle %}
-  Set payment link information - Create a payment link - {{currentService.name}} {{currentGatewayAccount.full_type}} - GOV.UK Pay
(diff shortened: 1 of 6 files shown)


app/views/payment-links/information.njk
```diff
@@ -94,24 +93,67 @@
...
         }
-      })
-    }}
+      }) }}
+
+      {% set formattingHelpHTML %}
+        <h3 class="govuk-heading-m">Links and URLs</h3>
+
+        <p>To add a link, use square brackets [ ] around the link text, and round brackets ( ) around the full URL. Make
+          sure there are no spaces between the two sets of brackets. For example:</p>
```
katstevens, line 102: [Notify have got some really solid Markdown advice in their docs](https://www.notifications.service.gov.uk/using-notify/links-and-URLs). I'd recommend we lift as much as possible from their guidance to help our service users create good content (for example recommending avoiding third party url-shorteners).

PR #4282, opened 2024-10-09
PP-13133 simplified account settings - edit service name

nlsteers:
## WHAT

- added middlewares and router supporting simplified account routes
  - simplified account opt in: directs the user to an appropriate error message if a simplified account route is accessed without the user being opted in and the feature being enabled for the environment
  - simplified account strategy: retrieves the correct gateway account for the service and account type and makes this available on the request object
- added service settings navigation nunjucks template and supporting sass
- added settings controllers
  - index: redirects the user to the correct default setting based on the account type and go live stage of the service
  - service name: get and post actions for viewing and updating the current service name in english and welsh
  - email notifications: bare bones implementation of the view
- external dependency: express-validations, creates easy to understand validation chains used for parsing request bodies
- added utilities
  - nunjucks filter for correctly capitalising service settings and categories
  - generator for simplified account paths
  - settings builder: determines which settings should be visible based on the service go live stage, account type and permissions of the viewing user


app/controllers/simplified-account/settings/index.controller.js
```diff
@@ -0,0 +1,20 @@
...
+function get (req, res) {
+  const account = req.account
+  const service = req.service
+  // the default setting for the index view is dependent on the account type and go live state
+  if (account.type === 'test' && service.currentGoLiveStage !== LIVE) {
+    return res.redirect(formatSimplifiedAccountPathsFor(paths.simplifiedAccount.settings.serviceName.index, req.account.service_id, req.account.type))
+  } else if (account.type === 'live') {
+    return res.redirect(formatSimplifiedAccountPathsFor(paths.simplifiedAccount.settings.serviceName.index, req.account.service_id, req.account.type))
+  } else {
+    return res.redirect(formatSimplifiedAccountPathsFor(paths.simplifiedAccount.settings.emailNotifications.index, req.account.service_id, req.account.type))
```
oswaldquek, line 14: I don't think we've taken into account the user permission here. We also want to show the email notifications page if the is a non-admin user. 
I think for this entire `get` function the logic could be:

```
if ( (user is non admin) OR (service is live AND account type is 'test') ) {
  redirect to email notifications page
} else {
  redirect to service name page
}
```

app/utils/display-converter.js
```diff
@@ -138,9 +139,13 @@ module.exports = function (req, data, template) {
   convertedData.isLive = req.isLive
   convertedData.humanReadableEnvironment = convertedData.isLive ? 'Live' : 'Test'
   const currentPath = (relativeUrl && url.parse(relativeUrl).pathname.replace(/([a-z])\/$/g, '$1')) || '' // remove query params and trailing slash
+  const currentUrl = req.baseUrl && req.baseUrl ? req.baseUrl + req.path : 'unavailable'
```
DomBelcher, line 142: Is the `req.baseUrl &&` necessary if it's also in the ternary? i.e. would just the ternary
`req.baseUrl ? req.baseUrl + req.path : 'unavailable'`
be equivalent

app/controllers/simplified-account/settings/index.controller.test.js
```diff
@@ -0,0 +1,53 @@
...
+      req.account.type = 'live'
+      indexController.get(req, res)
+      expect(res.redirect.calledWith(formatSimplifiedAccountPathsFor(paths.simplifiedAccount.settings.serviceName.index, SERVICE_ID, LIVE_ACCOUNT_TYPE))).to.be.true // eslint-disable-line
+    })
+
+    it('should redirect to email notifications index for test account on service with live account', () => {
+      req.service.currentGoLiveStage = LIVE
+      indexController.get(req, res)
+      expect(res.redirect.calledWith(formatSimplifiedAccountPathsFor(paths.simplifiedAccount.settings.emailNotifications.index, SERVICE_ID, TEST_ACCOUNT_TYPE))).to.be.true // eslint-disable-line
+    })
```
oswaldquek, line 51: Another test for non-admin user would be good here too.

app/controllers/simplified-account/settings/index.controller.js
nataliecarey, line 14: Either the `else`s or the `return`s can be removed here, both together is unnecessary.

Option a:

```
if (condition1) {
  return doAction1()
}
if (condition2) {
  return doAction2()
}
doDefaultAction()
```

Option b:

```
if (condition1) {
  doAction1()
} else if (condition2) {
  doAction2()
} else {
  doDefaultAction()
}
```

They are functionally equivalent so I think it's a case of chosing which one you prefer.

app/utils/nav-builder.js
```diff
@@ -23,7 +24,7 @@ const yourPspPaths = ['your-psp', 'notification-credentials']
 const additionalPspPaths = ['switch-psp']
 const webhookPaths = ['webhooks']
 
-const serviceNavigationItems = (currentPath, permissions, type, account = {}) => {
+const serviceNavigationItems = (currentPath, permissions, type, account = {}, isDegatewayed, currentUrl) => {
```
DomBelcher, line 27: Better to keep parameters with defaults at the end of the list

app/controllers/simplified-account/settings/index.controller.js
```diff
@@ -0,0 +1,20 @@
+const formatSimplifiedAccountPathsFor = require('../../../utils/simplified-account/format/format-simplified-account-paths-for')
+const paths = require('../../../paths')
+const { LIVE } = require('../../../models/go-live-stage')
+
+function get (req, res) {
+  const account = req.account
+  const service = req.service
+  // the default setting for the index view is dependent on the account type and go live state
+  if (account.type === 'test' && service.currentGoLiveStage !== LIVE) {
```
nataliecarey, line 9: Here we've got an `enum` (probably just a basic object acting as an enum) being used to check the go-live status against and a string literal for the account type.  I'm wondering whether we should introduce an `enum` (just a basic object acting as an enum) for test and live, that's for two reasons:

 - To track usage across the app
 - Just in case the value changes at some point in the future

app/utils/simplified-account/settings-builder.js
```diff
@@ -0,0 +1,124 @@
...
+
+module.exports = (account, currentUrl, goLiveStage, permissions) => {
+  const serviceSettings = {
+    'about your service': [],
+    'payments': [],
+    'developers': []
+  }
+  buildAboutServiceSettings(serviceSettings, account, currentUrl, permissions)
+  buildPaymentSettings(serviceSettings, account, currentUrl)
+  buildDeveloperSettings(serviceSettings, account, currentUrl)
```
DomBelcher, line 13: Do these return values just get thrown away?

app/middleware/simplified-account/simplified-account-strategy.middleware.js
```diff
@@ -0,0 +1,92 @@
...
+
+    if (err.errorCode === 404) {
+      logger.info('Gateway account not found', logContext)
+    } else {
+      logger.error('Error retrieving gateway account', logContext)
+    }
+  }
+}
+
+module.exports = async function getSimplifiedAccount (req, res, next) {
```
oswaldquek, line 63: I'm not sure about the method name `getSimplifiedAccount` here. The data model we have is simply "gateway account", not "simplified gateway account". I'd rename it to just `getGatewayAccount`.

app/middleware/simplified-account/simplified-account-strategy.middleware.test.js
```diff
@@ -0,0 +1,209 @@
...
+      errorCode: '404'
+    })
+    await simplifiedAccountStrategy(req, res, next)
+
+    const expectedError = sinon.match.instanceOf(Error)
+      .and(sinon.match.has('message', 'getGatewayAccountByServiceIdAndAccountType failed for provided parameters'))
+    sinon.assert.calledWith(next, expectedError)
+  })
+  describe('extend gateway account data with disableToggle3ds field', () => {
+    ['worldpay', 'smartpay', 'epdq'].forEach(function (paymentProvider) {
```
oswaldquek, line 126: Do we need smartpay and epdq here?

app/middleware/simplified-account/simplified-account-strategy.middleware.test.js
```diff
@@ -0,0 +1,209 @@
...
+        serviceExternalId: A_SERVICE_EXTERNAL_ID,
+        accountType: 'test'
+      })
+      await simplifiedAccountStrategy(req, res, next)
+      expect(req.account.disableToggle3ds).to.equal(true)
+      expect(req.account.external_id).to.equal(A_GATEWAY_EXTERNAL_ID)
+    })
+  })
+  describe('extend gateway account data with supports3ds field', () => {
+    ['worldpay', 'smartpay', 'epdq', 'stripe'].forEach(function (paymentProvider) {
```
oswaldquek, line 154: Do we need smartpay and epdq here?

app/controllers/simplified-account/settings/service-name/service-name.controller.js
```diff
@@ -0,0 +1,69 @@
...
+  // we don't check presence for welsh names
+  if (!editCy) {
+    validations.push(body('service-name-input').trim().notEmpty().withMessage('Service name is required'))
+  }
+
+  await Promise.all(validations.map(validation => validation.run(req)))
+  const errors = validationResult(req)
+  if (!errors.isEmpty()) {
+    const formattedErrors = formatValidationErrors(errors)
+    return response(req, res, 'simplified-account/settings/service-name/edit-service-name', {
```
nataliecarey, line 48: Is this our normal approach?  Responding to a browser `POST` with anything other than a redirect is usually not the best.  If this is how we normally do it then I don't see a reason to change that as part of this PR.

nataliecarey: I've just taken a quick look and seen at least one other place we're doing this.  I'll add it to my niggle list and there's no reason to update this PR.

app/middleware/simplified-account/simplified-account-strategy.middleware.js
```diff
@@ -0,0 +1,92 @@
...
+        throw new Error('Could not resolve service external ID or gateway account type from request params')
+      }
+
+      let gatewayAccount = await getGatewayAccountByServiceIdAndAccountType(serviceExternalId, accountType)
+      if (gatewayAccount) {
+        req.account = gatewayAccount
+        addField(keys.GATEWAY_ACCOUNT_ID, gatewayAccount.gateway_account_id)
+        addField(keys.GATEWAY_ACCOUNT_TYPE, gatewayAccount.type)
+      } else {
+        throw new Error('getGatewayAccountByServiceIdAndAccountType failed for provided parameters')
```
nataliecarey, line 79: Should this be `next(new Error(...))`?  I'm not sure if our handling makes those equivilent or not.

(If you're doing that you might also need to stop the other `next()` from firing.

app/middleware/simplified-account/simplified-account-strategy.middleware.test.js
```diff
@@ -0,0 +1,209 @@
...
+    gatewayAccountExternalId,
+    paymentProvider,
+    serviceExternalId,
+    accountType,
+    errorCode
+  } = options
+
+  req = {
+    params: { serviceExternalId, accountType },
+    user: buildUser(serviceExternalId, [`${gatewayAccountID}`])
```
nataliecarey, line 37: Should `[`${gatewayAccountID}`]` just be `[gatewayAccountID]`? Are you intentionally casting it to a string?

app/utils/simplified-account/settings/service-settings.js
```diff
@@ -0,0 +1,94 @@
...
+
+module.exports = (account, currentUrl, goLiveStage, permissions) => {
+  const settingsBuilder = new ServiceSettings(account, currentUrl, permissions, formatSimplifiedAccountPathsFor)
+  const result = settingsBuilder
+    .category('about your service')
+    .add({
+      id: 'service-name',
+      name: 'service name',
+      path: paths.simplifiedAccount.settings.serviceName.index,
+      permission: 'service_name_update' // TODO find a better way of defining these
```
oswaldquek, line 14: Nice 🎉

app/views/simplified-account/settings/service-name/index.njk
```diff
@@ -0,0 +1,63 @@
...
+{% endset %}
+
+{% set add_cy_link %}
+  <a class='govuk-link govuk-link--no-visited-state' href='{{ manage_cy }}'>Add Welsh service name</a>
+{% endset %}
+
+{% block mainContent %}
+  <div class="govuk-grid-column-three-quarters">
+    <h2 class="govuk-heading-l">Service name</h2>
+    {{ govukTable({
```
nataliecarey, line 29: Is there a reason we're not using the [Check Your Answers pattern](https://design-system.service.gov.uk/patterns/check-answers/) for this?

The component used there is `govukSummaryList`.

app/utils/simplified-account/settings-builder.js
```diff
@@ -0,0 +1,124 @@
...
+  return serviceSettings
+}
+
+const buildDeveloperSettings = (serviceSettings, account, currentUrl) => {
+  serviceSettings['developers'].push({
+    id: 'api-keys',
+    name: 'API keys',
+    url: formatSimplifiedAccountPathsFor(paths.simplifiedAccount.settings.index, account.service_id, account.type),
+    current: currentUrl.includes('simplified') && currentUrl.includes('settings/team-members'),
+    permitted: true, // TODO
```
nataliecarey: There's a TODO here, is that something that should be resolved before merging?

app/middleware/simplified-account/simplified-account-strategy.middleware.js
nlsteers: The name is temporary and allows us to easily differentiate between the new middleware strategy and the old one. 

I agree that it doesn't match the data model but it's important to make the differences obvious whilst we're in a transitionary phase

app/middleware/simplified-account/simplified-account-strategy.middleware.test.js
nlsteers: we can always add them back if we need to later, removed

app/utils/simplified-account/settings-builder.js
nlsteers: no, this PR scaffolds quite a bit of work yet to come. Any TODOs are to serve as 'helpful' reminders

app/views/simplified-account/settings/service-name/index.njk
nlsteers: the reason is I did not know there was a component for this already! Will refactor

app/controllers/simplified-account/settings/service-name/service-name.controller.js
nlsteers: Yeah this is a common pattern across the base.

We can update it though, I'd like to set some best practice examples for how to build the rest of the settings in this PR for others to follow.

What would you suggest / why is `.render` inappropriate here? 

We aren't changing URL (in the case of form errors) so `.render` felt like the appropriate choice.

app/middleware/simplified-account/simplified-account-strategy.middleware.js
oswaldquek: Yes but the file is called "simplified-account-strategy.middleware.js". But no big deal though.

app/utils/display-converter.js
nlsteers: ah good spot, it's supposed to be `req.baseUrl && req.path...`

app/middleware/simplified-account/simplified-account-strategy.middleware.test.js
nlsteers: It's intentional, that's how Users are set on the request

app/controllers/simplified-account/settings/service-name/service-name.controller.js
nataliecarey: At this point it's become an established way that I work and I haven't considered the reasons for a while.  I'll look up some stuff about it.

The most direct difference for the user is that if they refresh the page on that page they get a warning in the browser describing the dangers of repeating a `POST` request which isn't actually an issue in this case.

This PR introduces some good things, I don't think we need to introduce the redirect in the same PR (but if we were already using it we should have followed it which is where my initial comment started).

oswaldquek: The post-redirect-get pattern is common thing, but I wouldn't know how this would work with passing a bunch of contextual info like
```{
      errors: {
        summary: formattedErrors.errorSummary,
        formErrors: formattedErrors.formErrors
      },
      edit_cy: editCy,
      service_name: req.body['service-name-input'],
      back_link: formatSimplifiedAccountPathsFor(paths.simplifiedAccount.settings.serviceName.index, req.account.service_id, req.account.type),
      submit_link: formatSimplifiedAccountPathsFor(paths.simplifiedAccount.settings.serviceName.edit, req.account.service_id, req.account.type)
    }
```
to the view. 
Be good if we have a look at this in the future though.

oswaldquek:
There are some comments that would be good to address before this is approved:
https://github.com/alphagov/pay-selfservice/pull/4282/files#r1799127771
https://github.com/alphagov/pay-selfservice/pull/4282/files#r1799266722
https://github.com/alphagov/pay-selfservice/pull/4282/files#r1799586386

app/middleware/simplified-account/simplified-account-strategy.middleware.js
nlsteers: I think it does, throwing an error here correctly redirects the user to an appropriate 'something went wrong' page

app/middleware/simplified-account/simplified-account-strategy.middleware.js
nlsteers: I will find a better naming convention in the next PR

oswaldquek (dismissed):
🎉

oswaldquek (approved):

PR #4287, opened 2024-10-14
PP-13235: Using a mountebank replacement

nataliecarey:
## WHAT

As Mountebank is being deprecated we're moving onto our own `run-amock` project. which is a drop-in replacement for the functionality we've been using from Mountebank.

(related to https://github.com/alphagov/pay-frontend/pull/3913 and https://github.com/alphagov/pay-products-ui/pull/2713)


nataliecarey:
@marcotranchino Because it's a drop-in replacement for the parts of Mountebank we're using it needs some references to Mountebank - I'm expecting that the project will evolve and we'll make our own decisions about how we want these mocks to act, when we do that we can start removing references to Mountebank.

PR #4284, opened 2024-10-10
PP-13174 Do not show test stripe link for new services

james-peacock-gds:
Context: Changes to the onboarding flow on 29/08/2024 mean that local gov services are now automatically allocated a Stripe test account. Therefore the link to create a Stripe test account should not be displayed for services that onboarded since that date.

- add condition to only show the create Stripe test account link to services onboarding prior to 29/08/2024
- update Cypress tests accordingly

PR #4288, opened 2024-10-14
PP-13174 Update cutoff time for removing link to request stripe test account

DomBelcher:
## WHAT
 - Updates the cutoff time added in #4284 to approximately the time when the change to create a Stripe test account as part of the onboarding flow was deployed.
 - This is approx. `2024-08-29T10:08:00Z`, the time PR  #4256 was deployed (release tag `alpha_release-3056`, concourse build `698`)
 - There was one service created on  2024-08-29 before the above change went live which currently would be unable to request a Stripe test account, this change should rectify that


james-peacock-gds (approved):
LGTM

nataliecarey (approved):
LGTM

PR #4291, opened 2024-10-16
PP-13258 Update JS Standard style to v17.1.2

DomBelcher:
## WHAT
Update JS Standard style to v17.1.2

JS standard v17 enforces [object shorthand](https://eslint.org/docs/latest/rules/object-shorthand) where applicable. This is incompatible with IE, however this update should be safe as the frontend code in `assets` is not linted, and none of the code in `browsered` appears to make use of object shorthand

I'm also about 90% sure we don't need to worry about compatibility because any client facing JS is transpiled by babel, and we're using `@babel/preset-env` with no target specified, which means it uses the default ES5 target (going by the [babel docs](https://babeljs.io/docs/options#targets))


app/controllers/credentials/credentials-form.js
```diff
@@ -32,6 +32,7 @@ class CredentialsForm {
           errors[field.id] = validator.message
           return true
         }
+        return false
```
katstevens, line 35: I'd strongly recommend sanity-checking this form and its test coverage, just to make sure there's no unexpected behaviour with this change. It might look innocuous (and may well be 100% fine!) but we've been bitten before in these scenarios.

DomBelcher: I've added a more comprehensive test, I'm 99% sure this should be fine because the callback provided to `Array.prototype.some` should return a boolean value, and returning nothing would be the same was returning `undefined`, which is falsy, so returning `false` instead shouldn't affect the functionality

katstevens: Fab, thank you 👍 

I will leave it to other folks to review the rest of the PR 😸

app/models/DirectDebitGatewayAccount.class.js
```diff
@@ -33,6 +33,7 @@ class DirectDebitGatewayAccount {
     // compatibility with other parts of selfservice - recording as tech debt in jira
     this.payment_provider = gatewayAccountData.payment_provider
   }
+
```
nlsteers, line 36: huge

nlsteers (approved):
Let's not leave it this long again, think I went blind towards the end. Nice job

PR #4295, opened 2024-10-18
PP-13234: Using latest run-amock

nataliecarey:
## WHAT
Using the latest version of `run-amock` which uses case-sensitive query strings as we no longer have to maintain feature-compatibility with Mountebank.

This new version treats query strings as case-sensitive (as our production apps do) so one mock needed updating.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,7 +57,7 @@
         "@babel/core": "7.22.10",
         "@babel/polyfill": "7.12.1",
         "@babel/preset-env": "7.22.10",
-        "@govuk-pay/run-amock": "0.0.1",
+        "@govuk-pay/run-amock": "0.0.3",
         "@pact-foundation/pact": "^12.1.1",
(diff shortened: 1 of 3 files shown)


test/cypress/stubs/webhooks-stubs.js
```diff
@@ -21,7 +21,7 @@ function getWebhookMessagesListSuccess (opts = {}) {
   return stubBuilder('GET', path, 200, {
     query: {
       page: opts.page || 1,
-      ...opts.status && { status: opts.status }
+      ...opts.status && { status: opts.status.toUpperCase() }
```
DomBelcher, line 24: I'll admit I have little context on this but would it not make more sense to update the status where this is called, rather than obscure that it becomes upper case?

nataliecarey: Fair point, I've updated at the source.

DomBelcher (approved):
👍

PR #4300, opened 2024-10-22
BAU - match agreed case style for simplified accounts work

nlsteers:
## WHAT

- prefer camel over snake for object key notation


DomBelcher (approved):
👍

PR #4316, opened 2024-11-07
PP-13135: Edit send payment confirmation email functionality

oswaldquek:
Screenshot:
Screenshot 2024-11-07 at 15 09 23

As usual cypress tests will be in another PR.


app/controllers/simplified-account/settings/email-notifications/payment-confirmation-emails.controller.test.js
```diff
@@ -0,0 +1,103 @@
...
+        body: {
+          paymentConfirmationEmailToggle: 'true'
+        }
+      })
+      paymentConfirmationEmailsController.post(req, res)
+    })
+
+    it('should update refund email enabled', () => {
+      expect(setConfirmationEnabledByServiceIdAndAccountTypeStub.calledOnce).to.be.true // eslint-disable-line
+      expect(setConfirmationEnabledByServiceIdAndAccountTypeStub.calledWith(SERVICE_ID, ACCOUNT_TYPE, 'true')).to.be.true // eslint-disable-line
```
nlsteers, line 95: you could do something like
```js
sinon.assert.calledWith(setConfirmationEnabledByServiceIdAndAccountTypeStub, SERVICE_ID, ACCOUNT_TYPE, 'true')
```
which would give you a richer debugging response in the future if this assertion ever started failing. Currently this will just produce a `was false, expected true` on test failure which doesn't tell you anything about which args we're incorrect

app/controllers/simplified-account/settings/index.js
```diff
@@ -2,4 +2,5 @@ module.exports.index = require('./index.controller')
 module.exports.serviceName = require('./service-name/service-name.controller')
 module.exports.emailNotifications = require('./email-notifications/email-notifications.controller')
 module.exports.refundEmails = require('./email-notifications/refund-emails.controller')
+module.exports.paymentConfirmationEmails = require('./email-notifications/payment-confirmation-emails.controller')
```
nlsteers, line 5: Consider exporting the sub settings controllers in the `email-notifications.controller`, then you only need to add the parent setting controller to this index.

It also makes the code more readable when creating routes (imo) e.g. 
```js
serviceSettingsController.emailNotifications.paymentConfirmationEmails.get
```

nlsteers (approved):

PR #4329, opened 2024-11-13
PP-13298 Cypress test for remove users

james-peacock-gds:
Cypress test for removing a user from a service

### DIFF
--- a/test/cypress/integration/simplified-account/service-settings/team-members/remove-team-member.cy.js
+++ b/test/cypress/integration/simplified-account/service-settings/team-members/remove-team-member.cy.js
@@ -1,0 +1,83 @@
+const userStubs = require('../../../../stubs/user-stubs')
+const gatewayAccountStubs = require('../../../../stubs/gateway-account-stubs')
+const inviteStubs = require('../../../../stubs/invite-stubs')
+
+const ADMIN_USER_ID = 'admin-user-id'
+const VIEW_ONLY_USER_ID = 'view-only-user-id'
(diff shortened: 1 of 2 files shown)


test/cypress/integration/simplified-account/service-settings/team-members/remove-team-member.cy.js
```diff
@@ -0,0 +1,83 @@
...
+      cy.setEncryptedCookies(ADMIN_USER_ID)
+      cy.visit(TEAM_MEMBERS_SETTINGS_URL)
+      cy.get('#team-members-view-only-list').find('dl').first().find('a').contains('Remove').click()
+    })
+
+    it('should show the show the correct heading, title and form with correct elements', () => {
+      cy.get('h1').should('contain', 'Are you sure you want to remove view-only-user@example.com?')
+      cy.title().should('eq', 'Settings - Team members - Remove user - GOV.UK Pay')
+      cy.get('[data-cy=yes-radio]').should('not.be.checked')
+      cy.get('[data-cy=no-radio]').should('not.be.checked')
```
DomBelcher, line 56: I think it's better to avoid using `data-cy` unless absolutely necessary. It should be possible to select the radios based on the value, with a selector like `[value=yes]`

james-peacock-gds: updated

DomBelcher (approved):
LGTM

PR #4330, opened 2024-11-13
Pp 12020 upgrade to dsv5

iqbalgds:
PP-12020 Upgrade to design system v5
- Add the changes from the DSv5 release notes - https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0
- Most changes were updates to ensure compatibility with the updated tag component - https://design-system.service.gov.uk/components/tag/
  - Which requires all text to be in sentence case - i.e. the first letter needs to be in
    uppercase.
- These changes have been checked using Percy.
- Suppress warning about size 14 is deprecated and being removed in v6.  A seperate jira has been created to look into this.  - [Jira ticket](https://payments-platform.atlassian.net/browse/PP-13352)

A comparison of the old tag and new tag can be seen below:
image


app/views/simplified-account/settings/stripe-details/bank-account/index.njk
```diff
@@ -77,7 +77,6 @@
     document.getElementById('bank-account-form').addEventListener('submit', function () {
       submitButton.setAttribute('disabled', 'true')
       submitButton.setAttribute('aria-disabled', 'true')
-      submitButton.setAttribute('class', 'govuk-button govuk-button--disabled')
```
nataliecarey, line 80: I presume that the behaviour is still the same after removing this?

app/views/webhooks/message.njk
```diff
@@ -54,7 +54,7 @@
   }) }}
 
   <div class="govuk-body">
-    <details class="govuk-details" data-module="govuk-details">
+    <details class="govuk-details" >
```
nataliecarey, line 57: Tiny thing - there's a stray space in this element.

nataliecarey:
Just two small comments and the branch has conflicts.  Other than that it all looks good to me.

app/views/simplified-account/settings/stripe-details/bank-account/index.njk
iqbalgds: Yes - I tested this by creating a little test.
Good questions.

app/views/webhooks/message.njk
iqbalgds: Good spot - fixed.

nataliecarey (approved):

PR #4334, opened 2024-11-14
PP-13297 set up controller and template

james-peacock-gds:
- set up controller and template for 'change permission'
- will implement the functionality in a subsequent PR

Screenshot (role headings will be made bolder in a subsequent PR)
Screenshot 2024-11-14 at 16 46 10

### DIFF
--- a/app/controllers/simplified-account/settings/team-members/change-permission/change-permission.controller.js
+++ b/app/controllers/simplified-account/settings/team-members/change-permission/change-permission.controller.js
@@ -1,0 +1,34 @@
+const { findByExternalId } = require('@services/user.service')
+const { response } = require('@utils/response')
+const formatSimplifiedAccountPathsFor = require('../../../../../utils/simplified-account/format/format-simplified-account-paths-for')
+const paths = require('@root/paths')
+
+async function get (req, res, next) {
+  const externalServiceId = req.service.externalId
+  const accountType = req.account.type
+  const serviceHasAgentInitiatedMotoEnabled = req.service.agentInitiatedMotoEnabled
+  try {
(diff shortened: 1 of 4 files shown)


app/controllers/simplified-account/settings/team-members/change-permission/change-permission.controller.js
```diff
@@ -0,0 +1,33 @@
+const { findByExternalId } = require('@services/user.service')
+const { response } = require('@utils/response')
+const formatSimplifiedAccountPathsFor = require('../../../../../utils/simplified-account/format/format-simplified-account-paths-for')
```
DomBelcher, line 3: You can use `require('@utils/...')` here as well

app/controllers/simplified-account/settings/team-members/change-permission/change-permission.controller.js
```diff
@@ -0,0 +1,33 @@
...
+const paths = require('@root/paths')
+
+async function get (req, res, next) {
+  const externalServiceId = req.service.externalId
```
DomBelcher, line 7: I think just `serviceId` is fine and more in line with usage elsewhere

app/controllers/simplified-account/settings/team-members/change-permission/change-permission.controller.js
james-peacock-gds: done

app/controllers/simplified-account/settings/team-members/change-permission/change-permission.controller.js
james-peacock-gds: yeah good point

DomBelcher (approved):
LGTM

PR #4335, opened 2024-11-15
bau: Using singluar patch email method

oswaldquek:
Use singular `patchEmailNotificationByServiceIdAndAccountType` method on the connector client. This helps with keeping things DRY.

### DIFF
--- a/app/services/clients/connector.client.js
+++ b/app/services/clients/connector.client.js
@@ -424,25 +424,6 @@ ConnectorClient.prototype = {
   },
 
   /**
-   * @param {string} serviceId
-   * @param {string} accountType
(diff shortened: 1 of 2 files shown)


app/services/clients/connector.client.js
```diff
@@ -423,25 +423,6 @@ ConnectorClient.prototype = {
     return response.data
   },
 
-  /**
-   * @param {string} serviceId
-   * @param {string} accountType
-   * @param {boolean} enabled
-   */
```
DomBelcher, line 430: might as well move the jsdoc to the service methods

DomBelcher (approved):
Nice work

PR #4337, opened 2024-11-15
PP-13297 Add logic to roles and refactor template

james-peacock-gds:
- Add the explanatory content about each role to the roles file and add a method to get the relevant role information for a service. 
- Iterate over the roles in the change permission template to reduce repetition
- The above changes will also make the 'invite team member' page simpler.
- Also, make the radio button labels bold to match the design.

It should be possible to simplify the roles.js file further; I don't think that the `extId` property is actually necessary although it is referred to elsewhere, I will see if this can be refactored in a subsequent PR.

Still to do: Implement the functionality in the post controller for changing permission (and lots of tests)

Screenshot as of 15-11-2024
------
Screenshot 2024-11-15 at 14 31 27

### DIFF
--- a/app/utils/roles.js
+++ b/app/utils/roles.js
@@ -1,17 +1,45 @@
 const _ = require('lodash')
 
 const roles = {
-  admin: { extId: 200, name: 'admin', description: 'Administrator' },
-  'view-and-refund': { extId: 300, name: 'view-and-refund', description: 'View and refund' },
-  'view-only': { extId: 400, name: 'view-only', description: 'View only' },
-  'view-and-initiate-moto': { extId: 500, name: 'view-and-initiate-moto', description: 'View and take telephone payments' },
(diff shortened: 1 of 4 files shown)


app/utils/roles.js
```diff
@@ -20,5 +48,16 @@ module.exports = {
       }
     })
     return found
+  },
+  getAvailableRolesForService: agentInitiatedMotoEnabled => {
+    return Object.values(roles)
+      .filter(role => agentInitiatedMotoEnabled ? true : !role.agentInitiatedMotoServicesOnly)
```
DomBelcher, line 54: I think this logic could be simplified a bit, I think this is equivalent to `agentInitiatedMotoEnabled || !role.agentInitiatedMotoServicesOnly`

DomBelcher (approved):
LGTM

PR #4343, opened 2024-11-20
PP-13297 get users current role

james-peacock-gds:
- Get user's current role and check the appropriate radio button on the change permission form
- Prevent admin from removing self (apply same change to remove users controller, rather than having middleware do this because we need to render a specific error message)
- Ensure that the user is part of the relevant service
- Add unit test for the get controller

To do in subsequent PRs:
- Post controller - Implement the change permission logic 
- Cypress test

Screenshot showing change permission page for a user that is currently 'view-only'.
-----------
Screenshot 2024-11-20 at 12 14 30


app/controllers/simplified-account/settings/team-members/change-permission/change-permission.controller.test.js
```diff
@@ -0,0 +1,91 @@
+const proxyquire = require('proxyquire')
+const sinon = require('sinon')
+const { expect } = require('chai')
+const User = require('../../../../../models/User.class')
+const userFixtures = require('../../../../../../test/fixtures/user.fixtures')
```
DomBelcher, line 5: might as well use the module aliases, `@test` should work now as well

app/middleware/simplified-account/enforce-cannot-do-admin-action-on-self.middleware.js
```diff
@@ -0,0 +1,11 @@
+const { NotFoundError } = require('../../errors')
+
+module.exports = function getAdminActionOnSelfMiddleware (errorMessage) {
```
DomBelcher, line 3: love the use of function currying

app/middleware/simplified-account/enforce-cannot-do-admin-action-on-self.middleware.js
```diff
@@ -0,0 +1,11 @@
...
+  return function adminActionOnSelf (req, res, next) {
+    if (req.params.externalUserId === req.user.externalId) {
+      next(new NotFoundError(errorMessage))
```
DomBelcher, line 6: I think a `403` or `409` makes more sense, if we have errors for either of those

app/controllers/simplified-account/settings/team-members/change-permission/change-permission.controller.test.js
james-peacock-gds: good point, done

app/middleware/simplified-account/enforce-cannot-do-admin-action-on-self.middleware.js
james-peacock-gds: yes, change to NotAuthorisedError

james-peacock-gds: hmm actually I don't think that's appropriate either

james-peacock-gds: To allow a specific error message to be displayed to the user, I've removed the middleware and inserted a conditional in the controller. I've applied the same to the remove user controller and added tests for this.

app/controllers/simplified-account/settings/team-members/change-permission/change-permission.controller.test.js
```diff
@@ -0,0 +1,104 @@
...
+        expect(responseStub.args[0][3]).to.have.property('email').to.equal('user-to-change-permission@users.gov.uk')
+        expect(responseStub.args[0][3]).to.have.property('backLink').to.equal('/simplified/service/service-id-123abc/account/test/settings/team-members')
+      })
+    })
+    describe('failure - admin attempts to change own permissions', () => {
+      before(() => setupTest('get', adminUser, { params: { externalUserId: 'user-id-for-admin-user' } }))
+
+      it('should call the render method with an error', () => {
+        sinon.assert.calledWith(res.status, 403)
+        sinon.assert.calledWith(res.render, 'error', sinon.match({ message: 'You cannot update your own permissions' }))
```
DomBelcher, line 100: is it better to just test that it calls `renderErrorView`? This does call `response` under the hood but it's a bit more unit test-y to just test it calls `renderErrorView`

app/controllers/simplified-account/settings/team-members/remove-user/remove-user.controller.js
```diff
@@ -27,6 +30,11 @@ async function post (req, res, next) {
   const removerExternalId = req.user.externalId
   const externalServiceId = req.service.externalId
   const accountType = req.account.type
+
+  if (removerExternalId === userToRemoveExternalId) {
+    return renderErrorView(req, res, 'You cannot remove yourself from a service', 403)
+  }
```
DomBelcher, line 36: Is this the same conditional as on lines 10-12?

app/controllers/simplified-account/settings/team-members/remove-user/remove-user.controller.test.js
```diff
@@ -87,11 +93,20 @@ describe('Controller: settings/team-members/remove-user', () => {
         expect(responseStub.args[0][3]).to.have.property('backLink').to.equal('/simplified/service/service-id-123abc/account/test/settings/team-members')
       })
     })
+
+    describe('failure - admin attempts to remove self', () => {
+      before(() => setupTest('get', adminUser, { params: { externalUserId: 'user-id-for-admin-user' } }))
+
+      it('should call the render method with an error', () => {
+        sinon.assert.calledWith(res.status, 403)
+        sinon.assert.calledWith(res.render, 'error', sinon.match({ message: 'You cannot remove yourself from a service' }))
```
DomBelcher, line 102: same as above re. asserting on `renderErrorView` vs `response`

app/controllers/simplified-account/settings/team-members/remove-user/remove-user.controller.test.js
```diff
@@ -129,5 +144,19 @@ describe('Controller: settings/team-members/remove-user', () => {
...
+      before(() => setupTest('post', adminUser,
+        {
+          params: { externalUserId: 'user-id-for-admin-user' },
+          body: { email: 'admin-user@users.gov.uk', confirmRemoveUser: 'yes' },
+          flash: sinon.stub()
+        }
+      ))
+      it('should call the render method with an error', () => {
+        sinon.assert.calledWith(res.status, 403)
+        sinon.assert.calledWith(res.render, 'error', sinon.match({ message: 'You cannot remove yourself from a service' }))
```
DomBelcher, line 158: as above

DomBelcher (approved):
Nice work

PR #4349, opened 2024-11-22
PP-13297 change permission layout corrections

james-peacock-gds:
- Refactor the display of roles in the change permissions form to ensure that the ordering of roles matches the design
- Correct spacing between radios and 'save changes' button.

Screenshot for service that doesn't have agent-initiated moto enabled:
-------
Screenshot 2024-11-22 at 15 37 21

Screenshot for service that has agent-initiated moto enabled:
--------
Screenshot 2024-11-22 at 15 38 06

### DIFF
--- a/app/utils/roles.js
+++ b/app/utils/roles.js
@@ -49,11 +49,11 @@ module.exports = {
     })
     return found
   },
-  getAvailableRolesForService: agentInitiatedMotoEnabled => {
+  getAvailableRolesForService: serviceHasAgentInitiatedMotoEnabled => {
     const availableRoles = {}
     for (const roleName in roles) {
       const role = roles[roleName]
-      if (agentInitiatedMotoEnabled) {
+      if (serviceHasAgentInitiatedMotoEnabled) {
         if (roleName === 'admin') {
           // for agent-initiated moto services, add 'take telephone payments' to the explanation content for the admin role
           availableRoles[roleName] = {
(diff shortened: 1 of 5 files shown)


app/utils/roles.js
```diff
@@ -49,15 +49,24 @@ module.exports = {
...
+            explanation: 'They can view transactions, refund payments, take telephone payments and manage settings'
+          }
+        } else {
+          availableRoles[roleName] = { description: role.description, explanation: role.explanation }
+        }
+      } else if (!role.agentInitiatedMotoServicesOnly) {
+        availableRoles[roleName] = { description: role.description, explanation: role.explanation }
+      }
+    }
+    return availableRoles
```
DomBelcher, line 70: Does this do the same thing as the previous code but return it as an object instead of an array? you could map the array into an object by doing 
```
const availableRoles = roleArray.reduce((acc, curr) => Object.assign(acc, { [curr.name]: curr }), {})
```
nothing wrong with this approach though if it does the same thing

DomBelcher: alternatively if this is just so the order can be set for the roles, you could add a `displayOrder` property to the roles and sort on that. 

or you could do

```
const roleDisplayOrder = ['admin', 'view-refund-and-initiate-moto', 'view-and-refund', 'view-and-initiate-moto', 'view-only']

const availableRoles = roleDisplayOrder.map(roleName => roles[roleName])
  .filter(...)
  .map(...)
  .map(...)
```


if you wanted the role display order to be set as an array of the named roles. (this is also faster than sorting but we're only talking about a few roles so speed isn't an issue)

I think this is probably the nicest way since it avoids sorting and also saves doing the ordering in the template

DomBelcher (approved):
LGTM

PR #4352, opened 2024-11-25
PP-13296 invite post controller

james-peacock-gds:
- add the post controller to handle the invite user form, and a unit test for the happy path

To do in subsequent PRs:
- Error page for 412 response from admin for a user who is already a team member, and a corresponding unit test. I'm deferring this as it needs some UCD input.
- Cypress tests
- I've reverted from the ControllerTestBuilder back to the previous test setup but I will revisit this later to see if it's possible to use/adapt ControllerTestBuilder to use in this and other team-members controller tests.

Screenshots for validation errors:
----
Screenshot 2024-11-25 at 17 16 14

Success notification:
-----
Screenshot 2024-11-25 at 17 18 32

### DIFF
--- a/app/controllers/simplified-account/settings/team-members/invite/invite.controller.js
+++ b/app/controllers/simplified-account/settings/team-members/invite/invite.controller.js
@@ -2,21 +2,64 @@ const { response } = require('@utils/response')
 const { getAvailableRolesForService } = require('@utils/roles')
 const formatSimplifiedAccountPathsFor = require('@utils/simplified-account/format/format-simplified-account-paths-for')
 const paths = require('@root/paths')
+const { body, validationResult } = require('express-validator')
+const formatValidationErrors = require('@utils/simplified-account/format/format-validation-errors')
+const userService = require('@services/user.service')
+const lodash = require('lodash')
 
 async function get (req, res) {
   const serviceId = req.service.externalId
   const accountType = req.account.type
-  const serviceHasAgentInitiatedMotoEnabled = req.service.agentInitiatedMotoEnabled ?? false
-  const availableRoles = getAvailableRolesForService(serviceHasAgentInitiatedMotoEnabled)
   response(req, res, 'simplified-account/settings/team-members/invite',
     {
-      availableRoles,
+      availableRoles: getAvailableRolesForService(req.service.agentInitiatedMotoEnabled ?? false),
       backLink: formatSimplifiedAccountPathsFor(paths.simplifiedAccount.settings.teamMembers.index, serviceId, accountType)
     })
 }
 
 async function post (req, res, next) {
-  // TODO
+  const externalServiceId = req.service.externalId
+  const accountType = req.account.type
+  const adminUserExternalId = req.user.externalId
+  const invitedUserEmail = req.body.invitedUserEmail
+  const invitedUserRole = req.body.invitedUserRole
+  const teamMembersIndexPath = formatSimplifiedAccountPathsFor(paths.simplifiedAccount.settings.teamMembers.index, externalServiceId, accountType)
+
+  const validations = [
+    body('invitedUserRole')
+      .not().isEmpty().withMessage('Select a permission level'),
+    body('invitedUserEmail').isEmail().withMessage('Enter a valid email address')
+  ]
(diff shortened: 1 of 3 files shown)


app/controllers/simplified-account/settings/team-members/invite/invite.controller.js
```diff
@@ -2,21 +2,64 @@ const { response } = require('@utils/response')
...
+      backLink: teamMembersIndexPath
+    })
+  }
+
+  try {
+    await userService.createInviteToJoinService(invitedUserEmail, adminUserExternalId, externalServiceId, invitedUserRole)
+    if (lodash.has(req, 'session.pageData.invitee')) {
+      delete req.session.pageData.invitee
+    }
+    // send email
```
DomBelcher, line 56: does sending the email happen on line 52?

app/controllers/simplified-account/settings/team-members/invite/invite.controller.js
```diff
@@ -2,21 +2,64 @@ const { response } = require('@utils/response')
...
+      availableRoles: getAvailableRolesForService(req.service.agentInitiatedMotoEnabled ?? false),
+      backLink: teamMembersIndexPath
+    })
+  }
+
+  try {
+    await userService.createInviteToJoinService(invitedUserEmail, adminUserExternalId, externalServiceId, invitedUserRole)
+    if (lodash.has(req, 'session.pageData.invitee')) {
+      delete req.session.pageData.invitee
+    }
```
DomBelcher, line 55: i'm not sure this would be necessary if the controller just redirects on success, but I may be wrong

app/controllers/simplified-account/settings/team-members/invite/invite.controller.js
```diff
@@ -2,21 +2,64 @@ const { response } = require('@utils/response')
...
+  const accountType = req.account.type
+  const adminUserExternalId = req.user.externalId
+  const invitedUserEmail = req.body.invitedUserEmail
+  const invitedUserRole = req.body.invitedUserRole
+  const teamMembersIndexPath = formatSimplifiedAccountPathsFor(paths.simplifiedAccount.settings.teamMembers.index, externalServiceId, accountType)
+
+  const validations = [
+    body('invitedUserRole')
+      .not().isEmpty().withMessage('Select a permission level'),
+    body('invitedUserEmail').isEmail().withMessage('Enter a valid email address')
```
DomBelcher, line 31: I think it's worth adding a length constraint as well, you can just do `isLength({ max: <length> })` - the database field is a `VARCHAR(255)` so it'll fail on emails longer than that anyway, but we may as well catch it earlier.

You can also use `bail()` like
```
body('invitedUserEmail')
  .isLength({ ... }).withMessage('Enter an email address').bail()
  .isEmail().withMessage('Enter a valid email address')
```

which will mean it only returns the first error if there is no email entered, rather than both.

app/controllers/simplified-account/settings/team-members/invite/invite.controller.js
james-peacock-gds: oops - have removed this.

app/controllers/simplified-account/settings/team-members/invite/invite.controller.js
james-peacock-gds: agreed, the old controller was using session data and I didn't mean to bring this across. Removed.

app/controllers/simplified-account/settings/team-members/invite/invite.controller.js
james-peacock-gds: I've looked into this and the [isEmail validation](https://github.com/validatorjs/validator.js/blob/86911d817fbcd177c295b13d7ca0fdff87341191/src/lib/isEmail.js#L29) already checks for a default max length 254 which is now standard (as well as a maximum for the user part and domain part), so I don't think we need to validate length separately.

DomBelcher: ah very nice, no need to duplicate that then. good find

DomBelcher (approved):
LGTM

PR #4357, opened 2024-11-27
PP-13296 handle user already in team error

james-peacock-gds:
Context: If the invited user is already in the team or has been invited recently by a different user, an error message should be displayed. The content is taken from the existing error message but will be updated if necessary in a subsequent PR.

- if adminusers returns 412, display an error in the same pattern as validation errors.
- unit test for this scenario

For subsequent PRs
- error content update if necessary
- Cypress tests

Screenshot of error page:

Screenshot 2024-11-27 at 13 07 09


DomBelcher (approved):
LGTM

PR #4359, opened 2024-11-27
PP-13349 display year on payout pages

SandorArpa:
## WHAT

Payments into your bank accounts page only display the day & month (DD MON) of the payout date and we only support pagination to go back to previous months or years.

While adding some search boxes involves design and more work, we should at least display the year.

Changed dateformat to `D MMMM YYYY` to match other dateformat on page.

### DIFF
--- a/app/views/payouts/list.njk
+++ b/app/views/payouts/list.njk
@@ -54,8 +54,8 @@ Payments to your bank account - GOV.UK Pay
 {% set transactionsDownloadPath = routes.allServiceTransactions.download if filterLiveAccounts else routes.formattedPathFor(routes.allServiceTransactions.downloadStatusFilter, 'test') %}
 <div class="govuk-grid-column-two-thirds">
   {% for key, group in payoutSearchResult.groups %}
-    {% set dateString = group.date.format('DD MMMM') %}
-    <h2 class="govuk-heading-m">{{ dateString }}</h2>
+    {% set dateString = group.date.format('D MMMM YYYY') %}
+    <h2 class="govuk-heading-m" data-cy="payout-date">{{ dateString }}</h2>
 
     <table class="govuk-table" id="payout-list">
       <thead class="govuk-table__head">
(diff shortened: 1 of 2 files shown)


app/views/payouts/list.njk
```diff
@@ -54,8 +54,8 @@ Payments to your bank account - GOV.UK Pay
 {% set transactionsDownloadPath = routes.allServiceTransactions.download if filterLiveAccounts else routes.formattedPathFor(routes.allServiceTransactions.downloadStatusFilter, 'test') %}
 <div class="govuk-grid-column-two-thirds">
   {% for key, group in payoutSearchResult.groups %}
-    {% set dateString = group.date.format('DD MMMM') %}
-    <h2 class="govuk-heading-m">{{ dateString }}</h2>
+    {% set dateString = group.date.format('D MMMM YYYY') %}
```
alexbishop1, line 57: I believe this is a Moment object and according to https://momentjs.com/docs/#/displaying/ `YYYY` means the year in Moment, unlike [in Java where it means the week year](https://technology.blog.gov.uk/2021/03/24/how-we-fixed-a-strange-week-year-bug/).

Changing `DD` to`D` (no leading zero) is line with the GOV.UK style guide for dates: https://www.gov.uk/guidance/style-guide/a-to-z#dates

test/cypress/integration/payouts/payouts-list.cy.js
```diff
@@ -34,7 +34,7 @@ describe('Payout list page', () => {
 
   it('should correctly display payouts given a successful response from Ledger', () => {
     const payouts = [
-      { gatewayAccountId: liveGatewayAccountId, paidOutDate: '2019-01-29T08:00:00.000000Z' }
+      { gatewayAccountId: liveGatewayAccountId, paidOutDate: '2024-11-27T15:34:00.000000Z' }
```
alexbishop1, line 37: To make sure we’re really outputting the year and not the week year, can we make this 2024-12-31, which is in week year 2025 in both the UK and in the US? (UK because we’re in the UK and US because we it’s a common unintentional default software often gets run in.)

alexbishop1 (changes requested):
Looks good but I think we can make the test more sneaky.

PR #4365, opened 2024-11-29
PP-13313: Worldpay details index page

oswaldquek:
This commit deals with a moto-enabled gateway account that hasn't configured the one_off_customer_initiated credentials yet. We just want to show that there's an outstanding task ('Link your Worldpay account with GOV.UK Pay') here; the link currently doesn't go anywhere.

Screenshot 2024-12-02 at 11 49 21


app/models/WorldpayTasks.class.js
```diff
@@ -0,0 +1,32 @@
...
+
+class WorldpayTasks {
+  /**
+   * @param {GatewayAccount} gatewayAccount
+   */
+  constructor (gatewayAccount) {
+    this.tasks = []
+    this.incompleteTasks = true
+
+    const credential = getActiveCredential(gatewayAccount.rawResponse)
```
nlsteers, line 13: don't rely on the raw response for new code, it's only there as a compatibility fix for old code

instead rewrite/create a new version of `getActiveCredential` to make proper use of the GatewayAccount model add any rawResponse data you need to the relevant class constructor

if you write a new function instead of rewriting the old one, mark the old function as deprecated with jsdoc so we can clean up easier after the simplified account strategy is live

oswaldquek: I was hoping to avoid this but ok.

app/models/WorldpayTasks.class.js
```diff
@@ -0,0 +1,32 @@
+'use strict'
+
+const { getActiveCredential } = require('@utils/credentials')
+
+class WorldpayTasks {
```
DomBelcher, line 5: This probably doesn't need to be a class, if it just produces a list of tasks from a gateway account object it can just be a single function

nlsteers: you could follow the pattern established for stripe tasks and create a utility function: https://github.com/alphagov/pay-selfservice/blob/master/app/utils/simplified-account/settings/stripe-details/tasks.js

oswaldquek: The Stripe tasks are more complex and their state is stored in connector db, so I imagine that's why they'd be handled differently. For me the question to ask when reviewing is the code readable? I think what I've got here currently is easily understandable. We can refactor it to be more like the Stripe tasks pattern if we need be (for example when we come to add more Worldpay tasks). Trying to make this more like Stripe tasks is going to hurt my head at the moment.

app/models/GatewayAccount.class.js
```diff
@@ -34,13 +44,16 @@ class GatewayAccount {
...
+    this.allowMoto = gatewayAccountData.allow_moto
     this.analyticsId = gatewayAccountData.analytics_id
     this.toggle3ds = gatewayAccountData.toggle_3ds
     this.providerSwitchEnabled = gatewayAccountData.provider_switch_enabled
     this.recurringEnabled = gatewayAccountData.recurring_enabled
     if (gatewayAccountData?.gateway_account_credentials) {
       this.gatewayAccountCredentials = gatewayAccountData?.gateway_account_credentials
         .map(credentialData => new GatewayAccountCredential(credentialData))
+
+      this.activeCredential = this.gatewayAccountCredentials.filter((credential) => credential.state === CREDENTIAL_STATE.ACTIVE)[0] || null
```
nlsteers, line 56: don't forget to update the jsdoc

app/controllers/simplified-account/settings/worldpay-details/worldpay-details.controller.test.js
```diff
@@ -0,0 +1,62 @@
...
+
+const ACCOUNT_TYPE = 'live'
+const SERVICE_ID = 'service-id-123abc'
+
+const { req, res, call } = new ControllerTestBuilder('@controllers/simplified-account/settings/worldpay-details/worldpay-details.controller')
+  .withService(new Service({
+    external_id: SERVICE_ID
+  }))
+  .withAccountType(ACCOUNT_TYPE)
+  .withAccount(new GatewayAccount({
```
nlsteers, line 17: this is a good idea, we should probably update the ControllerTestBuilder to create a new GatewayAccount when calling `withAccount` instead of relying on individual tests

app/models/GatewayAccount.class.js
```diff
@@ -1,14 +1,16 @@
...
  * @class GatewayAccount
  * @property {string} name - The name of the gateway account
  * @property {string} id - The id of the gateway account
  * @property {string} type - The type of the gateway account (e.g. test/live)
  * @property {string} description - The description of the gateway account
+ * @property {boolean} allowMoto - whether MOTO payments are enabled on the gateway account
  * @property {string} analyticsId - Google analyticsId of the gateway account
  * @property {boolean} toggle3ds - whether 3DS is enabled or not on this gateway account
  * @property {[GatewayAccountCredential]} gatewayAccountCredentials - available credentials for gateway account
+ * @property {GatewayAccountCredential} activeCredential - the active credential for the gateway account
```
nlsteers, line 13: this is an optional property so should be wrapped in square brackets, see `gatewayAccountCredentials` above

oswaldquek: Are you sure? I thought `[GatewayAccountCredential]` means a list of credentials, which a gateway account does have.

app/views/simplified-account/settings/worldpay-details/index.njk
```diff
@@ -0,0 +1,42 @@
...
+
+{% block settingsContent %}
+  <h1 class="govuk-heading-l">Worldpay details</h1>
+
+  {% if incompleteTasks %}
+    <p class="govuk-body govuk-!-margin-bottom-6">
+      You need to link your Worldpay account to GOV.UK Pay.
+    </p>
+
+    <ul class="govuk-task-list">
```
nlsteers, line 35: we should be using the [task list component](https://design-system.service.gov.uk/components/task-list/) now that DSv5 is live.

See this [PR](https://github.com/alphagov/pay-selfservice/pull/4368/files#diff-161e20a8f11d7d595e5ce3788e032bc62a5d6dcee408f23e067a5c3b101f18e2)

nlsteers (changes requested):
the nunjucks component change is the important thing that needs to be done before this can be approved.

app/models/GatewayAccount.class.js
nlsteers: sorry I meant `@property {GatewayAccountCredential} [activeCredential]`. My B!

app/views/simplified-account/settings/worldpay-details/index.njk
```diff
@@ -0,0 +1,66 @@
...
+      {% if value.complete == true %}
+        Completed
+      {% else %}
+        <strong class="govuk-tag govuk-tag--blue">
+          Not yet started
+        </strong>
+      {% endif %}
+    </div>
+  </li>
+{% endmacro %}
```
nlsteers, line 25:
```suggestion
```

i don't think you need this anymore

nlsteers (approved):

PR #4369, opened 2024-12-03
PP-13348 Uprade AWS SDK to V3

DomBelcher:
## WHAT
 - Upgrade the AWS SDK to V3
 - This uses the new modular format of the SDK, so only the S3 client and S3 Request Presigner are required
 - This also removes the `getDocumentHtmlFromS3` function since it is not used


nlsteers (approved):
🚀

PR #4371, opened 2024-12-04
PP-13313: Validations for Worldpay credentials for a MOTO gateway account

oswaldquek:
Deal with validations. This PR does not deal with checking the creds against the Worldpay API and storing them in connector.
These should match the existing validations in the current [worldpay controller](https://github.com/alphagov/pay-selfservice/blob/master/app/controllers/credentials/worldpay.controller.js).

Screenshot 2024-12-04 at 12 45 49

Screenshot 2024-12-04 at 12 50 00

### DIFF
--- a/app/controllers/simplified-account/settings/worldpay-details/credentials/worldpay-credentials.controller.js
+++ b/app/controllers/simplified-account/settings/worldpay-details/credentials/worldpay-credentials.controller.js
@@ -1,5 +1,5 @@
 const { response } = require('@utils/response')
-const formatSimplifiedAccountPathsFor = require('../../../../../utils/simplified-account/format/format-simplified-account-paths-for')
+const formatSimplifiedAccountPathsFor = require('@utils/simplified-account/format/format-simplified-account-paths-for')
 const paths = require('@root/paths')
 const { body, validationResult } = require('express-validator')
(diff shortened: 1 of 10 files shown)


DomBelcher (commented):
Nice work, just a few small things to fix

app/controllers/simplified-account/settings/worldpay-details/credentials/worldpay-credentials.controller.js
```diff
@@ -0,0 +1,51 @@
...
+      req.service.externalId, req.account.type)
+  })
+}
+
+const worldpayCredentialsValidations = [
+  body('merchantCode').not().isEmpty().withMessage('Enter your merchant code').bail()
+    .custom((value, { req }) => {
+      if (req.account.allowMoto && !value.endsWith('MOTO') && !value.endsWith('MOTOGBP')) {
+        throw new Error('Enter a MOTO merchant code. MOTO payments are enabled for the account')
+      }
```
DomBelcher, line 19: This needs to return `true` if the merchant code is valid, otherwise the validation fails with a generic error

oswaldquek: OK I can add that in

DomBelcher (approved):

PR #4376, opened 2024-12-05
PP-13312 Settings for card types get controller

james-peacock-gds:
- Get controller for simplified settings / card types
- As design stipulates that admin users should see checkboxes and non-admins should see a simple list, implement two methods for formatting the card types data and two child templates to slot into the parent card types page.
- Retain existing logic to disable checkbox for Maestro if the service doesn't have 3ds enabled.
- Update existing logic to provide hint to enable with Worldpay for American Express and UnionPay - this is only necessary if it's a Worldpay service.

For subsequent PRs:
- implement the post controller
- Cypress tests

**Screenshot for admin user**
Screenshot 2024-12-05 at 13 04 39

**Screenshot for non-admin user**
Screenshot 2024-12-05 at 13 04 16

### DIFF
--- a/app/controllers/simplified-account/settings/card-types/card-types.controller.test.js
+++ b/app/controllers/simplified-account/settings/card-types/card-types.controller.test.js
@@ -17,6 +17,18 @@ const adminUser = new User(userFixtures.validUserResponse({
   }
 }))
 
+const viewOnlyUser = new User(userFixtures.validUserResponse(
+  {
+    external_id: 'user-id-for-view-only-user',
+    service_roles: {
+      service:
+        {
+          service: { external_id: SERVICE_ID },
+          role: { name: 'view-only' }
+        }
+    }
+  }))
+
 const allCardTypes = [{
   id: 'id-001',
   brand: 'visa',
@@ -37,10 +49,9 @@ const mockResponse = sinon.spy()
 const mockGetAllCardTypes = sinon.stub().resolves({ card_types: allCardTypes })
 const mockGetAcceptedCardTypesForServiceAndAccountType = sinon.stub().resolves({ card_types: acceptedCardTypes })
 
-const { req, res, call } = new ControllerTestBuilder('@controllers/simplified-account/settings/card-types/card-types.controller')
+const { res, nextRequest, call } = new ControllerTestBuilder('@controllers/simplified-account/settings/card-types/card-types.controller')
   .withServiceExternalId(SERVICE_ID)
   .withAccountType(ACCOUNT_TYPE)
-  .withUser(adminUser)
   .withStubs({
     '@utils/response': { response: mockResponse },
     '@services/card-types.service': {
@@ -51,8 +62,11 @@ const { req, res, call } = new ControllerTestBuilder('@controllers/simplified-ac
   .build()
 
 describe('Controller: settings/card-types', () => {
-  describe('get', () => {
+  describe('get for admin user', () => {
     before(() => {
+      nextRequest({
+        user: adminUser
+      })
       call('get')
     })
 
@@ -61,7 +75,7 @@ describe('Controller: settings/card-types', () => {
     })
 
     it('should pass req, res and template path to the response method', () => {
-      expect(mockResponse.args[0][0]).to.deep.equal(req)
+      expect(mockResponse.args[0][0].user).to.deep.equal(adminUser)
       expect(mockResponse.args[0][1]).to.deep.equal(res)
       expect(mockResponse.args[0][2]).to.equal('simplified-account/settings/card-types/index')
     })
@@ -72,6 +86,33 @@ describe('Controller: settings/card-types', () => {
       expect(mockResponse.args[0][3]).to.have.property('cardTypes').to.have.property('creditCards').length(1)
       expect(mockResponse.args[0][3].cardTypes.creditCards[0]).to.deep.include({ text: 'Visa credit', checked: false })
       expect(mockResponse.args[0][3]).to.have.property('isAdminUser').to.equal(true)
+    })
+  })
+
+  describe('get for non-admin user', () => {
+    before(() => {
+      nextRequest({
+        user: viewOnlyUser
+      })
+      call('get')
+    })
+
+    it('should call the response method', () => {
+      expect(mockResponse.called).to.be.true // eslint-disable-line
+    })
+
+    it('should pass req, res and template path to the response method', () => {
+      expect(mockResponse.args[0][0].user).to.deep.equal(viewOnlyUser)
+      expect(mockResponse.args[0][1]).to.deep.equal(res)
+      expect(mockResponse.args[0][2]).to.equal('simplified-account/settings/card-types/index')
+    })
+
+    it('should pass context data to the response method', () => {
+      expect(mockResponse.args[0][3]).to.have.property('cardTypes').to.have.property('Enabled debit cards').to.have.length(1).to.include('Visa debit')
+      expect(mockResponse.args[0][3].cardTypes).to.have.property('Not enabled debit cards').to.have.length(0)
+      expect(mockResponse.args[0][3].cardTypes).to.have.property('Enabled credit cards').to.have.length(0)
+      expect(mockResponse.args[0][3].cardTypes).to.have.property('Not enabled credit cards').to.have.length(1).to.include('Visa credit')
+      expect(mockResponse.args[0][3]).to.have.property('isAdminUser').to.equal(false)
     })
   })
 })
--- a/app/controllers/simplified-account/settings/card-types/card-types.controller.test.js
+++ b/app/controllers/simplified-account/settings/card-types/card-types.controller.test.js
@@ -1,8 +1,8 @@
 const sinon = require('sinon')
+const ControllerTestBuilder = require('@test/test-helpers/simplified-account/controllers/ControllerTestBuilder.class')
 const { expect } = require('chai')
 const User = require('@models/User.class')
 const userFixtures = require('@test/fixtures/user.fixtures')
-const proxyquire = require('proxyquire')
 
 const ACCOUNT_TYPE = 'live'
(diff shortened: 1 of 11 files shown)


app/utils/simplified-account/format/format-card-types.js
```diff
@@ -0,0 +1,72 @@
...
+      cardTypeChecklistItem.disabled = true
+      cardTypeChecklistItem.hint = {
+        html: account.type === 'test' ? `${cardTypeChecklistItem.text} is not available on test accounts` : `${cardTypeChecklistItem.text} cannot be used because 3D Secure is not available. Please contact support`
+      }
+    }
+    return cardTypeChecklistItem
+  }
+  const addHintForAmexAndUnionpay = (cardTypeChecklistItem) => {
+    if (['American Express', 'Union Pay'].includes(cardTypeChecklistItem.text)) {
+      if (account.paymentProvider === 'worldpay') {
```
DomBelcher, line 29: this could be an `if( A && B )` rather than nested if statements

james-peacock-gds: done

app/utils/simplified-account/format/format-card-types.test.js
```diff
@@ -0,0 +1,111 @@
...
+  },
+  {
+    id: 'id-006',
+    brand: 'maestro',
+    label: 'Maestro',
+    type: 'DEBIT',
+    requires3ds: true
+  }
+]
+describe('format-card-types for template', () => {
```
oswaldquek, line 48: Testing for these [hints](https://github.com/alphagov/pay-selfservice/pull/4376/files#diff-7ab8e2610a16c7fd8d73819249d18f2678d1d4e216024df5d39f519918bf52fbR22-R23) in this test would be good too?

app/utils/simplified-account/format/format-card-types.js
```diff
@@ -0,0 +1,78 @@
...
+    requires3ds: card.requires3ds
+  }
+}
+
+const disableCheckboxIf3dsRequiredButNotEnabled = (cardTypeChecklistItem, accountType, accountRequires3ds) => {
+  if (cardTypeChecklistItem.requires3ds && !accountRequires3ds) {
+    return {
+      ...cardTypeChecklistItem,
+      disabled: true,
+      hint: {
```
oswaldquek, line 22: How are these hints shown in the UI? For example, when the mouse hovers over the label?

app/utils/simplified-account/format/format-card-types.test.js
james-peacock-gds: good point - this is now done.

app/utils/simplified-account/format/format-card-types.js
james-peacock-gds: They are always displayed as additional text underneath the card type if the conditions are met. The text under 'Maestro' in the first screenshot is an example of this.

oswaldquek (approved):

oswaldquek:
Nice 1

app/controllers/simplified-account/settings/card-types/card-types.controller.test.js
```diff
@@ -0,0 +1,131 @@
...
+  return proxyquire('./card-types.controller', {
+    '@utils/response': { response: stubs.response },
+    '@services/card-types.service': {
+      getAllCardTypes: stubs.getAllCardTypes,
+      getAcceptedCardTypesForServiceAndAccountType: stubs.getAcceptedCardTypesForServiceAndAccountType
+    }
+  })
+}
+
+const setupTest = (method, user, additionalReqProps = {}, additionalStubs = {}) => {
```
nlsteers, line 58: can you refactor this test to use the new ControllerTestBuilder?

app/services/clients/connector.client.js
```diff
@@ -257,6 +257,21 @@ ConnectorClient.prototype = {
     return response.data
   },
 
+  /**
+   * Retrieves the accepted card Types for the given account
```
nlsteers, line 261: service external id and account type is a more appropriate description to help differentiate from old gateway account based endpoints

app/utils/simplified-account/format/format-card-types.js
```diff
@@ -0,0 +1,78 @@
...
+        html: accountType === 'test' ? `${cardTypeChecklistItem.text} is not available on test accounts` : `${cardTypeChecklistItem.text} cannot be used because 3D Secure is not available. Please contact support`
+      }
+    }
+  }
+  return cardTypeChecklistItem
+}
+
+const addHintForAmexAndUnionpayIfWorldpay = (cardTypeChecklistItem, paymentProvider) => {
+  if (['American Express', 'Union Pay'].includes(cardTypeChecklistItem.text) && paymentProvider === 'worldpay') {
```
nlsteers, line 31: we should put these in an object somewhere

app/utils/simplified-account/format/format-card-types.test.js
```diff
@@ -0,0 +1,122 @@
...
+  describe('present checkboxes for admin user', () => {
+    it('should return all card types with checked boxes if they are all accepted', () => {
```
nlsteers, line 50: I think we can improve these tests, something like:

```js
const expectedDebitCards = ['Visa debit', 'Mastercard debit', 'Maestro']
const expectedCreditCards = ['Visa credit', 'American Express', 'JCB']

expect(cards.debitCards).to.have.length(expectedDebitCards.length)
expect(cards.creditCards).to.have.length(expectedCreditCards.length)

cards.debitCards.forEach((card, idx) => {
  expect(card).to.deep.include({
    text: expectedDebitCards[idx],
    checked: true
  })
})

cards.creditCards.forEach((card, idx) => {
  expect(card).to.deep.include({
    text: expectedCreditCards[idx],
    checked: true
  })
})
```

nlsteers (changes requested):
few 💬

app/utils/simplified-account/format/format-card-types.js
james-peacock-gds: I've declared the array at the top of the file

app/controllers/simplified-account/settings/card-types/card-types.controller.test.js
james-peacock-gds: yes!

app/utils/simplified-account/format/format-card-types.js
```diff
@@ -0,0 +1,80 @@
...
+    .map(cardTypeChecklistItem => disableCheckboxIf3dsRequiredButNotEnabled(cardTypeChecklistItem, account.type, account.requires3ds))
+
+  const creditCardChecklistItems = allCards.filter(card => card.type === 'CREDIT')
+    .map(card => createCardTypeChecklistItem(card, acceptedCards))
+    .map(cardTypeChecklistItem => addHintForAmexAndUnionpayIfWorldpay(cardTypeChecklistItem, account.paymentProvider))
+
+  return { debitCards: debitCardChecklistItems, creditCards: creditCardChecklistItems }
+}
+
+const formatCardTypesForNonAdminTemplate = (allCards, acceptedCards) => {
```
DomBelcher, line 56: the best I could come up with for refactoring this is something like this, but I'm not convinced it's an improvement so feel free to ignore it and stick with this implementation

```suggestion
const isDisabled = (card, account) => {
  return card.requires3ds && !account.requires3ds
}

const hintText = (card, account) => {
  const cardLabel = formatLabel(card)
  const disabled = isDisabled(card, account)
  if (disabled && account.type === 'test') {
    return {
      text: `${cardLabel} is not available on test accounts`
    }
  } else if (disabled) {
    return {
      text: `${cardLabel} cannot be used because 3D Secure is not available. Please contact support`
    }
  }

  if (card.type !== 'CREDIT' || account.paymentProvider !== 'worldpay') {
    return undefined
  }

  if (['American Express', 'Union Pay'].includes(cardLabel)) {
    return {
      html: 'You must have already enabled this with Worldpay'
    }
  }

  return undefined
}

const formatCard = (card, accepted, account) => {
  const disabled = isDisabled(card, account)
  return {
    value: card.id,
    text: formatLabel(card),
    checked: !disabled && accepted,
    requires3ds: card.requires3ds,
    disabled,
    hint: hintText(card, account)
  }
}

const formatCardTypesForNonAdminTemplate = (allCards, acceptedCards) => {
  const items = allCards.reduce((acc, card) => {
    if (!acc[card.type]) {
      acc[card.type] = []
    }
    const accepted = acceptedCards.filter(c => c.type === card.type && c.id === card.id).length > 0
    acc[card.type].push(formatCard(card, accepted, account))

    return acc
  }, {})

  return {
    debitCards: Object.values(items.DEBIT),
    creditCards: Object.values(items.CREDIT)
  }
}
```

app/utils/simplified-account/format/format-card-types.js
```diff
@@ -0,0 +1,80 @@
...
+    'Not enabled debit cards': [],
+    'Enabled credit cards': [],
+    'Not enabled credit cards': []
+  }
+  allCards.forEach(card => {
+    const cardIsEnabled = acceptedCardTypeIds.includes(card.id) ? 'Enabled' : 'Not enabled'
+    formattedCardTypes[`${cardIsEnabled} ${card.type.toLowerCase()} cards`].push(formatLabel(card))
+  })
+  return formattedCardTypes
+}
```
DomBelcher, line 69: a slight change I might suggest would be something like

```suggestion
const formattedCardTypes = {
  'debit/enabled': {
     cards: [],
     heading: 'Enabled debit cards'
  }
  ...
}
allCards.forEach(card => {
  const cardIsEnabled = acceptedCardTypeIds.includes(card.id) ? 'enabled' : 'disabled'
  formattedCardTypes[`${card.type.toLowerCase()}/${cardIsEnabled}`].push(formatLabel(card))
})

```

just so the heading isn't the object key

nlsteers (approved):

PR #4381, opened 2024-12-11
BAU Refactor team-members controller test

james-peacock-gds:
- Refactor team members controller test to use the ControllerTestBuilder.

### DIFF
--- a/app/controllers/simplified-account/settings/team-members/team-members.controller.test.js
+++ b/app/controllers/simplified-account/settings/team-members/team-members.controller.test.js
@@ -83,6 +83,47 @@ describe('Controller: settings/team-members', () => {
       expect(mockResponse.args[0][3]).to.have.property('inviteTeamMemberLink')
         .to.equal('/simplified/service/service-id-123abc/account/test/settings/team-members/invite')
       expect(mockResponse.args[0][3]).to.have.property('numberInvitedMembers').to.equal(2)
+      expect(mockResponse.args[0][3]).to.have.property('isServiceAdmin').to.be.true // eslint-disable-line no-unused-expressions
+      expect(mockResponse.args[0][3]).to.have.property('roles').to.have.keys('admin', 'view-and-initiate-moto', 'view-and-refund', 'view-only', 'view-refund-and-initiate-moto')
+    })
+  })
+
+  describe('get', () => {
+    before(() => {
+      nextRequest({
+        user: adminUser
+      })
+      call('get')
+    })
+
+    it('should call the response method', () => {
+      expect(mockGetServiceUsers.called).to.be.true // eslint-disable-line
+      expect(mockResponse.called).to.be.true // eslint-disable-line
+      expect(mockResponse.called).to.be.true // eslint-disable-line
+    })
+
+    it('should pass req, res and template path to the response method', () => {
+      expect(mockResponse.args[0][0].user).to.deep.include(adminUser)
+      expect(mockResponse.args[0]).to.include(res)
+      expect(mockResponse.args[0]).to.include('simplified-account/settings/team-members/index')
+    })
+
+    it('should pass context data to the response method', () => {
+      expect(mockResponse.args[0][3]).to.have.property('teamMembers').to.have.property('admin').to.have.length(1)
+      expect(mockResponse.args[0][3].teamMembers.admin[0]).to.have.property('externalId').to.equal('user-id-for-admin-user')
+      expect(mockResponse.args[0][3].teamMembers).to.have.property('view-only').to.have.length(1)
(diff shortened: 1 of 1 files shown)


app/controllers/simplified-account/settings/team-members/team-members.controller.test.js
```diff
@@ -1,104 +1,88 @@
...
-    '@services/user.service':
-      { getServiceUsers: stubs.getServiceUsers, getInvitedUsers: stubs.getInvitedUsers }
-  })
-}
-
-const setupTest = (method, additionalReqProps = {}, additionalStubs = {}) => {
-  responseStub = sinon.spy()
-  const adminUser = new User(userFixtures.validUserResponse({
-    external_id: 'user-id-for-admin-user',
+const adminUser = new User(userFixtures.validUserResponse({
```
oswaldquek, line 10: Why are is the User being initialised with `userFixtures.validUserResponse` rather than the [userData](https://github.com/alphagov/pay-selfservice/blob/master/app/models/User.class.js#L38)?

app/controllers/simplified-account/settings/team-members/team-members.controller.test.js
```diff
@@ -1,104 +1,88 @@
...
+      expect(mockGetServiceUsers.called).to.be.true // eslint-disable-line
+      expect(mockResponse.called).to.be.true // eslint-disable-line
+      expect(mockResponse.called).to.be.true // eslint-disable-line
     })
 
     it('should pass req, res and template path to the response method', () => {
-      expect(responseStub.args[0]).to.include(req)
-      expect(responseStub.args[0]).to.include(res)
-      expect(responseStub.args[0]).to.include('simplified-account/settings/team-members/index')
+      expect(mockResponse.args[0][0].user).to.deep.include(adminUser)
```
oswaldquek, line 67: Don't we want to keep the testing of the `req` here? 
```
expect(mockResponse.args[0]).to.include(req)
```

app/controllers/simplified-account/settings/team-members/team-members.controller.test.js
```diff
@@ -1,104 +1,88 @@
...
+      expect(mockResponse.args[0][3]).to.have.property('invitedTeamMembers').to.have.property('admin').to.have.length(1)
+      expect(mockResponse.args[0][3].invitedTeamMembers.admin[0]).to.have.property('email').to.equal('invited-admin-user@user.gov.uk')
+      expect(mockResponse.args[0][3].invitedTeamMembers).to.have.property('view-only').to.have.length(1)
+      expect(mockResponse.args[0][3].invitedTeamMembers['view-only'][0]).to.have.property('email').to.equal('invited-view-only-user@user.gov.uk')
 
-      expect(responseStub.args[0][3]).to.have.property('inviteTeamMemberLink')
+      expect(mockResponse.args[0][3]).to.have.property('inviteTeamMemberLink')
         .to.equal('/simplified/service/service-id-123abc/account/test/settings/team-members/invite')
-      expect(responseStub.args[0][3]).to.have.property('numberInvitedMembers').to.equal(2)
+      expect(mockResponse.args[0][3]).to.have.property('numberInvitedMembers').to.equal(2)
```
oswaldquek, line 85: Should we test for the properties `isServiceAdmin` and `roles` here too?

app/controllers/simplified-account/settings/team-members/team-members.controller.test.js
james-peacock-gds: The existing user fixture provides some default values so it seemed sensible to reuse that rather than create a new user from scratch.

app/controllers/simplified-account/settings/team-members/team-members.controller.test.js
james-peacock-gds: I think the use of `nextRequest` means that we won't expect req to remain the same. Looking at the other uses of the ControllerTestBuilder when they use `nextRequest`, they are only asserting on the part that we have added

app/controllers/simplified-account/settings/team-members/team-members.controller.test.js
james-peacock-gds: good point, done.

app/controllers/simplified-account/settings/team-members/team-members.controller.test.js
james-peacock-gds: ah - but I see what you mean, there are some default values in userData too. I did try using this but hit a problem where it didn't recognise the service roles. I can go back and work through it if you think it's better using userData.

oswaldquek: I think is okay for now, as it's used in loads of places. No need to change.

oswaldquek (approved):

oswaldquek:
Can the title of the PR be changed from Bau to a JIRA ticket please.

PR #4387, opened 2024-12-16
PP-13335 PP-13335 Transactions details page - display corporate exemptions data.

iqbalgds:
- Updating the transaction details controller to append the following
  - Add 3DS secure information about the transaction.
  - Add information about corporate exemptions.
    - If corporate exemptions are disabled on the gateway, then we only show information older transactions when a corporate exemption has been requested.
    - If corporate exemptions are enabled, then it will show information at all times about the transaction.
      - It will show `not requested`, if a corporate exemption was not not requested for the current transaction.
- This PR is only for updating the controllers.
- A future PR will add the fields to the Nunjucks file to display the fields correctly.

image


SandorArpa (approved):
LGTM

PR #4389, opened 2024-12-16
PP-12416: Card Payment page

nataliecarey:
## WHAT
Added the Card Payment page, also included:

 - basic logger when env var `GOVUK_PAY__USE_BASIC_LOGGER` is set to `true`, then set that to `true` on Cypress tests
 - as I introduced a couple of new filters I also implemented them in other templates where the same functionality was used
 - I noticed a common issue with the use of visually hidden text in Summary Lists and corrected it everywhere I could see it

## HOW 

I added a couple of Nunjucks filters:

 - `boolToText('a value for true', 'a value for false')`
 - `boolToOnOrOff` which behaves the same but with `On` and `Off` specified to avoid repetition


app/utils/logger.js
```diff
@@ -59,6 +59,9 @@ const nsDebugLogger = createLogger({
 const nsDebug = process.env.NS_DEBUG === 'true'
 
 module.exports = (loggerName) => {
+  if (process.env.GOVUK_PAY__USE_BASIC_LOGGER === 'true') {
```
nlsteers, line 62: you can achieve a simple logger without adding another env var by just toggling `NS_DEBUG` (which probably shouldn't have been merged 🫠 ). 

If that's still too much detail you can comment out the `metadataStr` in `nsDebugFormat` above to have just single log lines (with the added benefit of colour for info, warn, error etc)

app/views/simplified-account/settings/card-payments/index.njk
```diff
@@ -0,0 +1,100 @@
...
+        text: collectBillingAddressEnabled | boolToOnOrOff
+      },
+      actions: {
+        items: [
+          {
+            href: collectBillingAddressLink,
+            text: "Change",
+            classes: "govuk-link--no-visited-state",
+            visuallyHiddenText: "collect billing address"
+          }
```
DomBelcher, line 30: these can be conditionally rendered if the change link is provided by the controller

test/cypress/integration/simplified-account/service-settings/card-payments/card-payments.cy.js
```diff
@@ -0,0 +1,157 @@
...
+    describe('for an admin user', () => {
+      it('should show the correct heading and title', () => {
+        setupStubs()
+        cy.visit(pageUrl)
+
+        cy.get('h1').should('contain.text', 'Card payments')
+        cy.title().should('eq', 'Settings - Card payments - GOV.UK Pay')
+      })
+
+      it('should display the provided card payment details (version 1 - everything on)', () => {
```
DomBelcher, line 55: Can this be in a `describe('when all settings are on')`? similarly for when all settings are off

test/cypress/integration/simplified-account/service-settings/card-payments/card-payments.cy.js
```diff
@@ -0,0 +1,157 @@
...
+      }
+      cy.get('.govuk-summary-list').eq(summaryListIndex).within(() => {
+        cy.get('.govuk-summary-list__row').eq(currentRowIndex).within(() => {
+          cy.get('.govuk-summary-list__key').should('contain', currentRowExpectations.key)
+          cy.get('.govuk-summary-list__value').should('contain', currentRowExpectations.value)
+        })
+      })
+    })
+  })
+}
```
DomBelcher, line 157: I think this is overly DRY and inlining the assertions would be clearer

nataliecarey: The reason I did this was for readability within the test itself:

```
        assertSummaryListsEqual([
          [
            {
              key: 'Collect billing address',
              value: 'On'
            },
            {
              key: 'Default billing address country',
              value: 'United Kingdom'
            }
          ],
          [

            {
              key: 'Apple Pay',
              value: 'On'
            },
            {
              key: 'Google Pay',
              value: 'On'
            }
          ]
        ])
      })
```

What I was trying to avoid was two blocks that look like this as the important detail is hidden in the numbers inside `eq()`:

```
        cy.get('.govuk-summary-list').within(() => {
          cy.get('.govuk-summary-list__key').eq(0).should('contain', 'Ask users for their email address')
          cy.get('.govuk-summary-list__value').eq(0).should('contain', 'On')
          cy.get('.govuk-summary-list__actions a').eq(0).should('not.contain', 'Change')
          cy.get('.govuk-summary-list__key').eq(1).should('contain', 'Send payment confirmation emails')
          cy.get('.govuk-summary-list__value').eq(1).should('contain', 'On')
          cy.get('.govuk-summary-list__actions a').eq(1).should('not.contain', 'Change')
          cy.get('.govuk-summary-list__key').eq(2).should('contain', 'Send refund emails')
          cy.get('.govuk-summary-list__value').eq(2).should('contain', 'On')
          cy.get('.govuk-summary-list__actions a').eq(2).should('not.contain', 'Change')
        })
```

app/utils/logger.js
nataliecarey: I just tried that, the log level display is nice but I'm getting a whole load of extra lines specifying the `logger_name`:

```
2024-12-17 12:02:12 [info]: Page viewed with account simplification turned on
{
  "logger_name": "/Users/natalie.carey/projects/alphagov/pay-selfservice/app/models/User.class.js"
}
```

Given the choice I would still go with the *very* basic option, but I could work with your approach.

I think we should standardise to one or the other and I'd suggest the `GOVUK_PAY__USE_BASIC_LOGGER` name because people can set that globally or project-by-project without risking conflicts and understanding what they're getting.

test/cypress/integration/simplified-account/service-settings/card-payments/card-payments.cy.js
```diff
@@ -0,0 +1,157 @@
...
+    //     })
+    //   })
+    //
+    //   it('should return a 403', () => {
+    //     cy.request({
+    //       url: pageUrl,
+    //       failOnStatusCode: false
+    //     }).then(response => expect(response.status).to.eq(403))
+    //   })
+    // })
```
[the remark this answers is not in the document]
nataliecarey: I was conflicted about whether to include this - it documents the current behaviour but it's not the intended behaviour.  I think I'll uncomment it in this commit and replace it in the next piece of work on the ticket.

app/simplified-account-routes.js
```diff
@@ -61,6 +61,9 @@ simplifiedAccount.post(paths.simplifiedAccount.settings.organisationDetails.edit
 // card types
 simplifiedAccount.get(paths.simplifiedAccount.settings.cardTypes.index, permission('transactions:read'), serviceSettingsController.cardTypes.get)
 
+// card payments
+simplifiedAccount.get(paths.simplifiedAccount.settings.cardPayments.index, permission('payment-types:update'), serviceSettingsController.cardPayments.get)
```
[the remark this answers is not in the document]
nataliecarey: Fair.

app/views/settings/index.njk
```diff
@@ -26,7 +26,7 @@
               classes: 'govuk-!-width-one-half'
             },
             value: {
-              text: "On" if currentGatewayAccount.allow_apple_pay else "Off",
+              text: currentGatewayAccount.allow_apple_pay | boolToOnOrOff,
```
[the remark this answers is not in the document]
nataliecarey: I'm happy to go with whatever the team consensus is on this.

test/cypress/integration/simplified-account/service-settings/card-payments/card-payments.cy.js
nataliecarey: Yeah, I kept switching between different options, the problem is that one of them is a country code so 'everything on' isn't a great description but when I was getting confused between the tests I added it to avoid confusion.  I have considered using `should display the default details` and `should display user-set details` but that implies that the test would be updated in line with any changes to the defaults and I doubt that would happen.

test/cypress/integration/simplified-account/service-settings/card-payments/card-payments.cy.js
DomBelcher: I'm not sure I understand what you mean, changing the numbers in `eq()` from being explicitly stated to being an array index doesn't (to my mind) elucidate any detail that would otherwise be hidden. In the example block above a few line breaks could improve the readability, along with perhaps the use of e.g. `.eq(0).within(() => {})`, which admittedly is less dry but in the event of a regression would make it very easy to identify where the regression occured with only a line number for the failed assertion.

app/utils/logger.js
nlsteers: yeah see the second part of my comment re `metadataStr` if you want it to look like the above screenshot

test/cypress/integration/simplified-account/service-settings/card-payments/card-payments.cy.js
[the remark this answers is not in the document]
DomBelcher: Given that the page should be viewable for a non-admin user, but without the change links and with the "You don't have permission to manage settings..." inset text, there should be a cypress test for this view, but that's fine to come in a future PR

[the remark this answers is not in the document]
DomBelcher: For now I think it's best to delete the test, the describe block can just be uncommented as it won't do anything without a test to run

app/controllers/simplified-account/settings/card-payments/card-payments.controller.js
```diff
@@ -0,0 +1,32 @@
...
+  const country = req.service.defaultBillingAddressCountry === GB_COUNTRY_CODE ? 'United Kingdom' : req.service.defaultBillingAddressCountry
+  const raw = req.account?.rawResponse
+  const applePay = raw?.allow_apple_pay
+  const googlePay = raw?.allow_google_pay
+
+  response(req, res, 'simplified-account/settings/card-payments/index', {
+    collectBillingAddressEnabled: billing,
+    collectBillingAddressLink: formatSimplifiedAccountPathsFor(cardPaymentsPaths.collectBillingAddress, serviceExternalId, accountType),
+    defaultBillingAddressCountry: country,
+    defaultBillingAddressCountryLink: formatSimplifiedAccountPathsFor(cardPaymentsPaths.defaultBillingAddressCountry, serviceExternalId, accountType),
```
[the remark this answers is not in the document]
nataliecarey: I'm going to tackle this in a later PR.  There are a few parts to the access management and I'm trying to do the multiple small pull requests thing.

app/views/simplified-account/settings/service-name/index.njk
```diff
@@ -27,7 +27,7 @@
             href: manageEn,
             text: "Change",
             classes: "govuk-link--no-visited-state",
-            visuallyHiddenText: "Edit English service name",
+            visuallyHiddenText: "English service name",
```
oswaldquek, line 30: Why is the "Edit" removed? I would've thought a verb would be useful for accessible content.

app/views/simplified-account/settings/stripe-details/responsible-person/check-your-answers.njk
```diff
@@ -22,7 +22,7 @@
           {
             href: changeResponsiblePersonLink,
             text: "Change",
-            visuallyHiddenText: "change name",
+            visuallyHiddenText: "name",
```
oswaldquek, line 25: Similar to the above "Edit" comment

test/cypress/integration/simplified-account/service-settings/card-payments/card-payments.cy.js
```diff
@@ -0,0 +1,119 @@
...
+
+        cy.visit(pageUrl)
+        cy.get('.govuk-summary-list').eq(0).within(() => {
+          cy.get('.govuk-summary-list__row').eq(0).within(() => {
+            cy.get('.govuk-summary-list__key').should('contain', 'Collect billing address')
+            cy.get('.govuk-summary-list__value').should('contain', 'Off')
+          })
+          cy.get('.govuk-summary-list__row').eq(1).within(() => {
+            cy.get('.govuk-summary-list__key').should('contain', 'Default billing address country')
+            cy.get('.govuk-summary-list__value').should('contain', 'IE')
```
oswaldquek, line 103: Eww why is "IE" (Ireland?) here but "United Kingdom" in version 1 of the test?

test/cypress/integration/simplified-account/service-settings/card-payments/card-payments.cy.js
```diff
@@ -0,0 +1,119 @@
...
+            cy.get('.govuk-summary-list__value').should('contain', 'On')
+          })
+          cy.get('.govuk-summary-list__row').eq(1).within(() => {
+            cy.get('.govuk-summary-list__key').should('contain', 'Google Pay')
+            cy.get('.govuk-summary-list__value').should('contain', 'On')
+          })
+        })
+      })
+
+      it('should display the provided card payment details (version 2 -everything off)', () => {
```
oswaldquek, line 86: ` - everything`

test/cypress/integration/simplified-account/service-settings/card-payments/card-payments.cy.js
nataliecarey: I took it from [the existing implementation](https://github.com/alphagov/pay-selfservice/blob/8d90726/app/views/settings/index.njk#L128) (because the ticket talks about the page contents not changing).  The existing behaviour is that 'GB' is replaced with 'United Kingdom' but everything else is left unedited.

I think this is worth looking into as `GB` and `UK` are not synonymous but we're treating them as though they are.

test/cypress/integration/simplified-account/service-settings/card-payments/card-payments.cy.js
nataliecarey: Fixed.

app/views/simplified-account/settings/service-name/index.njk
nataliecarey: I put a bullet point about this in the PR description, but it's a bit buried.  "I noticed a common issue with the use of visually hidden text in Summary Lists and corrected it everywhere I could see it".

Basically the old version outputs:

```
Edit <span class="govuk-visually-hidden">Edit English service name</span>
```

Which would read as "Edit Edit English service name", it's not clear from the variable names but you can see it in the examples on the Design System https://design-system.service.gov.uk/components/summary-list/ (you can compare the Nunjucks input to the HTML output).

app/views/simplified-account/settings/stripe-details/responsible-person/check-your-answers.njk
nataliecarey: As above - this will read as "Change change name" on a screen reader.

oswaldquek (approved):
Nice 1️⃣

PR #4397, opened 2024-12-30
PP-12395: Validation when creating API key

oswaldquek:
A description is not valid if it's empty or more than 50 chars. Given the text box has a maxLength of 50, the latter validation error can't occur in practice.

Screenshot 2024-12-30 at 12 04 32


james-peacock-gds (approved):
LGTM

PR #4399, opened 2024-12-31
PP-12395: Change api key name

oswaldquek:
Changing the name uses the same view as the "API key name" view in https://github.com/alphagov/pay-selfservice/pull/4394

### DIFF
--- a/app/controllers/simplified-account/settings/api-keys/change-name/change-name.controller.js
+++ b/app/controllers/simplified-account/settings/api-keys/change-name/change-name.controller.js
@@ -1,10 +1,10 @@
-const { response } = require('@utils/response')
-const formatSimplifiedAccountPathsFor = require('@utils/simplified-account/format/format-simplified-account-paths-for')
 const paths = require('@root/paths')
+const { validationResult } = require('express-validator')
+const formatSimplifiedAccountPathsFor = require('@utils/simplified-account/format/format-simplified-account-paths-for')
+const formatValidationErrors = require('@utils/simplified-account/format/format-validation-errors')
+const { response } = require('@utils/response')
 const { changeApiKeyName } = require('@services/api-keys.service')
 const DESCRIPTION_VALIDATION = require('@controllers/simplified-account/settings/api-keys/validations')
-const { validationResult } = require('express-validator')
-const formatValidationErrors = require('@utils/simplified-account/format/format-validation-errors')
 
 function get (req, res) {
   return response(req, res, 'simplified-account/settings/api-keys/api-key-name', {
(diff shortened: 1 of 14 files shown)


app/controllers/simplified-account/settings/api-keys/change-name/change-name.controller.js
```diff
@@ -0,0 +1,36 @@
+const { response } = require('@utils/response')
+const formatSimplifiedAccountPathsFor = require('@utils/simplified-account/format/format-simplified-account-paths-for')
+const paths = require('@root/paths')
+const { changeApiKeyName } = require('@services/api-keys.service')
+const DESCRIPTION_VALIDATION = require('@controllers/simplified-account/settings/api-keys/validations')
+const { validationResult } = require('express-validator')
+const formatValidationErrors = require('@utils/simplified-account/format/format-validation-errors')
```
james-peacock-gds, line 7: I think there is a convention to order these require statements to improve readability, with core modules first and destructured imports last.

james-peacock-gds (changes requested):
I've added one comment about the order of require statements.
Also:
- I assume there will be a cypress test for the change key name in a subsequent PR?
- The Figma says that the 'Add description for the key' label should no longer be shown on this page.

test/cypress/integration/simplified-account/service-settings/api-keys/api-keys.cy.js
```diff
@@ -166,6 +167,45 @@ describe('Settings - API keys', () => {
...
+      it('should re-name the api key successfully', () => {
+        cy.get('.govuk-summary-card').within(() => {
+          cy.contains('h2', 'mathematical clothes').should('exist')
+          cy.contains('a', 'Change name').click()
+        })
+        cy.get('input[id="description"]').type(NEW_API_KEY_NAME)
+        cy.contains('button', 'Continue').click()
+        cy.contains('h1', 'Test API keys').should('exist')
+        cy.get('.govuk-summary-card').within(() => {
+          cy.contains('h2', 'mathematical clothes').should('exist')
```
james-peacock-gds, line 205: shouldn't this change to the new name?

james-peacock-gds (changes requested):

test/cypress/integration/simplified-account/service-settings/api-keys/api-keys.cy.js
oswaldquek: Ohh you're right. I'm not sure this test line is relevant because we're stubbing out the response [here](https://github.com/alphagov/pay-selfservice/pull/4399/files#diff-62c503bb6b4cf2b9a5a9e018275ec1634f9a980d7226e4fd240fa8313c21aaadR27). I remember mentioning in one of your PRs cypress testing isn't testing the database layer. I think we just need to know upon changing the name, the controller redirects to the index page so I think I can get rid of this line.

james-peacock-gds (approved):
LGTM

PR #4401, opened 2025-01-06
PP-13452 remove nunjucksify dependency from build toolchain

nlsteers:
## WHAT

build simplification as part of wider migration to a more modern toolchain

- bundle nunjucks in clientside code and inline multiselect template


app/browsered/multi-select.js
```diff
@@ -43,7 +67,7 @@ function progressivelyEnhanceSelects () {
         return { value, id, checked, text }
       })
     }
-    select.outerHTML = multiSelect(configuration)
+    select.outerHTML = nunjucks.renderString(multiSelectTemplate, configuration)
```
DomBelcher, line 70: Could you not just use `nunjucks.render(<template path>)` to render the template file, rather than having to inline it? not particularly important but nicer to not inline it if possible i think

DomBelcher: I realise this won't work as it's client side, my bad

nataliecarey: If this is the only usage and we're continuing to look into how we want to handle the build (I believe both are true) then I think this is a reasonable change.

As this seems to be the only case I wonder whether we should consider making an endpoint which renders this template and returns partial HTML to be injected - that way we can more sensibly use things like Design System Components without the complexity spiralling.

PR #4409, opened 2025-01-15
BAU add cypress dev server config

nlsteers:
### WHAT

- add npm run command to start cypress server with inline source maps and hot reload enabled
- update `dev-server.mjs` to load cypress test .env
- update README with usage instructions


DomBelcher (approved):
LGTM

PR #4411, opened 2025-01-16
PP-13313 Add Cypress tests for Worldpay credentials form

DomBelcher:
### WHAT
 - Cypress tests for Worldpay credentials form in Worldpay details settings


test/cypress/integration/simplified-account/service-settings/worldpay-details/worldpay-details.cy.js
```diff
@@ -180,4 +189,151 @@ describe('Worldpay details settings', () => {
...
+    })
+
+    describe('for a non-Worldpay account', () => {
+      beforeEach(() => {
+        setupStubs({
+          role: 'view-and-refund',
+          paymentProvider: STRIPE
+        })
+      })
+
```
nlsteers, line 330: might want to test that the nav option doesn't show up for non worldpay accounts

nlsteers (dismissed):
one suggestion, can be added in another PR so 🚀

nlsteers (approved):

PR #4414, opened 2025-01-17
BAU Update stubs in Cypress tests to be more robust

DomBelcher:
### What
- Update stubs in Cypress tests to use full queries & request bodies where possible, use `deepMatchRequest: false` where not possible
- Fix a few bugs in Axios clients with logging metadata incorrectly being sent in requests

### DIFF
--- a/src/services/clients/ledger.client.js
+++ b/src/services/clients/ledger.client.js
@@ -80,15 +80,14 @@ const transactions = async function transactions (gatewayAccountIds = [], filter
     url,
     'List transactions for a given gateway account ID',
     {
-      data: {
+      loggingMetadata: {
         gateway_account_ids: gatewayAccountIds,
         multiple_accounts: gatewayAccountIds.length > 1,
         filters: Object.keys(filters).sort().join(', ')
       }
     }
   )
-  const body = legacyConnectorTransactionsParity(response.data)
-  return body
+  return legacyConnectorTransactionsParity(response.data)
 }
 
 const transactionSummary = async function transactionSummary (gatewayAccountId, fromDate, toDate, options = {}) {
--- a/test/cypress/stubs/gateway-account-stubs.js
+++ b/test/cypress/stubs/gateway-account-stubs.js
@@ -196,15 +196,17 @@ function patchAccountEmailCollectionModeSuccess (opts) {
 function patchAccountEmailCollectionModeSuccessByServiceIdAndAccountType (serviceId, accountType, emailCollectionMode = 'MANDATORY') {
   const path = `/v1/api/service/${serviceId}/account/${accountType}`
   return stubBuilder('PATCH', path, 200,
-    { op: 'replace', path: 'email_collection_mode', value: emailCollectionMode })
+    { request: { op: 'replace', path: 'email_collection_mode', value: emailCollectionMode } })
 }
 
 function setRefundEmailEnabledByServiceIdAndAccountType (serviceId, accountType, enabled) {
   const path = `/v1/api/service/${serviceId}/account/${accountType}/email-notification`
   const payload = {
-    op: 'replace',
-    path: '/refund_issued/enabled',
-    value: enabled
+    request: {
+      op: 'replace',
+      path: '/refund_issued/enabled',
+      value: enabled
+    }
   }
   return stubBuilder('PATCH', path, 200, payload)
 }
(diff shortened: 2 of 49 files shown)


test/cypress/stubs/gateway-account-stubs.js
```diff
@@ -304,9 +313,19 @@ function postUpdateWorldpay3dsFlexCredentials (opts) {
...
+    request: [{
+      op: 'replace',
+      path: patchOpts.path,
+      value: patchOpts.value
+    }, {
+      op: 'replace',
+      path: 'last_updated_by_user_external_id',
+      value: patchOpts.userExternalId
+    }]
+  })
```
oswaldquek, line 328: I think this method signature has potential to be confusing. I'd prefer if a list of patchOpts were passed in, for e.g.
```
patchUpdateCredentialsSuccess(1, 123bc, [{path: 'my-path', value: 'my-value'}, {path: 'last_updated_by_user_external_id', value: 'a-user-external-id'}]

File: src/services/clients/ledger.client.js [hunk 2]
```diff
@@ -80,15 +80,14 @@ const transactions = async function transactions (gatewayAccountIds = [], filter
     url,
     'List transactions for a given gateway account ID',
     {
-      data: {
+      loggingMetadata: {
```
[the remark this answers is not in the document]
DomBelcher: This key being `data` means the value is sent as the request body, as if it were a POST/PUT/PATCH request, however for a GET method there should be no request body. Prior to the axios migration this was called `additionalLoggingFields`, why it was renamed `data` I don't know, but it seems to have simply been a mistake

test/cypress/stubs/gateway-account-stubs.js
DomBelcher: Yeah this may not be perfect, perhaps `userExternalId` should instead be a separate parameter to pass in to the method instead. The `last_updated_by_user_external_id` replace op should always performed on this PATCH endpoint, so it probably doesn't need to be configured each time the stub is used, rather configured by default

oswaldquek: Yeah I think `userExternalId` should be a separate parameter. One annoying thing I find about method signatures with just `(opts)` ([example](https://github.com/alphagov/pay-selfservice/blob/master/test/cypress/stubs/agreement-stubs.js#L36) is that when calling the method from another file I have to look into the method implementation to find out what fields are available on the `opts`; this is something I'd like to see less of.

oswaldquek: `udpatedByUserExternalId` would be clearer than just `userExternalId` even.

DomBelcher: I do agree about the use of `opts`, it's very opaque about what's required. I've used `patchOpts` in a few places so I could add a JSdoc for it to make it more transparent

test/cypress/stubs/gateway-account-stubs.js
```diff
@@ -196,35 +198,41 @@ function patchAccountEmailCollectionModeSuccess (opts) {
...
   const path = `/v1/api/service/${serviceId}/account/${accountType}/email-notification`
   const payload = {
-    op: 'replace',
-    path: '/refund_issued/enabled',
-    value: enabled
+    request: {
+      op: 'replace',
+      path: '/refund_issued/enabled',
+      value: enabled
+    }
```
oswaldquek, line 211: Interesting how this works. According to the spec the body is
```
{
  "op": "replace",
  "path": "/payment_confirmed/enabled",
  "value": false
}
```
rather than 
```
request: 
{
  "op": "replace",
  "path": "/payment_confirmed/enabled",
  "value": false
}
```

test/cypress/stubs/gateway-account-stubs.js
oswaldquek: Cool, whatever's easiest (adding a parameter or JSdoc)

test/cypress/stubs/gateway-account-stubs.js
DomBelcher: This is just how you interact with the stub builder, this is passed in on the `additionalParams` object; where `request` is set it matches the request body against the provided object

test/cypress/stubs/products-stubs.js
```diff
@@ -35,9 +35,10 @@ function getProductsByGatewayAccountIdAndTypeFailure (gatewayAccountId, productT
   })
 }
 
-function postCreateProductSuccess () {
+function postCreateProductSuccess (opts = {}) {
```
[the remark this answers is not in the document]
DomBelcher: No, I think this was a hangover from me trying to fix a particularly tricky test - good spot, I'll remove it

test/cypress/stubs/stub-builder.js
```diff
@@ -21,10 +21,11 @@ function stubBuilder (method, path, responseCode, additionalParams = {}) {
   }
 
   let predicate
-  if (!additionalParams.hasOwnProperty('deepMatchRequest') || additionalParams.deepMatchRequest) { // eslint-disable-line no-prototype-builtins
-    predicate = { deepEquals: request }
-  } else {
+  // cy.log(additionalParams?.deepMatchRequest)
+  if (additionalParams?.deepMatchRequest === false) { // eslint-disable-line no-prototype-builtins
```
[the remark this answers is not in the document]
DomBelcher: I'm unfortunately not fully sure what it does, particularly what shallow matching does

oswaldquek (approved):

PR #4422, opened 2025-01-24
PP-13136 Settings Webhooks landing page

james-peacock-gds:
- implement landing page for simplified settings / webhooks as per designs.
- move webhooks service to services folder for consistency

For subsequent PRs:
- Cypress test
- implementation of the create webhook page.

Not in scope for this story: view and update webhook pages.

**Service with 2 webhooks**
---
Screenshot 2025-01-28 at 12 01 03

**Service with no webhooks**
---
Screenshot 2025-01-28 at 10 42 36


src/views/simplified-account/settings/webhooks/index.njk
```diff
@@ -1,9 +1,75 @@
...
+  </p>
+
+
+  {{ govukButton({
+    text: "Create a new webhook",
+    classes: "govuk-button",
+    href: createWebhookLink
+  }) }}
+
+  {% for webhook in webhooks %}
```
DomBelcher, line 31: In the design there is a line which says either "There are no configured webhooks" or "Configured webhooks (<some number>)" depending on whether there are webhooks configured. The design shows two webhooks but says "Configured webhooks (1)" so I'm not sure whether this refers to the number of active configured webhooks, or if it should be "Configured webhooks (2)"

james-peacock-gds: oops - big oversight there, well spotted, will sort that out.

james-peacock-gds: Following discussion with designers, I've split the headings into active and deactivated and applied the counts accordingly.

DomBelcher (approved):
Nice work

PR #4425, opened 2025-01-24
PP-13312 cypress test card types setting

nlsteers:
### WHAT

- add cypress test for simplified accounts card types setting


DomBelcher (approved):
LGTM

PR #4429, opened 2025-01-29
PP-13136 Create webhook page

james-peacock-gds:
- Update the create webhook page to match design
- Remove some duplicate lines that were spotted in other simplified settings tests.
- NB: This step is just to present the form as it appears on the design. Functionality to be implemented later. This page will also be used for updating existing webhooks, but this responsibility isn't implemented either.

For subsequent PRs:
- logic to handle form submission, including validation and saving the webhook.
- using this page for updating existing webhooks.
- Cypress tests

Screenshot 2025-01-29 at 13 10 41

### DIFF
--- a/src/views/simplified-account/settings/webhooks/create.njk
+++ b/src/views/simplified-account/settings/webhooks/create.njk
@@ -1,10 +1,68 @@
 {% extends "../settings-layout.njk" %}
 
-{% block settingsPageTitle %}
-  Create a webhook
-{% endblock %}
+{% set settingsPageTitle = "Webhook details" %}
 
 {% block settingsContent %}
-  <h1 class="govuk-heading-l">Webhook details</h1>
+  <h1 class="govuk-heading-l">{{ settingsPageTitle }}</h1>
+  <form method="post" novalidate>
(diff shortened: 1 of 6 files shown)


src/views/simplified-account/settings/webhooks/create.njk
```diff
@@ -1,10 +1,68 @@
...
+        classes: "govuk-label--s"
+      },
+      hint: { text: "The HTTPS URL GOV.UK Pay will send webhook messages to" },
+      id: "callback_url",
+      name: "callback_url",
+      value: form.values.callback_url,
+      type: "url",
+      spellcheck: false,
+      errorMessage: form.errors.callback_url and {
+        text: form.errors.callback_url
```
DomBelcher, line 28: Can the uses of `callback_url` be camelCase instead? (with the exception of the ID, which may need to be kebab-case for the validation logic)

james-peacock-gds: ah, yes, I've now removed these errorMessage and value properties because they aren't needed for the purpose of this PR, I will add those in later.

DomBelcher (approved):
LGTM

PR #4432, opened 2025-01-30
PP-13511: Validation for a fixed amount for a payment link

oswaldquek:
Validation shown for when a payment is for less than 30p and the gateway account uses Stripe:

Screenshot 2025-01-30 at 15 37 38


DomBelcher (approved):
LGTM

PR #4433, opened 2025-01-30
PP-12427 switch to worldpay simplified setting scaffold

nlsteers:
### WHAT

- path and route
- basic controller with unit test
- index view
- cypress E2E test for admin and non-admin users

### DIFF
--- a/src/utils/simplified-account/settings/SettingsBuilder.class.js
+++ b/src/utils/simplified-account/settings/SettingsBuilder.class.js
@@ -22,6 +22,7 @@ module.exports = class SettingsBuilder {
       throw new Error('Cannot add setting without category, use .category(name) first.')
     }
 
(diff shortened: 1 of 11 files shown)


src/utils/simplified-account/settings/SettingsBuilder.class.js
```diff
@@ -22,6 +22,7 @@ module.exports = class SettingsBuilder {
       throw new Error('Cannot add setting without category, use .category(name) first.')
     }
 
+    const urlParts = ['simplified', 'settings', `${id}`]
```
DomBelcher, line 25: `/${id}` might be slightly safer just in case one setting id is a substring of another

nlsteers: done

DomBelcher (approved):
LGTM

PR #4436, opened 2025-01-31
PP-13136 Validate and create webhook

james-peacock-gds:
- Validation for create webhook
- Call pay-webhooks to create the webhook
- Handle validation errors from pay-webhooks (pay-webhooks validation includes a check against an allow list of domains).
NB Currently following webhook creation there is a redirect to the main webhooks page. Discussing with designers whether this is the right approach, could potentially direct to the webhook details page (which hasn't yet been added).

For subsequent PRs
- Cypress tests for main webhook page and create webhook page.
- Other webhook settings pages are subject of separate JIRA tickets.

**Screenshot 1. Callback url empty**
Screenshot 2025-01-31 at 13 01 26

---
**Screenshot 2. Callback url non-https**
Screenshot 2025-01-31 at 13 02 07

---
**Screenshot 3. Callback url has domain not in allow list (live accounts only)**
Screenshot 2025-01-31 at 13 14 12

---
**Screenshot 4. Description empty**
Screenshot 2025-01-31 at 13 01 36

---
**Screenshot 5. Description too long**
Screenshot 2025-01-31 at 13 17 14
---

**Screenshot 6. No payment events selected**
Screenshot 2025-01-31 at 13 01 48

---

### DIFF
--- a/src/controllers/simplified-account/settings/webhooks/create/create.controller.js
+++ b/src/controllers/simplified-account/settings/webhooks/create/create.controller.js
@@ -29,7 +29,6 @@ async function post (req, res, next) {
   }
   try {
     await webhooksService.createWebhook(req.service.externalId, req.account.id, accountIsLive, req.body)
-    req.flash('messages', { state: 'success', icon: '&check;', heading: 'Successfully created webhook' })
     res.redirect(formatSimplifiedAccountPathsFor(paths.simplifiedAccount.settings.webhooks.index, req.service.externalId, req.account.type))
   } catch (createWebhookError) {
     if (createWebhookError.errorIdentifier in webhookErrorIdentifiers) {
--- a/src/controllers/simplified-account/settings/webhooks/create/create.controller.js
+++ b/src/controllers/simplified-account/settings/webhooks/create/create.controller.js
@@ -1,7 +1,11 @@
+const paths = require('@root/paths')
 const { response } = require('@utils/response')
 const { constants } = require('@govuk-pay/pay-js-commons')
-const formatSimplifiedAccountPathsFor = require('../../../../../utils/simplified-account/format/format-simplified-account-paths-for')
-const paths = require('@root/paths')
+const { validationResult } = require('express-validator')
+const { webhookSchema, webhookErrorIdentifiers } = require('@utils/simplified-account/validation/webhook.schema')
+const formatSimplifiedAccountPathsFor = require('@utils/simplified-account/format/format-simplified-account-paths-for')
+const formatValidationErrors = require('@utils/simplified-account/format/format-validation-errors')
+const webhooksService = require('@services/webhooks.service')
 
 async function get (req, res) {
   response(req, res, 'simplified-account/settings/webhooks/create', {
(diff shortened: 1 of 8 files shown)


src/controllers/simplified-account/settings/webhooks/create/create.controller.js
```diff
@@ -10,10 +14,47 @@ async function get (req, res) {
...
+      }
+      return responseWithErrors(req, res, callbackUrlError)
+    } else {
+      next(createWebhookError)
+    }
+  }
+}
+
+const responseWithErrors = (req, res, errors) => {
+  const subscriptions = typeof req.body.subscriptions === 'string' ? [req.body.subscriptions] : req.body.subscriptions
```
oswaldquek, line 48: Why would the type of subscriptions be different here? Feels like it should always be the same type

src/services/clients/webhooks.client.js
```diff
@@ -81,7 +81,7 @@ async function createWebhook (serviceId, gatewayAccountId, isLive, options = {})
     service_id: serviceId,
     gateway_account_id: gatewayAccountId,
     live: isLive,
-    callback_url: options.callback_url,
+    callback_url: options.callbackUrl || options.callback_url,
```
oswaldquek, line 84: I think for ease of development, specifying callbackUrl, subscriptions and descriptions in the method signature would be preferable to passing in `options`. This means the developer won't have to look into this webhooks.client.js file is order to figure out what the options are.
Furthermore allowing callbackUrl or callback_url here seems a bit messy.

src/utils/simplified-account/validation/webhook.schema.test.js
```diff
@@ -0,0 +1,103 @@
+const { expect } = require('chai')
+const { webhookSchema } = require('./webhook.schema')
+const { validationResult } = require('express-validator')
+
+describe('Webhook Validation', () => {
+  let BASE_REQ
```
oswaldquek, line 6: A better variable name would be `VALID_REQUEST`.

src/utils/simplified-account/validation/webhook.schema.js
```diff
@@ -0,0 +1,40 @@
+const { body } = require('express-validator')
+
+const WEBHOOK_DESCRIPTION_MAX_LENGTH = 40
```
oswaldquek, line 3: Even though we have this, it would be good to limit the text box in the njk file to 40 chars too. That might save a round trip for the user.

src/controllers/simplified-account/settings/webhooks/create/create.controller.js
james-peacock-gds: if only one box is checked it goes into the req body as a string. If it's multiple then it's an array. I think this is just how that component works (is the same with card types)

src/services/clients/webhooks.client.js
james-peacock-gds: I thought camel case was our agreed convention, so that's what I used in the form. But the old settings form uses snake case.  Doing this was the simplest way I could think of making this work without having to refactor the 'old' webhooks settings code. Similarly for the options. It will be easier to make these sorts of improvements once we've deleted the old settings code, won't it?

src/controllers/simplified-account/settings/webhooks/create/create.controller.js
oswaldquek: I see. That's unfortunate.

src/services/clients/webhooks.client.js
oswaldquek: Oh yeah we did agree on camel case. Yeah we refactor this at a later date. Sorry but can you put something like `// TODO refactor to explicitly pass all params in the method once old webhooks controller code is deleted`?

src/views/simplified-account/settings/webhooks/create.njk
```diff
@@ -33,7 +38,12 @@
       hint: { text: "Short summary of what this webhook will be used for" },
       id: "description",
       name: "description",
-      spellcheck: true
+      value: form.description,
+      attributes: {
+        maxlength: "50"
```
oswaldquek, line 43: Should this be 40?

james-peacock-gds: oops. yes

oswaldquek (approved):

PR #4439, opened 2025-02-04
PP-13557: Specifying that it the stub should only match when there is no query string.

nataliecarey:
### WHAT
There was a breaking change to Run Amock which was intended to make this test work.  The desired behaviour for this test is "it should only match when no query string is present", that lead to a change where all query strings, headers and request bodies were treated in a stricter way than Mountebank (which Run Amock is based on) handles them.

As we're reverting that change this breaking change, this test needs to specify that it should only match when there is no query string, that's done by specifying `query: {}`.

Note - there was a bug in Run Amock which meant that this approach did not work, this is resolved in PR https://github.com/alphagov/pay-run-amock/pull/16

Related PRs:

The change in Run Amock: https://github.com/alphagov/pay-run-amock/pull/16

Updates to other projects:
https://github.com/alphagov/pay-products-ui/pull/2742
https://github.com/alphagov/pay-frontend/pull/3950

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,7 +55,7 @@
       },
       "devDependencies": {
         "@cypress/webpack-preprocessor": "^6.0.2",
-        "@govuk-pay/run-amock": "0.0.8",
+        "@govuk-pay/run-amock": "0.0.9",
         "@pact-foundation/pact": "^12.1.1",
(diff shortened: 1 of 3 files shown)


james-peacock-gds (dismissed):
Still looks good

james-peacock-gds (approved):
LGTM

PR #4440, opened 2025-02-04
Pp 13136 Webhook settings Cypress tests

james-peacock-gds:
- Cypress test for the main simplified settings webhook page
- Tests for the create webhook page will be done in a subsequent PR

### DIFF
--- a/test/cypress/integration/simplified-account/service-settings/webhooks/webhooks.cy.js
+++ b/test/cypress/integration/simplified-account/service-settings/webhooks/webhooks.cy.js
@@ -61,22 +61,29 @@ describe('for an admin', () => {
     ])
     cy.visit(WEBHOOKS_SETTINGS_URL)
   })
-  describe('the settings nav', () => {
+
+  it('should show the settings nav correctly', () => {
     checkSettingsNavigation()
   })
-  describe('the page', () => {
+
+  it.only('should show title, heading and create Webhook button', () => {
     checkTitleAndHeading()
-    it.only('should show webhooks as summary cards', () => {
-      cy.get('div.govuk-summary-card').should('have.length', 3)
-      cy.get('div.govuk-summary-card').first().find('h2').first().contains(DESCRIPTION).get('strong.govuk-tag--blue').contains('Active')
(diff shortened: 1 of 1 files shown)


test/cypress/integration/simplified-account/service-settings/webhooks/webhooks.cy.js
```diff
@@ -0,0 +1,124 @@
...
+  it('should show webhooks as summary cards', () => {
+    cy.get('div.govuk-summary-card').should('have.length', 3)
+    cy.get('div.govuk-summary-card').first().find('h2').first().contains(DESCRIPTION).get('strong.govuk-tag--blue').contains('Active')
+    cy.get('div.govuk-summary-card').first().find('a').first().contains('View')
+    cy.get('div.govuk-summary-card').first().find('a').eq(1).contains('Update')
+    cy.get('div.govuk-summary-card').first().find('dd').first().contains(CALLBACK_URL)
+    cy.get('div.govuk-summary-card').first().find('dd').eq(1).contains('Payment succeeded')
+    cy.get('div.govuk-summary-card').first().find('dd').eq(1).contains('Payment captured')
+    cy.get('div.govuk-summary-card').eq(1).find('h2').first().get('strong.govuk-tag--blue').contains('Active')
+    cy.get('div.govuk-summary-card').eq(2).find('h2').first().get('strong.govuk-tag--yellow').contains('Inactive')
```
DomBelcher, line 92: not super important since it'll still fail if these don't exist but these aren't strictly assertions, it's just trying to find these things in the DOM, adding `should('exist')` on the end of each makes them proper assertions - or equally the `.contains(...)` can be `.should('contain.text', ...)`

DomBelcher: may as well add a few more assertions for the other cards, particularly the inactive webhook card, just to check it shows everything it should

james-peacock-gds: thanks - I've added properties for the other two webhooks and asserted on them

PR #4443, opened 2025-02-04
PP-12427 task list macro

nlsteers:
### WHAT

- move to macro implementation of task list component
- align stripe task model with worldpay
- move `spinner` and `system-messages` macros to macro folder
- break credential forms for worldpay details views into partials for use in switch-psp views


DomBelcher (approved):
LGTM

PR #4444, opened 2025-02-05
PP-12440 Cypress tests for 3DS Flex credentials task

DomBelcher:
### WHAT
 - Cypress tests for displaying 3DS Flex credentials task on index page
 - tests for form flow will follow in later PR


james-peacock-gds (approved):
LGTM

PR #4448, opened 2025-02-06
PP-12427 switching credential logic

nlsteers:
### WHAT

- new 'enum' of credential states
- new connector stripe account setup strategy middleware
- remove call to connector from account strategy middleware for stripe account setup
- add stripe account strategy middleware to stripe details router
- stop mutating the gateway account model with stripe account setup progress
  - move stripe account setup progress to a specific key on the request
- update gateway account model with class method for identifying a switching credential

### DIFF
--- a/src/models/GatewayAccount.class.js
+++ b/src/models/GatewayAccount.class.js
@@ -1,17 +1,20 @@
-const { GatewayAccountCredential, CREDENTIAL_STATE } = require('@models/gateway-account-credential/GatewayAccountCredential.class')
+const { GatewayAccountCredential } = require('@models/gateway-account-credential/GatewayAccountCredential.class')
+const CREDENTIAL_STATE = require('@models/credential-state')
 const Worldpay3dsFlexCredential = require('@models/gateway-account-credential/Worldpay3dsFlexCredential.class')
+const { InvalidConfigurationError } = require('@root/errors')
+const pendingCredentialStates = [CREDENTIAL_STATE.CREATED, CREDENTIAL_STATE.ENTERED, CREDENTIAL_STATE.VERIFIED]
 
 /**
  * @class GatewayAccount
  * @property {string} name - The name of the gateway account
  * @property {string} id - The id of the gateway account
  * @property {string} type - The type of the gateway account (e.g. test/live)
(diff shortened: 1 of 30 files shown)


src/models/GatewayAccount.class.js
```diff
@@ -1,17 +1,20 @@
-const { GatewayAccountCredential, CREDENTIAL_STATE } = require('@models/gateway-account-credential/GatewayAccountCredential.class')
+const { GatewayAccountCredential } = require('@models/gateway-account-credential/GatewayAccountCredential.class')
+const CREDENTIAL_STATE = require('@models/credential-state')
 const Worldpay3dsFlexCredential = require('@models/gateway-account-credential/Worldpay3dsFlexCredential.class')
+const { InvalidConfigurationError } = require('@root/errors')
+const pendingCredentialStates = [CREDENTIAL_STATE.CREATED, CREDENTIAL_STATE.ENTERED, CREDENTIAL_STATE.VERIFIED]
```
DomBelcher, line 5: Would these be better exported by `@models/credential-state`? or are they only ever going to be used here?

src/utils/simplified-account/settings/stripe-details/tasks.test.js
```diff
@@ -74,41 +68,32 @@ describe('friendlyStripeTasks', () => {
...
     expect(result[1].status).to.equal(NOT_STARTED) // eslint-disable-line
     expect(result[2].status).to.equal(CANNOT_START)
   })
 
   it('should handle empty progress object', () => {
-    const accountWithEmptyProgress = {
-      ...account,
-      connectorGatewayAccountStripeProgress: {}
-    }
+    const accountWithEmptyProgress = {}
```
DomBelcher, line 94: I guess this is really just `emptyProgress` since the account is no longer part of it. Same for the other `accountWith...` ones

src/models/GatewayAccount.class.js
nlsteers: given this is just a subset of the constant I didn't think it was necessary to export it specifically

DomBelcher: sure, can always be moved if it ends up being needed somewhere else as well

DomBelcher (approved):
LGTM

PR #4462, opened 2025-02-14
PP-12416: Add more cypress tests to card-payments

kerrr:
### WHAT

Add cypress tests to check:
* the navigation highlight is correct
* the change links go to the correct urls
* can navigate to detail page and return

### DIFF
--- a/test/cypress/integration/simplified-account/service-settings/card-payments/index.cy.js
+++ b/test/cypress/integration/simplified-account/service-settings/card-payments/index.cy.js
@@ -7,6 +7,37 @@ const {
 
 const pageUrl = `/simplified/service/${SERVICE_EXTERNAL_ID}/account/${ACCOUNT_TYPE}/settings/card-payments`
 
+const cardPaymentParams = [
+  {
+    title: 'Collect billing address',
+    url: '/collect-billing-address',
+    listLocation: [0, 0],
+    onValue: 'On',
+    offValue: 'Off'
+  },
+  {
+    title: 'Default billing address country',
+    url: '/default-billing-address-country',
+    listLocation: [0, 1],
+    onValue: 'United Kingdom',
+    offValue: 'None'
+  },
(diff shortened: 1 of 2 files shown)


test/cypress/integration/simplified-account/service-settings/card-payments/index.cy.js
```diff
@@ -73,28 +81,41 @@ describe('Card payments page', () => {
...
+              .should('contain', 'Change')
           })
           cy.get('.govuk-summary-list__row').eq(1).within(() => {
             cy.get('.govuk-summary-list__key').should('contain', 'Default billing address country')
             cy.get('.govuk-summary-list__value').should('contain', 'None')
-            cy.get('.govuk-summary-list__actions').should('contain', 'Change')
+            cy.get('.govuk-summary-list__actions')
+              .find('a')
+              .should('have.attr', 'href', `${pageUrl}/default-billing-address-country`)
+              .should('contain.text', 'Change')
```
nlsteers, line 95: checking the href for correctness is an improvement but I think actually following the link and then navigating back to this page would be a more comprehensive test. There's an example of a test that does something similar here:

https://github.com/alphagov/pay-selfservice/blob/52379b3da97d925141c86582254cdf135f49ba7f/test/cypress/integration/simplified-account/service-settings/stripe-details/task-summary-test-helpers.js#L25

nlsteers (commented):
good to see the settings nav checks in place, I still think we can make the summary list actions test more comprehensive though

test/cypress/integration/simplified-account/service-settings/card-payments/index.cy.js
kerrr: Updated with navigation to the details and back.

nlsteers (approved):
ran the additional tests locally and all is well, great work

PR #4471, opened 2025-02-19
PP-13261: Worldpay index page cypress test for an RCP service

oswaldquek:
Cypress test for https://github.com/alphagov/pay-selfservice/pull/4469

### DIFF
--- a/test/cypress/integration/simplified-account/service-settings/worldpay-details/worldpay-details.cy.js
+++ b/test/cypress/integration/simplified-account/service-settings/worldpay-details/worldpay-details.cy.js
@@ -37,7 +37,7 @@ const setupStubs = (opts = {}, additionalStubs = []) => {
     gatewayAccountStubs.getAccountByServiceIdAndAccountType(SERVICE_EXTERNAL_ID, ACCOUNT_TYPE, {
       gateway_account_id: GATEWAY_ACCOUNT_ID,
       payment_provider: options.paymentProvider,
-      recurring_enabled: opts.recurringEnabled,
+      recurring_enabled: options.recurringEnabled,
       gateway_account_credentials: [{
         payment_provider: options.paymentProvider,
         credentials: options.credentials,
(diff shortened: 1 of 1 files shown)


test/cypress/integration/simplified-account/service-settings/worldpay-details/worldpay-details.cy.js
```diff
@@ -36,6 +37,7 @@ const setupStubs = (opts = {}, additionalStubs = []) => {
     gatewayAccountStubs.getAccountByServiceIdAndAccountType(SERVICE_EXTERNAL_ID, ACCOUNT_TYPE, {
       gateway_account_id: GATEWAY_ACCOUNT_ID,
       payment_provider: options.paymentProvider,
+      recurring_enabled: opts.recurringEnabled,
```
DomBelcher, line 40: I assume this should be `options.recurringEnabled`, otherwise the default value on line 24 isn't used. Not that it matters much since `opts.recurringEnabled` will be `undefined` if not provided, which is falsy

test/cypress/integration/simplified-account/service-settings/worldpay-details/worldpay-details.cy.js
```diff
@@ -185,7 +315,7 @@ describe('Worldpay details settings', () => {
             })
           })
 
-          it('should show the "Configure 3DS" task as "Cannot start yet"', () => {
+          it('should show the "Configure 3DS" task as "Not yet started"', () => {
```
DomBelcher, line 318: good spot

test/cypress/integration/simplified-account/service-settings/worldpay-details/worldpay-details.cy.js
DomBelcher: admittedly I think I wrote that initially and could have chosen better names than `options` and `opts`

DomBelcher (approved):
LGTM

PR #4473, opened 2025-02-20
PP-12427 cypress test switch PSP to Worldpay (for a MOTO service) user journey

nlsteers:
### WHAT

- add test for task overview
- add test for add worldpay credential task

### DIFF
--- a/test/cypress/integration/simplified-account/service-settings/switch-psp/switch-to-worldpay/add-worldpay-credential.cy.js
+++ b/test/cypress/integration/simplified-account/service-settings/switch-psp/switch-to-worldpay/add-worldpay-credential.cy.js
@@ -1,0 +1,237 @@
+const userStubs = require('@test/cypress/stubs/user-stubs')
+const ROLES = require('@test/fixtures/roles.fixtures')
+const gatewayAccountStubs = require('@test/cypress/stubs/gateway-account-stubs')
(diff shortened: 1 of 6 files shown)


test/cypress/integration/simplified-account/service-settings/switch-psp/switch-to-worldpay/add-worldpay-credential.cy.js
```diff
@@ -0,0 +1,237 @@
...
+  })
+
+  describe('For an admin', () => {
+    describe('For a MOTO service', () => {
+      describe('When no credentials have been entered previously', () => {
+        beforeEach(() => {
+          setStubs({
+            moto: true
+          })
+          cy.visit(SWITCH_TO_WORLDPAY_ADD_CREDENTIAL_TASK_SETTINGS_URL)
```
oswaldquek, line 88: Think it would be useful to test we can navigate to this page from the "Switch to Worldpay" screen too.

nlsteers: I agree! There's already something in the `switch-to-worldpay` cypress test that checks that all tasks are navigable from the tasks index

oswaldquek (approved):

PR #4475, opened 2025-02-21
 PP-12427 cypress test switch psp to worldpay (MOTO service)

nlsteers:
### WHAT
 
- add cypress test for the make a live payment task
 - update switch to worlday task test to validate the 'switch to worldpay' function once all tasks are complete


DomBelcher (approved):
Nice work

PR #4478, opened 2025-02-24
PP-13353: Add 3DS flex creds when switching to worldpay

oswaldquek:
Cypress test to follow in another PR.

3DS Flex creds screen when switching to Worldpay:
Screenshot 2025-02-24 at 13 57 15


src/controllers/simplified-account/settings/switch-psp/switch-to-worldpay/add-flex-credentials.controller.test.js
```diff
@@ -0,0 +1,215 @@
...
+  nextStubs,
+  call
+} = new ControllerTestBuilder('@controllers/simplified-account/settings/switch-psp/switch-to-worldpay/add-flex-credentials.controller')
+  .withServiceExternalId(SERVICE_EXTERNAL_ID)
+  .withAccount({
+    type: ACCOUNT_TYPE,
+    providerSwitchEnabled: true,
+    paymentProvider: STRIPE,
+    allowMoto: true
+    // getSwitchingCredential: getSwitchingCredentialStub
```
james-peacock-gds, line 33: remove comment or is this a TODO?

src/utils/simplified-account/validation/worldpay/validations.schema.js
```diff
@@ -0,0 +1,48 @@
+const { body } = require('express-validator')
+
+const ONE_OFF_CUSTOMER_INITIATED_SCHEMA = {
```
james-peacock-gds, line 3: I'm not sure why the validations have to be nested inside the properties here but for THREE_DS_FLEX_VALIDATION they are not.

james-peacock-gds (approved):
A couple of minor points which you can clear up in the next PR if necessary. I also noticed that the design indicates that the example JWT key should wrap to the next line whereas in your screenshot it doesn't - presumably because of the hyphens. If this is important then it could be fixed in the next PR.

src/controllers/simplified-account/settings/switch-psp/switch-to-worldpay/add-flex-credentials.controller.test.js
oswaldquek: Whoops. Might as well remove it in the next PR.

src/utils/simplified-account/validation/worldpay/validations.schema.js
oswaldquek: So the ONE_OFF_CUSTOMER_INITIATED_SCHEMA is not a validation I wrote and [each property seems to be used individually for example](https://github.com/alphagov/pay-selfservice/blob/master/src/controllers/simplified-account/settings/switch-psp/switch-to-worldpay/add-worldpay-credentials.controller.js#L28-L30). Whereas with the THREE_DS_FLEX_VALIDATION I'm doing the validation [all in one shot](https://github.com/alphagov/pay-selfservice/blob/master/src/controllers/simplified-account/settings/switch-psp/switch-to-worldpay/add-flex-credentials.controller.js#L23).

PR #4480, opened 2025-02-25
PP-13687 fix authorisation summary display

SandorArpa:
### WHAT

Admin tool is not showing information about 3D Secure.

- refactor to use right value (`three_d_secure`)


AbdirizakIdris (approved):
Looks good to me!

PR #4485, opened 2025-02-26
PP-13583: Edit route permission for non-admin users

oswaldquek:
Non-admin users should not be allowed to visit /settings/email-notifications/email-collection-mode directly.


james-peacock-gds (approved):
runs locally for me and looks good

PR #4486, opened 2025-02-26
PP-13525 show one test gateway account for degatewayed users

nlsteers:
### WHAT

- update `my-services` controller to enable conditional filtering of test gateway accounts when user is opted into degateway
- simplify `my services` styles and views
- remove reference to `service-switcher` now that accounts are presented as links, not forms
- link to simplified settings for manage team, organisation details and service name settings
  - dynamically load the correct account type for these links depending on the go live status of the service
- update routes and paths
- remove unnecessary call to connector when determining if the service is a Worldpay test service in the `dashboard activity` controller
- remove old `ui` tests, prefer cypress and mocha
- fix issue where account and service resolution middleware would throw an error when adminusers and connector disagree about which gateway account should be returned for a service
  - the middleware now logs a warning when there is a mismatch between connector and adminusers, we prefer connector as the source of truth in this situation

#### NOT DEGATEWAYED

Screenshot 2025-02-28 at 19 06 15

Screenshot 2025-02-28 at 19 06 10

#### DEGATEWAYED

Screenshot 2025-02-28 at 19 06 23

Screenshot 2025-02-28 at 19 05 54


oswaldquek (approved):

oswaldquek:
Nice work

PR #4487, opened 2025-02-28
PP-13522 Update webhook controller tests

DomBelcher:
### WHAT
 - controller tests for update webhook settings
 - add validation to create/update webhook to enforce only valid webhook subscriptions

### DIFF
--- a/src/controllers/simplified-account/settings/webhooks/update/update-webhook.controller.test.js
+++ b/src/controllers/simplified-account/settings/webhooks/update/update-webhook.controller.test.js
@@ -125,5 +125,29 @@ describe('Controller: settings/webhooks/update', () => {
         })
       })
     })
+
+    describe('when the callback URL is not https', () => {
+      before(async () => {
+        nextRequest({
+          body: {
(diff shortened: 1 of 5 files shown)


src/controllers/simplified-account/settings/webhooks/update/update-webhook.controller.test.js
```diff
@@ -0,0 +1,129 @@
...
+        res.redirect.should.have.been.calledOnce // eslint-disable-line no-unused-expressions
+        res.redirect.should.have.been.calledWith('/simplified/service/service-id-123abc/account/test/settings/webhooks')
+      })
+    })
+
+    describe('when there are validation errors', () => {
+      before(async () => {
+        nextRequest({
+          body: {
+            callbackUrl: 'not-a-valid-url',
```
oswaldquek, line 79: Think we should validate if a callback url starts with "http://" in addition?

oswaldquek (commented):
Think we could add another test case

oswaldquek (approved):

PR #4496, opened 2025-03-04
PP-13525 add unit tests for my-services.controller

nlsteers:
### WHAT

- unit tests for the `my-services` controller. Given this controller is temporary, I've covered off all the important cases around filtering, prioritisation and flag setting

### DIFF
--- a/src/controllers/simplified-account/services/my-services.controller.test.js
+++ b/src/controllers/simplified-account/services/my-services.controller.test.js
@@ -1,0 +1,410 @@
+const sinon = require('sinon')
+const paths = require('@root/paths')
+const ControllerTestBuilder = require('@test/test-helpers/simplified-account/controllers/ControllerTestBuilder.class')
+const ServiceRole = require('@models/ServiceRole.class')
+const GatewayAccount = require('@models/GatewayAccount.class')
+const { expect } = require('chai')
+const { WORLDPAY, SANDBOX, STRIPE } = require('@models/constants/payment-providers')
+const { LIVE } = require('@models/constants/go-live-stage')
+const { validServiceResponse } = require('@test/fixtures/service.fixtures')
+const { validGatewayAccount } = require('@test/fixtures/gateway-account.fixtures')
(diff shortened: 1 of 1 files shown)


oswaldquek:
Why is the my-services controller temporary?

nlsteers:
> Why is the my-services controller temporary?

it's temporary in the sense that we will be replacing it when we move off of settings and onto services / service dashboard

src/controllers/simplified-account/services/my-services.controller.test.js
```diff
@@ -0,0 +1,410 @@
...
+            }
+          })
+          call('get')
+        })
+
+        it('should set isWorldpayTestService on the service', () => {
+          sinon.assert.calledWithMatch(mockResponse,
+            sinon.match.any,
+            sinon.match.any,
+            sinon.match.any,
```
oswaldquek, line 108: Might as well match on req, res, and 'simplified-account/services/my-services/index' here?

src/controllers/simplified-account/services/my-services.controller.test.js
```diff
@@ -0,0 +1,410 @@
...
+            sinon.match({
+              services: [
+                sinon.match({
+                  name: SERVICE_NAME,
+                  gatewayAccounts: sinon.match(accounts => {
+                    expect(accounts.length).to.equal(3)
+                    expect(accounts[0].type).to.equal('live')
```
oswaldquek, line 169: Could add the following?
```
expect(accounts[1].type).to.equal('test')
expect(accounts[2].type).to.equal('test')
```

oswaldquek (changes requested):

src/controllers/simplified-account/services/my-services.controller.test.js
nlsteers: I already test the presence and correctness of these keys in other tests so I've omitted them here in favour of making this test easier to reason about

oswaldquek (approved):

PR #4497, opened 2025-03-04
PP-13522 Cypress tests for update webhook flow

DomBelcher:
### WHAT
 - cypress tests for updating webhook
 - return 404 when webhook not found


nlsteers (approved):
Looks good, one comment on how errors are thrown in the webhooks service. 

Can be addressed separately.

PR #4510, opened 2025-03-10
PP-13629 Correct select permission error link

james-peacock-gds:
- remove ids from radio buttons for setting permission on the invite team member page so that the ids are generated automatically, which ensures that the validation error link points to the first radio button.
- fix order of validation error messages to match the order the elements appear in the form

https://github.com/user-attachments/assets/285f59c3-7731-49de-ba11-43aad9a3dcd9


DomBelcher (approved):
LGTM

oswaldquek (approved):

PR #4512, opened 2025-03-10
PP-13522: Flash messages on successful change of status

oswaldquek:
### WHAT

### HOW 
_Steps to test or reproduce:_

### SCREENS
_Screenshots if view has been modified:_


james-peacock-gds (approved):
LGTM - maybe add to the cypress test too in another PR?

oswaldquek:
> LGTM - maybe add to the cypress test too in another PR?

Flash messages is a common functionality in settings-layout.njk so in my view we don't need to be pedantic and test for every message in a cypress test.

PR #4518, opened 2025-03-12
PP-13356 Use ControllerTestBuilder in remove user test

james-peacock-gds:
- Refactor remove user controller test to use ControllerTestBuilder
- also: minor refactor to invite user controller test to avoid using multiple stubs for same method

### DIFF
--- a/src/controllers/simplified-account/settings/team-members/remove-user/remove-user.controller.test.js
+++ b/src/controllers/simplified-account/settings/team-members/remove-user/remove-user.controller.test.js
@@ -1,22 +1,12 @@
-const proxyquire = require('proxyquire')
 const sinon = require('sinon')
 const { expect } = require('chai')
 const User = require('@models/User.class')
 const userFixtures = require('@test/fixtures/user.fixtures')
 const paths = require('@root/paths')
+const ControllerTestBuilder = require('@test/test-helpers/simplified-account/controllers/ControllerTestBuilder.class')
 
 const ACCOUNT_TYPE = 'test'
 const SERVICE_ID = 'service-id-123abc'
-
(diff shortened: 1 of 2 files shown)


src/controllers/simplified-account/settings/team-members/remove-user/remove-user.controller.test.js
```diff
@@ -42,77 +32,87 @@ const viewOnlyUser = new User(userFixtures.validUserResponse(
...
-
-  removeUserController = getController({
-    response: responseStub,
-    renderErrorView: renderErrorViewStub,
-    findByExternalId: findByExternalIdStub,
-    delete: deleteStub
+const mockResponse = sinon.spy()
+const mockRenderErrorView = sinon.spy()
+const mockFindByExternalIdGetsAdminUser = sinon.stub().resolves(adminUser)
+const mockFindByExternalIdGetsViewOnlyUser = sinon.stub().resolves(viewOnlyUser)
```
DomBelcher, line 38: you don't actually need to set up two separate stubs, you can just have one `mockFindUser` included in the `withStubs` in the builder, then set `mockFindUser.resolves(<admin user | view-only user>)` for each test, and then you don't need the `nextStubs()` calls

DomBelcher: (not that there's anything wrong with doing it this way, just a suggestion)

james-peacock-gds: thanks - makes so much sense that I've also applied this to the invite controller test

DomBelcher (approved):
Nice work

PR #4519, opened 2025-03-12
PP-13625 Correct Digital Wallets heading to h2

james-peacock-gds:
On the ‘Card payments' screen heading level 2 (H2) is skipped. ‘Card payments’ is h1 and then ‘Digital wallets’ is h2.
- correct 'Digital wallets' heading to h2


DomBelcher (approved):
big

PR #4524, opened 2025-03-13
PP-13631 Correct webhook form validation error messages

james-peacock-gds:
- ensure order of error messages reflects order of elements in form
- ensure callbackUrl error link points to callbackUrl form element


DomBelcher (approved):
Nice work

PR #4539, opened 2025-03-14
PP-13796 fix api key generation on refresh 

nlsteers:
### WHAT

- reorganise api key controllers
- add new controller showing new key details upon creation
  - when creating a new api key, users are no longer able to refresh the page and continuously generate new keys, they are shown the key details once before being redirected back to the api keys setting index
- extend test coverage
- move api key validations to schema
- split api key name form into nunjucks partial


DomBelcher (approved):

DomBelcher (dismissed):

DomBelcher (approved):
big

PR #4540, opened 2025-03-18
PP-13521: Show webhook event resource

kerrr:
### WHAT

Update the webhook event detail page to show a json representation of the resource from the event, in a design system detail component. This is to match the UI  design.
    
Update the Cypress test for webhook events details page to check for the event body detail.

Update the table showing delivery attempts for webhook events, so that when there is no status code we do not leave a blank cell. This is to improve accessability.

#### AFTER

Screenshot 2025-03-18 at 10 07 12

Screenshot 2025-03-18 at 10 07 45

### DIFF
--- a/src/views/simplified-account/settings/webhooks/event.njk
+++ b/src/views/simplified-account/settings/webhooks/event.njk
@@ -9,8 +9,6 @@
 
 {% block settingsContent %}
 
-
-
   <h1 class="govuk-heading-l">{{ title }}</h1>
 
   {% set resourceLink %}
@@ -44,6 +42,11 @@
         }
       }
     ]
+  }) }}
+
+  {{ govukDetails({
+    summaryText: title + " event body",
+    html: '<pre><code>' + event.resource | dump(4) + '</code></pre>'
   }) }}
 
   <h2 class="govuk-heading-m">Delivery Attempts</h2>
--- a/src/views/simplified-account/settings/webhooks/event.njk
+++ b/src/views/simplified-account/settings/webhooks/event.njk
@@ -50,13 +50,12 @@
   }) }}
 
   <h2 class="govuk-heading-m">Delivery Attempts</h2>
-
   {% set attemptRows = [] %}
   {% for attempt in attempts %}
     {% set attemptRow = [
       { text: attempt.send_at | datetime("datetime") },
       { html: messageStatusTag(attempt.status) },
-      { html: attempt.status_code },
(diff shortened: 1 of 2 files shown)


src/views/simplified-account/settings/webhooks/event.njk
```diff
@@ -46,14 +44,19 @@
...
-  <h2 class="govuk-heading-m">Delivery Attempts</h2>
+  {{ govukDetails({
+    summaryText: title + " event body",
+    html: '<pre><code>' + event.resource | dump(4) + '</code></pre>'
+  }) }}
 
+  <h2 class="govuk-heading-m">Delivery Attempts</h2>
   {% set attemptRows = [] %}
   {% for attempt in attempts %}
+{#    {% set statusCode = attempt.satus_code %}#}
```
DomBelcher, line 55: Delete this

test/cypress/integration/simplified-account/service-settings/webhooks/webhook-event.cy.js
```diff
@@ -106,6 +106,11 @@ describe('for an admin', () => {
     cy.get('h1.govuk-heading-l').should('have.text', 'Payment captured')
   })
 
+  it('should show event body in detail component', () => {
+    cy.get('.govuk-details summary').should('contain.text', 'Payment captured event body')
+    cy.get('.govuk-details div.govuk-details__text pre code').should('contain.text', '{')
```
DomBelcher, line 111: does the formatting of the JSON make it awkward to test for the full event body?

src/views/simplified-account/settings/webhooks/event.njk
kerrr: Thanks. Done.

test/cypress/integration/simplified-account/service-settings/webhooks/webhook-event.cy.js
kerrr: I've had another look at this. I've got a matching object to compare. It's just matching in the string values, so I'm not sure how reliable that is going to be. Do you know if there's a straight forward way for the cypress test to match these as json objects?

DomBelcher: I think matching the stringified JSON works, seems like any option may be a tad awkward. I couldn't find any sensible alternative in the cypress docs

DomBelcher (approved):
LGTM

PR #4543, opened 2025-03-19
PP-13763 `dob` form validation cont

nlsteers:
### WHAT

- when a validation error is returned for the `dob` component where the error is not specific to an individual field within said component, ensure that the error summary links to the first field (`dobDay`)
- tidy up the `dob` custom validations on the stripe person schema

### SCREENS

https://github.com/user-attachments/assets/cdaab9dc-3727-41dc-b58e-3fe32ad406d3


james-peacock-gds (approved):
LGTM

PR #4544, opened 2025-03-19
PP-13624 Prepopulate api key name field

james-peacock-gds:
The ‘API key name’ field is not populated with the previous saved name when the user selects ‘Change name' from the 'API keys’ page. 

- prepopulate the api key name field with the saved value
 
In the screenshots below the original value is 'my original name' and this is shown when the user clicks on 'Change name'.

Screenshot 2025-03-19 at 15 29 46

Screenshot 2025-03-19 at 15 29 53

Screenshot 2025-03-19 at 15 30 02

Screenshot 2025-03-19 at 15 30 08


DomBelcher (approved):
LGTM

PR #4545, opened 2025-03-19
PP-13822 API key name page heading as form label

james-peacock-gds:
Accessibility - On the ‘'API key name’ screen there isn't a form label associated with the text input. 
- add a label to the api key name field
- make this label the page heading

Screenshot 2025-03-19 at 17 12 22


james-peacock-gds:
dealt with in a separate PR

PR #4546, opened 2025-03-19
Revert "PP-13624 Prepopulate api key name field"

nlsteers:
Reverts alphagov/pay-selfservice#4544


james-peacock-gds (approved):
yes agreed we should revert as this is covered in another PR

PR #4547, opened 2025-03-20
PP-13356 Refactor service name controllers to consistent get/post style

DomBelcher:
### WHAT
 - Refactor service name settings controllers to use get/post methods in line with other settings


oswaldquek (commented):
Usually the controllers for the index page sit directly under the `/settings/<functionaliy>` folder so `src/controllers/simplified-account/settings/service-name/index/service-name-index.controller.js` could just be `src/controllers/simplified-account/settings/service-name/service-name.controller.js`

oswaldquek (approved):

PR #4552, opened 2025-03-21
PP-13760 Fix headers and error messages not wrapping when screen resized

james-peacock-gds:
Context: H1 'Change permission for [email address]' does not adjust to screen size so some of the characters cannot be seen without horizontal scrolling. The same applies to the remove user page. Additionally, the validation error summary (and form error message will also extend outside of the boundary of the error box when the screen is resized. This problem also applies to webhooks where a 50-character webhook description may be used in the heading.
- apply `word break: break-word;` across the simplified settings pane

**change permissions before**
-----

Screenshot 2025-03-21 at 14 47 57

-----
**change permissions after**
-----

Screenshot 2025-03-21 at 15 06 38

-----
**remove user after**
-----
Screenshot 2025-03-21 at 14 49 26

### DIFF
--- a/src/assets/sass/components/service-settings.scss
+++ b/src/assets/sass/components/service-settings.scss
@@ -49,7 +49,7 @@
 .service-settings-pane {
   max-width: 630px;
   h1, .service-settings-break-word  {
-    word-break: break-word;
+    overflow-wrap: anywhere;
   }
 }
 
--- a/src/assets/sass/components/service-settings.scss
+++ b/src/assets/sass/components/service-settings.scss
@@ -48,7 +48,9 @@
 
 .service-settings-pane {
   max-width: 630px;
-  word-break: break-word;
+  h1, .service-settings-break-word  {
+    word-break: break-word;
+  }
 }
 
 .service-settings-inset-text--grey {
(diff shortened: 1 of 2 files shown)


james-peacock-gds:
after the last change the screenshot for remove user is as follows:

DomBelcher (dismissed):
nice work

nlsteers (changes requested):
https://caniuse.com/mdn-css_properties_word-break_break-word

james-peacock-gds:
After some research and experimentation:

- use overflow-wrap: anywhere instead of word-break, because word-break is deprecated whereas overflow-wrap is supported by the vast majority of browsers.
- overflow-wrap is not supported by IE11 but in the unlikely event of admin tool users accessing Pay from IE11, this would not affect functionality.
NB I experimented with govuk-!-text-break-word which is designed to address this sort of issue, but it applies a slightly different breaking method (overflow-wrap: break-word) and I found some odd interaction with the max-width property which prevented the word from breaking.

This change ensures that headings, error summary and error messages within simplified settings will wrap appropriately when there is a long word like an email, e.g.

Screenshot 2025-03-25 at 11 29 50

PR #4555, opened 2025-03-24
PP-13817: Add visually hidden text to GOV.UK Payment ID link

oswaldquek:
https://www.w3.org/WAI/WCAG22/Techniques/css/C7 states:

"The objective of this technique is to supplement the link text by adding additional text that describes the unique function of the link"

Therefore, add the hidden text "GOV.UK Payment ID" before the transaction id [link] for a screen reader to read.

### DIFF
--- a/src/views/simplified-account/settings/webhooks/event.njk
+++ b/src/views/simplified-account/settings/webhooks/event.njk
@@ -12,7 +12,7 @@
   <h1 class="govuk-heading-l">{{ title }}</h1>
 
   {% set resourceLink %}
-    <a class="govuk-link" href="{{ resourceLink }}">{{ event.resource_id }}</a>
+    <span class="govuk-visually-hidden">GOV.UK Payment ID</span><a class="govuk-link" href="{{ resourceLink }}">{{ event.resource_id }}</a>
   {% endset %}
   {{ govukSummaryList({
     classes: "webhooks-summary-card",
--- a/src/views/simplified-account/settings/webhooks/event.njk
+++ b/src/views/simplified-account/settings/webhooks/event.njk
@@ -12,7 +12,7 @@
   <h1 class="govuk-heading-l">{{ title }}</h1>
 
   {% set resourceLink %}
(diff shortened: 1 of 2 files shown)


src/views/simplified-account/settings/webhooks/event.njk
```diff
@@ -12,7 +12,7 @@
   <h1 class="govuk-heading-l">{{ title }}</h1>
 
   {% set resourceLink %}
-    <a class="govuk-link" href="{{ resourceLink }}">{{ event.resource_id }}</a>
+    <span class="govuk-visually-hidden">GOV.UK Payment ID</span><a class="govuk-link" href="{{ resourceLink }}">{{ event.resource_id }}</a>
```
kerrr, line 15: Should this have a space after the ID, inside the span? Just wondering how a screenreader would interpret there being no space between "... ID" and the link?

oswaldquek: I have no idea about this, @sallytolson1 can you comment?

kerrr: I've had a dig into the Design System for the [govukSummaryList](https://design-system.service.gov.uk/components/summary-list/#options-summary-list-example--rows-value). Though it doesn't mention this scenario, the sample HTML output when they use the `visuallyHiddenText` in  actions, DOES add a space in the span that is vissually hidden.

```
actions: {
        items: [
          {
            href: "#",
            text: "Change",
            visuallyHiddenText: "address"
          }
        ]
      }
```

becomes

```
<dd class="govuk-summary-list__actions">
      <a class="govuk-link" href="#">Change<span class="govuk-visually-hidden"> address</span></a>
    </dd>
```

So there is some precedent.

sallytolson1: Good point @kerrr, I think we should include a space after "GOV.UK payment ID". Thanks

oswaldquek: Thanks @sallytolson1 , I've committed the change.

PR #4556, opened 2025-03-24
PP-13818: Add visually hidden heading to 'Signing secret'

oswaldquek:
The reason for this change is the same as https://github.com/alphagov/pay-selfservice/pull/4555


DomBelcher (approved):
LGTM

PR #4580, opened 2025-04-08
PP-13622 Fix incorrect negation of value in the admin tool search when a dispute is won

marcotranchino:
With this change, we are fixing a bug for which disputed amounts are shown in the admin tool always as negative numbers.

The logic will now consider also the state of the dispute and present the value as positive when the dispute has been won by the service.

Further information in Jira.

https://payments-platform.atlassian.net/browse/PP-13622

### Screenshot from the cypress tests

Won disputes are positive values

### DIFF
--- a/src/utils/transaction-view.js
+++ b/src/utils/transaction-view.js
@@ -109,9 +109,10 @@ module.exports = {
       }
 
       const transactionType = element.transaction_type && element.transaction_type.toLowerCase()
-      if (transactionType === 'refund' || transactionType === 'dispute') {
+      if (transactionType === 'refund' || (transactionType === 'dispute' && element.state.status !== 'won')) {
         element.amount = `–${element.amount}`
       }
+
       delete element.created_date
     })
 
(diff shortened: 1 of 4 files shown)


src/utils/transaction-view.js
```diff
@@ -109,9 +109,14 @@ module.exports = {
       }
 
       const transactionType = element.transaction_type && element.transaction_type.toLowerCase()
-      if (transactionType === 'refund' || transactionType === 'dispute') {
+      if (transactionType === 'dispute') {
+        console.log('\n\n\nDispute transaction type found. Element state=', element.state)
+        console.log('\n\n\n')
+      }
```
kerrr, line 115: Can we need these logs?

marcotranchino: No, we don't. I'll remove them before publishing the PR. 👍

marcotranchino: Logs removed, now. Thank you

hjvoid (approved):
LGTM 👍

PR #4589, opened 2025-04-16
PP-13526 Remove degateway checks

DomBelcher:
## What

PP-13526
 - remove checks for if a user is degatewayed or simplified settings are enabled
 - this globally enables simplified settings
 - remove a lot of old code/tests which relate to old settings


nlsteers (approved):
🔥

PR #4594, opened 2025-04-24
PP-13984 Upgrade pay-selfservice with newest pay-js-commons

marcotranchino:
## WHAT

With this change, we are updating pay-selfservice to use the new pay-js-commons.

We are also updating the version for pay-js-metrics.


marcotranchino:
No longer needed

PR #4599, opened 2025-04-28
PP-14020 Fix feedback form so that email cannot be replaced

marcotranchino:
With this change, we are fixing a minor issue with an internal feedback form so that service admins cannot replace the email address when sending feedback.

Further information in Jira.

https://payments-platform.atlassian.net/browse/PP-14020


hjvoid (approved):
`email: email` could just be `email` here, but think it's probably more readable to leave as is so LGTM 🙂

PR #4604, opened 2025-05-07
PP-14053 Update payment link deletion functionality

marcotranchino:
## What

With this change, we are changing the functionality for Service Admins to
delete payment links.

We are adding the use of an anti-CSRF token and also using the method POST
instead of GET as for all the other state changing actions.

### How

* Log on as a service admin
* Create a payment link
* Delete the payment link

### Accessibility considerations

No views / interactions have been changed.


src/routes.js
```diff
@@ -372,7 +372,8 @@ module.exports.bind = function (app) {
 
   account.get(paymentLinks.manage.index, permission('transactions:read'), paymentLinksController.getManage)
   account.get(paymentLinks.manage.disable, permission('tokens:create'), paymentLinksController.getDisable)
-  account.get(paymentLinks.manage.delete, permission('tokens:create'), paymentLinksController.getDelete)
+  account.get(paymentLinks.manage.delete, permission('tokens:create'), paymentLinksController.getDelete.get)
```
SandorArpa, line 375: Perhaps renaming getDelete to delete? getDelete.post is a bit contradictory

SandorArpa (changes requested):

src/routes.js
marcotranchino: I was avoiding renaming the file itself. But I agree with you. I will update it now. Thanks!

SandorArpa (approved):
LGTM

src/routes.js
marcotranchino: Amended, thanks!

PR #4609, opened 2025-05-12
BAU fix bug in redirect after login code

nlsteers:
## What

- when accessing a selfservice URL when not logged in, upon logging in you are now redirected back to the URL you were trying to access


manishlad-cs (approved):
auth.service entries look good to me.
Unfamiliarity with code so one question are the noautherror entries also directly for this?

PR #4610, opened 2025-05-13
PP-14012 fix bug in redirect after login code

nlsteers:
## What

- when accessing a selfservice URL when not logged in, upon logging in you are now redirected back to the URL you were trying to access

PP-14012


manishlad-cs (approved):
Looks good to me, and also discussed directly with Nat.

PR #4629, opened 2025-06-16
PP-14222 scaffold new payment links pages

nlsteers:
## What

PP-14222

- add empty views and controllers for new payment links pages
- add new experimental features middleware
  - this middleware function checks for the presence of `EXPERIMENTAL_FEATURES_FLAG` in the environment, returning a 404 if the flag is not set or is false
- update display converter to use the `EXPERIMENTAL_FEATURES_FLAG` environment value when evaluating whether to render the new service navigation


DomBelcher (approved):
LGTM

PR #4630, opened 2025-06-17
PP-14193 Migrate 'test with your users' views to service routes

DomBelcher:
## What

PP-14193
- move `test-with-your-users` pages to service-level routes
- update associated controllers to use the service model

Controllers will be re-written in typescript in later PR, this allows them to be re-written one by one on the new routes 

N/A - views should not have changed


nlsteers (approved):
Approved, don't forget to migrate the views to the new `service-layout` when you do the TS conversion on the controllers

PR #4657, opened 2025-07-15
PP-14218 Spike: Creating payment links in one go

marcotranchino:
## What

Payment link creation SPIKE. All in one go. To be split in stages.


marcotranchino:
I will be pushing a new branch with payment link creation all in one go but using the existing review page.

PR #4667, opened 2025-07-18
PP-14193 Convert create, disable and links controllers to typescript

DomBelcher:
## What

PP-14193
- convert `create`, `disable` and `links` controllers in `test-with-your-users` journey to typescript
- add typescript class for flash messages


ralphm10 (approved):
LGTM

PR #4668, opened 2025-07-18
PP-14241 Remove superfluous fieldset on transactions page

alexbishop1:
Remove superfluous and `legend`-less `fieldset` for card brand form controls on transactions page.

On the same page, WAVE complains that the ‘Card brand’ and ‘Payment status’ `label` elements are orphans because they reference `button` elements after the JavaScript enhances the select to a multi-select. However, the HTML standard says it is valid for a `label` element to reference a `button` element. Add Nunjucks comments clarifying this.

### DIFF
--- a/src/views/transactions/filter.njk
+++ b/src/views/transactions/filter.njk
@@ -110,7 +110,8 @@
           <legend class="govuk-visually-hidden">
               Card brand
           </legend>
-          <fieldset class="govuk-fieldset">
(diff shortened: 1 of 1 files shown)


src/views/transactions/filter.njk
```diff
@@ -130,14 +131,15 @@
               items: cardBrands
             })
           }}
-          </fieldset>
         </fieldset>
```
SandorArpa, line 134: What is this terminating if <fieldset> was removed?

SandorArpa (changes requested):

PR #4694, opened 2025-08-14
PP-14331 Restyle web address widget for new payment link

ralphm10:
## What

PP-[14331](https://payments-platform.atlassian.net/browse/PP-14331)

- remove hardcoding of URL prefix 
- this is not required as set by the environment variable `PRODUCTS_FRIENDLY_BASE_URI`
- if this is not set (in local for example) no default value is provided

> 
Screenshot 2025-08-14 at 11 52 07


nlsteers (approved):
squash these commits when you merge

PR #4701, opened 2025-08-20
PP-14401 Enable testing of *.ts test files, fix failing Typescript tests

DomBelcher:
## What

PP-14401

- typescript tests weren't being run by `npm test`. this fixes that
- this also adds a call to `sinon.resetHistory()` after each test. stubs are therefore reset between tests, meaning `beforeEach` must be used in place of `before` in any test setup that affects stubs
- this also requires the use of `sinon` stubs over spies for any mocking which is used in multiple tests (without additional manual cleanup), as `sinon.resetHistory` only resets stubs

### DIFF
--- a/src/middleware/restrict-to-switching-account.test.ts
+++ b/src/middleware/restrict-to-switching-account.test.ts
@@ -30,10 +30,6 @@ describe('restrictToSwitchingAccount middleware', () => {
     next = sinon.spy()
   })
 
-  afterEach(() => {
-    sinon.restore()
-  })
-
   describe('when account is switching to stripe', () => {
     it('should call next() without error when switching credential provider is stripe', () => {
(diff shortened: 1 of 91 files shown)


src/middleware/restrict-to-switching-account.test.ts
```diff
@@ -31,7 +31,7 @@ describe('restrictToSwitchingAccount middleware', () => {
   })
 
   afterEach(() => {
-    sinon.restore()
+    // sinon.restore()
```
nlsteers, line 34: delete more comments

nlsteers (commented):
looks fine, some errant comments to bin, undraft and lets ship this

nlsteers (approved):

PR #4703, opened 2025-08-26
PP-14352 Retain Welsh language when creating payment links

marcotranchino:
With this change, we are saving the Welsh language option when creating a payment link in Welsh.

Previously, the language was omitted, causing the product to be created incorrectly in English.

### BEFORE

BEFORE

### AFTER

AFTER


DomBelcher (approved):
LGTM

PR #4714, opened 2025-09-02
BAU hygiene

nlsteers:
## What

- delete `tasks.js`, prefer StripeTaskIdentifiers
  - update all controllers to use StripeTaskIdentifiers when evaluating task completeness
- show the name of the deleted metadata column in the success toast when deleting a metadata column from a payment link
- use esm export declaration for PublicAuthClient and ProductsClient as these are only used in typescript
  - update calling services with new imports


src/services/clients/pay/AdminUsersClient.class.ts
```diff
@@ -0,0 +1,27 @@
...
+    super(SERVICE_BASE_URL, SERVICE_NAME)
+    this.users = this.usersClient
+  }
+
+  private get usersClient() {
+    return {
+      getByExternalId: async (externalId: string) => {
+        const path = '/v1/api/users/{externalId}'
+          .replace('{externalId}', encodeURIComponent(externalId))
+        const response = await this.get<UserData>(path, 'get a product')
```
ralphm10, line 20: should this be 'get a user'?

nlsteers: lol yes, good spot

ralphm10 (approved):

PR #4722, opened 2025-09-09
PP-14429 - "Sign in" header not bold when navigating to privacy page

Rafiq-Gasmi:
## What

PP-14429 - "Sign in" header should not be active/bold when on privacy page. Set condition to check if currentPath is /login which sets active/bold text for the sign in header

### How
1) Navigate to [login page](https://selfservice.testing.payments.service.gov.uk/login)
2) Check "Sign in" header is bold
3) Scroll to the bottom of the page
4) Navigate to [Privacy notice](https://selfservice.testing.payments.service.gov.uk/privacy)
5) Check "Sign in" header is NOT bold 

### Screenshots
**Before** 
image

**After**
image

### DIFF
--- a/src/views/includes/header.njk
+++ b/src/views/includes/header.njk
@@ -1,56 +1,60 @@
 {% if loggedIn %}
-  {% set payHeaderNavigation = [
-    {
-      text: "My profile",
-      href: routes.user.profile.index,
-      active: email.length > 0,
-      attributes: {
-        id: "my-profile"
+  {%
+    set payHeaderNavigation = [
+      {
+        text: "My profile",
+        href: routes.user.profile.index,
+        active: email.length > 0,
+        attributes: {
+          id: "my-profile"
+        }
+      },
+      {
+        text: "Features",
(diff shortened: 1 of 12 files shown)


src/views/includes/header.njk
```diff
@@ -45,7 +45,7 @@
     {
       text: "Sign in",
       href: routes.user.logIn,
-      active: true,
+      active: currentPath == "/login",
```
nlsteers, line 48: there are a few 'login' paths including the mfa challenge path that we will still want this link to have active styling for, I don't think this will work as it stands

Rafiq-Gasmi: Changed

src/views/privacy/privacy.njk
```diff
@@ -1,4 +1,5 @@
 {% extends "../layout.njk" %}
+{% set isSignInActive = false %}
```
DomBelcher, line 2: since `undefined` is falsy, you don't actually need to set this as `false`

PR #4728, opened 2025-09-17
PP-14468 Test Welsh service name detour from payment link creation

marcotranchino:
## What

PP-**[14468 ]**

With this change, we are adding tests to make sure that when an admin user creates a Welsh payment link for a service for which the welsh service name had not been set, then they are given the chance to either enter a Welsh service name or to skip this for now and continue with the English service name.

The test checks that:
- this functionality is restricted for live accounts only
- the form validation works well
- when the user skips setting the welsh service name, then the English service name is used in the URL

Further information in JIRA.

https://payments-platform.atlassian.net/browse/PP-14468


ralphm10 (approved):
LGTM

PR #4730, opened 2025-09-18
PP-14545 Save reference data on payment link creation

DomBelcher:
## What

PP-14545

- fix bug with reference data not being saved on payment link creation


ralphm10 (approved):
LGTM

PR #4731, opened 2025-09-22
PP-14283 Transactions view

DomBelcher:
WIP

### DIFF
--- a/src/simplified-account-routes.js
+++ b/src/simplified-account-routes.js
@@ -278,6 +278,13 @@ simplifiedAccount.post(
   servicesController.agreements.cancel.post
 )
 
+// transactions
+simplifiedAccount.get(
+  paths.simplifiedAccount.transactions.index,
+  experimentalFeature,
+  servicesController.transactions.get
+)
+
 // settings index
 simplifiedAccount.get(paths.simplifiedAccount.settings.index, defaultViewDecider)
 
--- a/src/assets/sass/components/transaction-filter.scss
+++ b/src/assets/sass/components/transaction-filter.scss
@@ -10,6 +10,17 @@
   }
 }
 
+.transactions-filter-simplified-account {
+  form {
+    margin: 0 govuk-spacing(3);
+    padding: govuk-spacing(3) 0;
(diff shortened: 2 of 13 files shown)


src/utils/custom-nunjucks-filters/datetime.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const zonedDate = (dateString: string) => {
+  const date = DateTime.fromISO(dateString)
+  if (!date.isValid) {
+    return ''
+  }
+
+  const formattedDate = date.toFormat('H:mm:ss (ZZZZ)')
+  return formattedDate.includes('GMT+1') ? date.toFormat('H:mm:ss') + ' (BST)' : formattedDate
```
ralphm10: The designs specify BST or GMT.
Because BST is not a time zone, (but an offset) we need this logic

src/simplified-account-routes.js
```diff
@@ -323,6 +322,14 @@ simplifiedAccount.post(
   servicesController.agreements.cancel.post
 )
 
+// transactions
+simplifiedAccount.get(
+  paths.simplifiedAccount.transactions.index,
+  experimentalFeature,
+  permission('transactions:read'),
+  servicesController.transactions.list.get
+)
```
DomBelcher: there's already an entry for this route on line 288 - it does need the `experimentalFeature` middleware there though

DomBelcher: if you could add the middleware for the transaction details route on line 294 as well that would be great

src/assets/sass/components/transaction-list.scss
```diff
@@ -19,3 +19,6 @@
 .capitalize {
   text-transform: capitalize;
 }
+
+.transactions-list--item {
+}
```
DomBelcher: is this class meant to have any styling?

DomBelcher (commented):
Nice work, just a few things that may need tweaking

ralphm10 (approved):
LGTM

PR #4743, opened 2025-10-06
PP-14586 Convert card payments settings controllers to typescript

DomBelcher:
## What

PP-14586

 - migrate card payments settings controllers & tests to Typescript


marcotranchino (approved):
LGTM

PR #4744, opened 2025-10-06
PP-14586 Covnert card types settings to typescript

DomBelcher:
## What

PP-14586

 - migrate card types settings controllers & tests to Typescript


src/services/clients/pay/ConnectorClient.class.ts
```diff
@@ -152,6 +163,41 @@ class ConnectorClient extends BaseClient {
       },
     }
   }
+
+  private get cardTypesClient() {
+    return {
+      getAll: async () => {
```
kbottla, line 169: For the new client methods, we need to update any Pact tests to use the new connector client, ensuring integration continues to work as expected.

kbottla (approved):

PR #4749, opened 2025-10-08
PP-14586 Convert Users service to Typescript

DomBelcher:
## What

PP-14586
 - convert users service to Typescript with Typescript adminusers client


marcotranchino (approved):
LGTM

PR #4872, opened 2026-04-08
PP-14622 Display mode links conditionally

ralphm10:
## What

PP-14622

- Add logic to conditionally display the "opposite mode link" only when live accounts exist
- Add Cypress tests
- Add not found error if navigating to mode with no gateway accounts


DomBelcher (approved):
Nice work

PR #4874, opened 2026-04-09
[DO NOT MERGE] PP-15031 Upgrade to Express 5

iqbalgds:
## What

This is a draft PR and will be split up into smaller PRs.

Keep it open in draft for reference.

PP-15031

- Update paths as cannot use regex in paths anymore
- Created a new middleware to set the status filter on routes that use
  a regex to set the status filter.
- Update the catch all from '*' to '*splat' - which is the common used
  convention in Express 5.

### How
- Check a few pages and make sure app is still running.
- In particular, check the `payouts` and `all services transactions` page.

N/A

N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,7 @@
         "common-password": "0.1.2",
         "connect-flash": "0.1.1",
         "cookie-parser": "1.4.x",
-        "express": "4.21.2",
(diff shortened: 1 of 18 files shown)


iqbalgds:
This has been broken down into smaller pull requests and is no longer required.

PR #4877, opened 2026-04-16
PP-15166 Globally set Luxon timezone to Europe/London

DomBelcher:
## What

PP-15166
 - globally set the timezone in `luxon` to `Europe/London` and the `luxon` locale to `en-GB`
   - this ensures that times are always correctly interpreted by `luxon` as being in UK time, with offset adjustmemts calculated correctly when in BST
 - set times in some integration tests to be hardcoded rather than dynamic to ensure dates are formatted as expected
 - add tests for parsing dates & times in transactions searches
 - this fixes a bug where a Cypress test was failing in CI, as the Github action for the Cypress tests is set to UTC by default


DomBelcher:
closing in favour of the approach in #4885

PR #4884, opened 2026-04-23
PP-15031 - update util functions - Express 5 prep

iqbalgds:
## What

PP-**15031**

- Express 5 preparation of code.
- Need to update the utility functions that generate the full paths to roots
  - So that it can handle both Express 4 and Express 5.
- In future PRs we will be upgrading to Express 5
  - At which point the code for Express 4 will be removed.
- We need two utility functions that do the same thing:
  - One for the current code.
  - One for the  newer Simplify accounts code.

### How
- Check the Transactions view for all services for both live and test accounts.

- N/A

- N/A


SandorArpa (approved):
LOKTM

PR #4885, opened 2026-04-28
PP-15166 Parse Luxon DateTimes as Europe/London Timezone

DomBelcher:
## What

PP-15166
- parse `luxon` DateTimes as `Europe/London` times to ensure BST is handled correctly
  - this fixes a bug with a skipped Cypress test failing in CI due to Github actions runtime defaulting to UTC
- send times to Ledger as times in UTC rather than with offsets to match Ledger spec
- add `TimeConstants` class with useful times for fixed points in the past
  - class getters ensure time constants are calculated at (and therefore relative to) the time they are called, rather than at app startup time
- hardcode some times in cypress tests to ensure dates are displayed correctly


ralphm10 (approved):
LGTM - extracting constants, defaults etc. makes it a lot clearer

PR #4886, opened 2026-04-28
PP-15031 - Express 5 prep - use matchedData()

iqbalgds:
## What

PP-**15031**

- Getting things ready for the Express 5 upgrade.
- Update to use the express-validator > matchedData() function
  - Instead of reading the updated req.query object
  - As Express 5 has made req.query read only.

### How
Check the following pages are still working:
- Agreements list page
- Webhooks detail page

N/A

N/A


hjvoid (approved):
LGTM 👍

PR #4888, opened 2026-04-29
PP-15031 Express 5 prep - update (live|test) routes

iqbalgds:
## What

PP-**15031**

- Express 5 does not allow paths such as: `/all-service-transactions/:statusFilter(test|live)`
  - Created a new middleware instead to restrict values to test or live.
  - Update paths to remove the (test|live) as this will be handled by the middleware instead

### How
- Check that pages with  the sub domain `/all-service-transactions` continue to work.
  - Check for both test and live transactions.

N/A

N/A


hjvoid (approved):
LGTM 👍

PR #4891, opened 2026-05-05
PP-15348 4.2.1

marcotranchino:
Further information in the Jira ticket.

https://payments-platform.atlassian.net/browse/PP-15348

### How to reproduce

- Log on as a service admin
- Visit http://127.0.0.1:3000/service/{SERVICE-ID}/account/test/settings/organisation-details
- Enter or update the organisation addressLine1 with <script>alert('hello')</script>
- Save
- When the page loads after the update you should not get an alert 
- and the text <script>alert('hello')</script> should be displayed

Previously the text would not be displayed and the alert would be executed and shown on the webpage.

n/a

alert

### DIFF
--- a/src/utils/simplified-account/validation/organisation-details.schema.ts
+++ b/src/utils/simplified-account/validation/organisation-details.schema.ts
@@ -19,6 +19,7 @@ const organisationDetailsSchema = {
       .notEmpty()
       .withMessage('Enter an organisation name')
       .bail()
+      .escape()
       .isLength({ max: ORGANISATION_NAME_MAX_LENGTH })
       .withMessage(`Organisation name must be ${ORGANISATION_NAME_MAX_LENGTH} characters or fewer`),
   },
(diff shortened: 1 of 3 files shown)


src/utils/simplified-account/validation/organisation-details.schema.ts
```diff
@@ -19,6 +19,7 @@ const organisationDetailsSchema = {
       .notEmpty()
       .withMessage('Enter an organisation name')
       .bail()
+      .escape()
```
DomBelcher, line 22: Since `.escape()` is a sanitizer, it won't work on upgrading to Express 5. If using this, it would be worth using the `matchedData` workaround similar to #4886 

Alternatively you could take the approach of using validation to reject any inputs containing characters such as `<`,`>`, etc

marcotranchino:
I will change approach.

DomBelcher (commented):
As an aside, the reason this is possible is because of this: https://github.com/alphagov/pay-selfservice/blob/fd99ba1025c100f02cf05df5b588b3f2fce18cdb/src/views/simplified-account/settings/organisation-details/index.njk#L41

The address is converted to a HTML string so it can be displayed on multiple lines with `<br />` tags. It _might_ be better to be more strict about what that HTML can be, e.g. with a custom bit of nunjucks, like

```
{% set addressHTML %}
{{ address.line1 }}
...
{% endset %}
```
This may be annoying though with the optional line 2, so may not be a good solution

src/utils/simplified-account/validation/organisation-details.schema.ts
marcotranchino: Thanks Dom, yes, I was just talking with Sandor about this. Will do that.

DomBelcher: I'm completely wrong, this only applies to query parameters, since `req.query` is read-only in Express 5, but `req.body` is still writeable

PR #4892, opened 2026-05-05
PP-15350 validate input for payment_link fields

SandorArpa:
## What

Add extra validation, deny <, >, and | characters to be used for different text. This is to sanitise HTML tags.


marcotranchino (approved):
Nice stuff! LGTM! 👍

PR #4897, opened 2026-05-07
PP-15348 Remediate 4.2.1-B

marcotranchino:
Further information in the Jira ticket.

https://payments-platform.atlassian.net/browse/PP-15348

### How to reproduce

- Log on as a service admin
- Visit http://localhost:3000/service/{SERVICE-ID}/account/test/settings/service-name
- Enter or update the service name (English or Welsh) with: `<script>alert('hello')</script>`
- You should get an error message stating: You cannot use any of the following characters < > | in the service name

Previously the input would be saved, on refresh it would not be displayed and the alert would be executed and shown on the webpage.

n/a

### SCREENSHOT

service name

### DIFF
--- a/src/controllers/simplified-account/settings/service-name/edit/edit-service-name.controller.ts
+++ b/src/controllers/simplified-account/settings/service-name/edit/edit-service-name.controller.ts
@@ -46,18 +46,21 @@ async function post(req: ServiceRequest<EditServiceNameBody>, res: ServiceRespon
   const validations = [
     body('serviceName')
       .trim()
+      .isLength({ max: SERVICE_NAME_MAX_LENGTH })
+      .withMessage(`Service name must be ${SERVICE_NAME_MAX_LENGTH} characters or fewer`)
+      .bail()
       .unescape()
       .matches(/^[^<>|]*$/) // no '<' or '>' or '|' characters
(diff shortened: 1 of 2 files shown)


src/controllers/simplified-account/settings/service-name/edit/edit-service-name.controller.ts
```diff
@@ -46,12 +46,25 @@ async function post(req: ServiceRequest<EditServiceNameBody>, res: ServiceRespon
   const validations = [
     body('serviceName')
       .trim()
+      .unescape()
+      .matches(/^[^<>|]*$/) // no '<' or '>' or '|' characters
+      .withMessage('You cannot use any of the following characters < > | in the service name')
+      .bail()
       .isLength({ max: SERVICE_NAME_MAX_LENGTH })
       .withMessage(`Service name must be ${SERVICE_NAME_MAX_LENGTH} characters or fewer`),
```
DomBelcher: Better to check the length constraint before the regex, so we don't allow unbounded regex matching

marcotranchino: Amended, thanks!

DomBelcher (approved):
LGTM

PR #4898, opened 2026-05-07
PP-15348 render webhook event page safely

SandorArpa:
## What

Refactor web event page to escape json so no malicious script can be executed.

Screenshot 2026-05-07 at 16 55 46

Screenshot 2026-05-07 at 16 56 43

### DIFF
--- a/src/views/simplified-account/settings/webhooks/json/json.njk
+++ b/src/views/simplified-account/settings/webhooks/json/json.njk
@@ -1,0 +1,15 @@
+{% macro json(title, content) %}
+  <div class="govuk-body">
+    <details class="govuk-details">
+      <summary class="govuk-details__summary">
+        <span class="govuk-details__summary-text"> {{ title }} </span>
+      </summary>
+      <div class="govuk-details__text scrollflow">
+        <pre>
+        <code>{{ content | dump('\t') }}</code>
+        </pre
+        >
+      </div>
+    </details>
+  </div>
+{% endmacro %}
--- a/test/cypress/integration/simplified-account/service-settings/webhooks/webhook-event.cy.js
+++ b/test/cypress/integration/simplified-account/service-settings/webhooks/webhook-event.cy.js
@@ -117,10 +117,16 @@ describe('for an admin', () => {
 
   it('should show event body in detail component', () => {
     cy.get('.govuk-details summary').should('contain.text', 'Payment captured event body').click()
-    cy.get('.govuk-details div.govuk-details__text pre code').should(
-      'contain.text',
-      JSON.stringify(WEBHOOK_EVENT_RESOURCE, null, 4)
-    )
+    cy.get('.govuk-details div.govuk-details__text pre code')
+      .invoke('text')
+      .then((text) => {
+        return text.replaceAll('\t', '').replaceAll('\n', '').replaceAll(' ', '')
+      })
+      .then((trimmedText) => {
+        expect(trimmedText).to.eq(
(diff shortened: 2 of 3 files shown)


src/views/simplified-account/settings/webhooks/json/json.njk
```diff
@@ -0,0 +1,15 @@
+{% macro json(title, content) %}
+  <div class="govuk-body">
+    <details class="govuk-details">
+      <summary class="govuk-details__summary">
+        <span class="govuk-details__summary-text"> {{ title }} </span>
+      </summary>
+      <div class="govuk-details__text scrollflow">
+        <pre>
+        <code>{{ content | dump('\t') }}</code>
```
alexbishop1, line 9: This looks like what we use in toolbox so I assume it’s tried and tested.

src/views/simplified-account/settings/webhooks/json/json.njk
```diff
@@ -0,0 +1,15 @@
...
+        </pre
+        >
```
alexbishop1: Is there supposed to be a line break here?

test/cypress/integration/simplified-account/service-settings/webhooks/webhook-event.cy.js
```diff
@@ -117,10 +117,16 @@ describe('for an admin', () => {
...
+    cy.get('.govuk-details div.govuk-details__text pre code')
+      .invoke('text')
+      .then((text) => {
+        return text.replaceAll('\t', '').replaceAll('\n', '').replaceAll(' ', '')
+      })
+      .then((trimmedText) => {
+        expect(trimmedText).to.eq(
+          JSON.stringify(WEBHOOK_EVENT_RESOURCE, null, 0).replaceAll('\n', '').replaceAll(' ', '')
+        )
+      })
```
alexbishop1, line 129: How feasible would it be to check that the output text is actually rendered correctly? As in, could we have a webhook event message where the body contains a literal `<script>` and ensure that we get that as literal text (and not an actual `<script>` tag) on the rendered page?

alexbishop1 (commented):
Does this change how the page appears at all (apart from the obvious changes to rendering of potential XSS content in webhook messages)?

src/views/simplified-account/settings/webhooks/json/json.njk
SandorArpa: Yes, it's a one on one copy from toolbox

src/views/simplified-account/settings/webhooks/json/json.njk
SandorArpa: Prettier did it, I'll try to override it

test/cypress/integration/simplified-account/service-settings/webhooks/webhook-event.cy.js
SandorArpa: I can try and spend some time on it

src/views/simplified-account/settings/webhooks/json/json.njk
SandorArpa: Yeah, Prettier doesn't like </pre> 🤷 
`Checking formatting...
[warn] src/views/simplified-account/settings/webhooks/json/json.njk
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
❌ detected files are not formatted correctly`

alexbishop1 (approved):

PR #4899, opened 2026-05-07
PP-15349 Remove invalid fromDate/toDate from search params in transactions search

DomBelcher:
## What

PP-15349
- update the transaction filter parsing to enforce correctly formatted dates in `toDate` and `fromDate` field
- removes any invalid date strings from the filter object


SandorArpa (dismissed):
LOKTM

SandorArpa (commented):
LEBTMN

SandorArpa (approved):

PR #4907, opened 2026-05-11
PP-15156 Handle search timeout working with nosearch page

DomBelcher:
## What

PP-15156

Adds handling for search timeouts on the All Service Transactions page

- throws a `GatewayTimeoutError` in `transactions.service`
- `GatewayTimeoutError` is caught in the `all-service-transactions` controller, triggering a redirect to `/transactions/:modeFilter/timeout`
- `/transactions/:modeFilter/timeout` renders an error page informing the user the search timed out, with link back to transactions search
- link back to transactions search takes the user to `/transactions/:modeFilter/nosearch` which shows the user the All Service Transactions search view, but does not run a search on load, preventing the user from being stuck in a timeout loop
- Searching on `/transactions/:modeFilter/timeout` takes the user to the standard `/transactions/:modeFilter` search page, displaying search results as normal

This changes behaviour from the existing implementation, which simply renders an error page when a `GatewayTimeoutForAllServicesSearchError` is thrown, with custom content on the error page set by catching this error
- this simplifies the logic, with a dedicated route and controller for this error view, removing the need for varying the rendered HTML on the standard error page
- this also means the user's client is presented with a `302` status instead of a `504`. The user does not need to know that Selfservice is acting as a gateway for Ledger, and so a `504` error is not necessary as there is no obvious gateway timeout from the user's perspective.
- this also reduces noise from `504` errors in Selfservice, which we expect to be thrown by searches timing out
- this only applies to the `all-service-transactions` pages, a `504` on the per-service Transactions pages still returns a 504 to the user. This is much less frequent and not considered desired behaviour, hence the decision to propagate the error
- this may be revised in future, however `504`s from the standard Transactions pages are more important to have monitoring and alerting on, as this is rare and unexpected, whereas timeouts on the All Service Transactions pages are more frequent, and as such less cause for concern

Adds a `ViewMode` class for routes which are not specific to a service, but can be viewed in either test or live

- this is used in the All Service Transactions routes, as they can be viewed in either test or live, to view the corresponding set of transactions
- this class can be used to determine all services which can be viewed in a given mode (i.e. which have the relevant gateway account), optionally with a given permission, and determine if the user has services viewable in the counterpart mode
- this object is attached to relevant routes by the `view-mode-strategy` middleware

> [!IMPORTANT]
> Complete this checklist when adding/modifying nunjucks templates

<details>
<summary>checklist</summary>

- [x] keyboard navigation is validated (can navigate all interactive elements with a keyboard)
- [x] skip to main content takes you to the main content (not a nav link e.g. back link)
- [ ] cypress tests include the `cy.checkAccessibility` command for each unique view
- [ ] the page title is unique
- [x] all page content fits within desktop and mobile view ports (unless specified in the designs)
- [x] links clearly explain where the link will take the user
- [ ] ‘change’ links have visually hidden text providing additional screen reader context
</details>

New views:

Timeout error page:
Screenshot 2026-06-12 at 14 56 38

Nosearch page:
Screenshot 2026-06-12 at 15 33 13


src/controllers/simplified-account/home/all-service-transactions/all-service-transactions-detail.controller.ts
```diff
@@ -6,6 +6,8 @@ import { Transaction } from '@models/transaction/Transaction.class'
 import { TITLE_FRIENDLY_DATE_TIME } from '@models/constants/time-formats'
 
 async function get(req: ServiceRequest, res: ServiceResponse) {
+  req.serviceView.showHeader = false
```
ralphm10, line 9: is `showHeader` misleading here? As this toggles the left hand nav, rather than the header?

src/controllers/simplified-account/home/all-service-transactions/all-service-transactions-refund.controller.ts
```diff
@@ -13,6 +13,8 @@ import { Message } from '@utils/types/express/Message'
 import { TITLE_FRIENDLY_DATE_TIME } from '@models/constants/time-formats'
 
 async function get(req: ServiceRequest, res: ServiceResponse) {
+  req.serviceView.showHeader = false
```
ralphm10, line 16: as above

src/middleware/view-mode-strategy.middleware.ts
```diff
@@ -0,0 +1,38 @@
+import express from 'express'
+import { NotFoundError } from '@root/errors'
+import { ViewMode } from '@models/view-mode/ViewMode.class'
+import { AuthenticatedRequest } from '@utils/types/express'
+
+// for use on routes where determining if the user has services in the given mode is not important
```
ralphm10: could we change this wording slightly to replace `important` - maybe `required` as that would be the opposite of line 20

src/models/gateway/gateway-name.ts
```diff
@@ -0,0 +1,11 @@
+const LEDGER = 'ledger'
+const CONNECTOR = 'connector'
+
+// collection of upstream gateways to which HTTP requests may me made
+// TODO add others as necessary
```
ralphm10, line 5: do we need this TODO? 
If others will be added as part of future feature work, I think it can be removed

src/models/gateway/gateway-name.ts
```diff
@@ -0,0 +1,11 @@
+const LEDGER = 'ledger'
+const CONNECTOR = 'connector'
+
+// collection of upstream gateways to which HTTP requests may me made
```
ralphm10: typo - but would it be more succulent to change to `are made`?

src/models/view-mode/ViewMode.class.ts
```diff
@@ -0,0 +1,94 @@
...
+    this.gatewayAccountIds = gatewayAccountIds
+    this.paymentProviders = paymentProviders
+    this.permission = permission
+
+    this._locals = {
+      links: new ViewModeLinksGenerator(this),
+    }
+  }
+
+  // determines all services viewable in specified mode (test|live)
```
ralphm10: should we keep the comment styling consistent with lines 8-11 and the expression `(test or live)`

src/models/view-mode/ViewMode.class.ts
```diff
@@ -0,0 +1,94 @@
...
+      opposite(modeFilter),
+      hasServicesInOppositeMode,
+      gatewayAccountIdsForMode,
+      paymentProviders,
+      permission
+    )
+  }
+}
+
+function opposite(modeName: ViewModeName) {
```
ralphm10: could this switch be replaced by a key value object - record possibly? 
And then do away with the unknown case?

ralphm10 (commented):
Looks very good overall - just a few minor comments/questions

src/controllers/simplified-account/home/all-service-transactions/all-service-transactions-detail.controller.ts
DomBelcher: This hides this banner on the All Service Transactions detail page, since this is piggy-backing off the service path and middleware to get the account information. It doesn't make sense to show the banner here because this view is "outside" the service

Screenshot 2026-06-15 at 17 13 18

src/models/gateway/gateway-name.ts
DomBelcher: I just put it in so it's clear the list isn't exhaustive

ralphm10 (approved):
great work

PR #4911, opened 2026-05-18
PP-15389 update status filter

ralphm10:
## What

PP-15389

- update filter to use `matchedData` instead of `req.query`
- previously, this was returning a status of `all` to ledger which causes an error as it is not in the valid filters list


DomBelcher (approved):
LGTM

PR #4922, opened 2026-06-04
PP-15384 Fix inaccuracies in how old transactions are presented and described in admin tool

ralphm10:
## What

PP-15384

- update label of all-time filter
- remove date parameters of last 7 years

## How

- create a transaction locally
- modify the created_date to be at least 7 years ago in your local database
- navigate to transactions list page for your service
- select filter "Show all" from date filters
- your transaction should be visible, and no dates should appear in Start/End date

***

Screenshot 2026-06-04 at 16 53 02

### DIFF
--- a/src/client-side/date-select.ts
+++ b/src/client-side/date-select.ts
@@ -38,20 +38,19 @@ function hideTimePicker() {
 }
 
 function clearDates() {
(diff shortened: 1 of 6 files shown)


Rafiq-Gasmi (approved):
LGTM

PR #4925, opened 2026-06-10
PP-15409 Upgrade PACT NPM modules

iqbalgds:
## What

PP-**15409**

- Bump PACT NPM modules
- Update import statements to use the V2 style tests.
  - This will ensure our tests continue to work.
  - In the future, we can consider updating the tests to use the new V4 style.
- Updated the way we publish the PACT tests inline with the new PACT 16 way of doing things.

### How
- Check the PACT tests are running on the deploy environment.


GabrielBoardman-MadeTech (approved):

PR #4929, opened 2026-06-15
PP-14569: Create an env variable feature flag for Adyen go-live request

Pippa101:
## What
PP-14569

- I created an environment variable `ENABLE_GO_LIVE_REQUESTS_FOR_ADYEN` to toggle Adyen Wording in go-live request journey and added it to dev and test config.
- I renamed `experimental-features `to `features.ts` and updated imports and uses where applicable.


src/config/features.ts
```diff
@@ -1,5 +1,6 @@
 const EXPERIMENTAL_FEATURES_FLAG = process.env.EXPERIMENTAL_FEATURES_FLAG === 'true'
 const EXPERIMENTAL_FEATURES_LIST = process.env.EXPERIMENTAL_FEATURES_LIST ?? ''
+const ENABLE_GO_LIVE_REQUESTS_FOR_ADYEN = process.env.ENABLE_GO_LIVE_REQUESTS_FOR_ADYEN
```
DomBelcher: Env vars will always be strings, so if you want to use it as a boolean you should do `process.env.ENABLE_GO_LIVE_REQUESTS_FOR_ADYEN === 'true'` similar to `EXPERIMENTAL_FEATURES_FLAG`

This also covers the case where the env var isn't set, in which case `process.env.ENABLE_GO_LIVE_REQUESTS_FOR_ADYEN` would be `undefined` rather than a string.

DomBelcher (approved):
LGTM

PR #4930, opened 2026-06-16
PP-15483 Transactions list view persist filters

ralphm10:
## What

PP-15483

- Make previously selected filters persist when navigating from transaction detail view to transaction list view
- Applies to both service level and all service transaction pages

### How

- From the all service, or service level transaction list page, select one or more filters
- Navigate to an individual transaction
- Click the `Back to transactions` link
- The URL should include the filters you previously selected, and the list should include filtered transactions only, NOT all transactions

### DIFF
--- a/src/controllers/simplified-account/home/all-service-transactions/all-service-transactions-detail.controller.ts
+++ b/src/controllers/simplified-account/home/all-service-transactions/all-service-transactions-detail.controller.ts
@@ -22,8 +22,10 @@ async function get(req: ServiceRequest, res: ServiceResponse) {
   // sort by most recent first
   events.sort((eventA, eventB) => (eventA.timestamp > eventB.timestamp ? -1 : 1))
 
+  const transactionFilters = req.session.transactionFilters as string
+
   return response(req, res, 'simplified-account/services/all-service-transactions/detail/index', {
-    backLink: formattedPathFor(paths.allServiceTransactions.simplifiedAccount.index, req.account.type),
+    backLink: `${formattedPathFor(paths.allServiceTransactions.simplifiedAccount.index, req.account.type)}${transactionFilters ? `?${transactionFilters}` : ''} `,
     events,
     transaction,
     dispute: disputes.length > 0 && disputes[0],
--- a/src/controllers/simplified-account/services/transactions/detail/transactions-detail.controller.ts
+++ b/src/controllers/simplified-account/services/transactions/detail/transactions-detail.controller.ts
@@ -26,16 +26,19 @@ async function get(req: ServiceRequest, res: ServiceResponse) {
   // sort by most recent first
   events.sort((eventA, eventB) => (eventA.timestamp > eventB.timestamp ? -1 : 1))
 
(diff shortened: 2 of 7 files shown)


src/controllers/simplified-account/services/transactions/detail/transactions-detail.controller.ts
```diff
@@ -26,16 +26,18 @@ async function get(req: ServiceRequest, res: ServiceResponse) {
   // sort by most recent first
   events.sort((eventA, eventB) => (eventA.timestamp > eventB.timestamp ? -1 : 1))
 
+  const transactionFilters = req.session.transactionFilters as string
+
   return response(req, res, 'simplified-account/services/transactions/detail/index', {
-    backLink: formatServiceAndAccountPathsFor(
+    backLink: `${formatServiceAndAccountPathsFor(
```
DomBelcher, line 32: If we want to be clever, we could extract the filters from the session on the search page (assuming no filters were present in the query), and search for those, that way it would work using the browser back button too)

src/controllers/simplified-account/home/all-service-transactions/all-service-transactions-detail.controller.ts
```diff
@@ -24,8 +22,10 @@ async function get(req: ServiceRequest, res: ServiceResponse) {
   // sort by most recent first
   events.sort((eventA, eventB) => (eventA.timestamp > eventB.timestamp ? -1 : 1))
 
+  const transactionFilters = req.session.transactionFilters as string
```
DomBelcher, line 25: It's not so important for this because it's just a string, but it could be useful to do something a bit like the `PaymentLinkCreationSession` in https://github.com/alphagov/pay-selfservice/blob/master/src/controllers/simplified-account/services/payment-links/create/constants.ts

Something like 
```typescript
class TransactionFilterSession {
  filters: string

  static extract (req): string {
    return req.session.transactionFilters as string ?? ''
  }

  static set (req) {...}

  static clear (req) {...}
}
```

which provides a fairly robust interface for the session data.

src/controllers/simplified-account/services/transactions/detail/transactions-detail.controller.ts
DomBelcher: Combined with the `TransactionFilterSession` interface it could be fairly clean, the session can be set, read, and cleared all from the list page.

DomBelcher (commented):
A suggestion for an alternative approach, but not a requirement to do this

src/controllers/simplified-account/home/all-service-transactions/all-service-transactions-detail.controller.ts
DomBelcher: This would leave us open to do all sorts of fancy things like setting an expiry on that specific session data, but again, not required

DomBelcher (approved):
Nice work. Above suggestion is optional for future improvement, but this is working and sufficient for current purposes.

PR #4931, opened 2026-06-16
PP-15485 Update links from all service transactions refund

ralphm10:
## What

PP-15485

- Update links from all service transactions refund to correctly redirect to all service views

### How
- from the all service transactions view, select a transaction that is refundable (for example with a status of success)
- Click `Refund payment` button
- Select `Full amount` radio button
- Click `Confirm refund` button
- You should be directed to the all transactions refund page (with no side bar navigation) and a banner should display with text including `Refund successful`


DomBelcher (approved):
Nice 🎉

PR #4934, opened 2026-06-17
PP-15481 Show card brand for refund and dispute rows in Transactions list

DomBelcher:
## What

PP-15481
- Fix bug where card brands were missing for refund and dispute rows in the Transactions lists
- Modify `card_details`/`cardDetails` types for Transactions to make all properties possibly undefined
  - some of these values may always be returned by Ledger, but guaranteeing this is extremely hard and error prone
  - it's safer to just assume they may not be present, and show empty strings instead if so
- delete unused `TransactionCardDetails` file

Screenshot 2026-06-17 at 12 25 28
The above screenshot is slightly inaccurate as it shows no card brand for the "Dispute awaiting evidence" transaction. This row also shows the incorrect reference. This is due to data being manipulated locally (as simulating a dispute locally is non-trivial) and should show the correct data in live environments


src/models/common/card-details/dto/CardDetails.dto.ts
```diff
@@ -1,15 +1,20 @@
 export interface CardDetailsData {
-  cardholder_name: string
-  billing_address: {
+  // it is unclear whether these fields are always present
+  // but it is safest to assume not
+  card_brand?: string
```
kbottla, line 2: Do we need additional comments? `?` implies they are not present always

DomBelcher: I can't work out what values are or aren't always present, beyond knowing that some of them aren't. I thought it was safest to mark them `?` so the case where they aren't present has to be explicitly handled, so we wouldn't get errors in production if there were transactions without any of this data

kbottla (approved):

PR #4935, opened 2026-06-17
PP-15481 Add content for payments which are non-refundable due to being disputed

DomBelcher:
## What

PP-15481
- add content on Transaction Detail page for disputed payments which are not refundable
- this is existing behaviour in the current Transactions pages, however the existing pages display the same content for all dispute states, specifically `You cannot refund this payment because it is being disputed.`
  - this adds slightly different content for payments where the dispute has been lost, and so it is no longer accurate to say the payment is _being_ disputed.

Screenshot 2026-06-17 at 15 01 01

### DIFF
--- a/src/views/simplified-account/services/transactions/detail/index.njk
+++ b/src/views/simplified-account/services/transactions/detail/index.njk
@@ -9,8 +9,6 @@
 
 {% block serviceContent %}
   <h1 class="govuk-heading-l govuk-!-margin-top-6">Transaction details</h1>
-  {% include './partials/_details-page.njk' %}
-
   {% if isOverSevenYearsOld %}
     {{
       govukInsetText({
--- a/src/views/simplified-account/services/transactions/detail/index.njk
+++ b/src/views/simplified-account/services/transactions/detail/index.njk
@@ -28,9 +28,9 @@
   {% elif dispute and dispute.isDisputePending() %}
     <p class="govuk-body">You cannot refund this payment because it is being disputed.</p>
   {% elif dispute and dispute.isDisputeLost() %}
-    <p class="govuk-body"
-      >You cannot refund this payment because it was disputed and the paying user won the dispute.</p
-    >
+    <p class="govuk-body">
+      You cannot refund this payment because it was disputed and the paying user won the dispute.
+    </p>
   {% endif %}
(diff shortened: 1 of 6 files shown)


src/views/simplified-account/services/transactions/detail/index.njk
```diff
@@ -26,6 +27,12 @@
         href: transaction._locals.links.refund
       })
     }}
+  {% elif dispute and dispute.isDisputePending() %}
```
kbottla, line 28: Is there a test for a disputed transaction which can assert this message?

DomBelcher: I'll add some cypress tests

test/cypress/integration/simplified-account/transactions/transaction-detail.cy.ts
```diff
@@ -419,115 +420,265 @@ describe('Transaction details page', () => {
       })
   })
 
-  it('should display dispute information', () => {
-    const parentTransactionOfDispute = new TransactionFixture({ disputed: true }).toTransactionData()
-
-    const disputeTransaction = {
-      gateway_account_id: GATEWAY_ACCOUNT_ID,
+  describe.only('for disputed payments', () => {
```
kbottla: only?
```suggestion
  describe('for disputed payments', () => {
```

src/views/simplified-account/services/transactions/detail/index.njk
```diff
@@ -26,6 +25,12 @@
...
+    <p class="govuk-body">You cannot refund this payment because it is being disputed.</p>
+  {% elif dispute and dispute.isDisputeLost() %}
+    <p class="govuk-body"
```
kbottla: With this format, prettier won't make the tag endings weird.

```suggestion
 <p class="govuk-body">
      You cannot refund this payment because it was disputed and the paying user won the dispute.
 </p>
```

test/cypress/integration/simplified-account/transactions/transaction-detail.cy.ts
DomBelcher: oops, left that in by mistake. good catch

kbottla (approved):
LGTM!

PR #4936, opened 2026-06-17
PP-15481 Fix bug with displayed dates & times changing in filters on search

DomBelcher:
## What

PP-15481
There is a bug where dates and times in the text inputs on the Transactions list page change on searching for transactions
- The start date would go back one day
- both the start and end times would go back by one hour

This only happened for times during BST, and was seemingly caused by the dates being interpreted as BST, then formatted as UTC

This PR fixes this bug

https://github.com/user-attachments/assets/984e464d-8ee8-49ae-95ef-027d75157392


Rafiq-Gasmi (approved):
LGTM

PR #4938, opened 2026-06-18
PP-15440 Add Your provider is changing page

kbottla:
## What
- Added a new page with information about the change to Adyen
- Only available to Stripe admin users

image


src/controllers/simplified-account/settings/switch-psp/switch-to-adyen/switch-to-adyen.controller.ts
```diff
@@ -0,0 +1,18 @@
+import type { ServiceRequest, ServiceResponse } from '@utils/types/express'
+import { response } from '@utils/response'
+import formatServiceAndAccountPathsFor from '@utils/simplified-account/format/format-service-and-account-paths-for'
+import paths from '@root/paths'
+
+function getProviderChangeToAdyenPage(req: ServiceRequest, res: ServiceResponse) {
```
DomBelcher: Can this be named simply `get` in line with other controllers?

src/paths.js
```diff
@@ -265,6 +265,10 @@ module.exports = {
         event: '/settings/webhooks/:webhookExternalId/event/:eventId',
       },
       switchPsp: {
+        switchToAdyen: {
+          providerChangeToAdyen: '/settings/provider-change-to-adyen',
+          adyenFees: '/settings/provider-change-to-adyen/adyen-fees',
```
DomBelcher: Do we want the URLs to include `/switch-psp/`?

test/cypress/integration/simplified-account/service-settings/switch-psp/switch-to-adyen/provider-change-to-adyen.controller.cy.js
```diff
@@ -0,0 +1,92 @@
...
+        cy.get('h1').should('contain.text', 'An error occurred')
+      })
+    })
+    describe('Stripe test account', () => {
+      it('should show page not found error to admin user', () => {
+        setStubs({
+          role: 'admin',
+          gatewayAccountType: GatewayAccountType.TEST,
+        })
+        cy.visit(PROVIDER_CHANGE_TO_ADYEN, { failOnStatusCode: false })
```
DomBelcher, line 71: The setup here is slightly off I think, the `PROVIDER_CHANGE_TO_ADYEN` URL includes `/live`, so the middleware will attempt to resolve a live account, but the stubs are configured for a test account. 

This is a valid test case I suppose, as a user could enter this URL themselves when they have no live account, although I'm a bit confused as to why it seems to produce the correct page title. I would have thought this would still produce a 404, which would result in the standard `An error occurred - GOV.UK Pay` page title.

DomBelcher: Ah, I see what's going on - the stub setup on line 28 is hardcoded to `LIVE_ACCOUNT_TYPE` for the parameter in the URL, even though the stub setup sets the account type as `test`, so the stub is set up on `/service/.../account/live/...`, but returns a test account.

src/controllers/simplified-account/settings/switch-psp/switch-to-adyen/switch-to-adyen.controller.ts
kbottla: We will need to add more pages (prepare to switch and switch to Adyen) along with this one. So, we will end up with a few get methods

src/paths.js
kbottla: Yes, good spot! Changed to `settings/switch-psp/switch-to-adyen/provider-change-to-adyen` so all live under `switch-to-adyen` context.

test/cypress/integration/simplified-account/service-settings/switch-psp/switch-to-adyen/provider-change-to-adyen.controller.cy.js
kbottla: Thanks! Yes, missed changing LIVE_ACCOUNT_TYPE. Updated

src/controllers/simplified-account/settings/switch-psp/switch-to-adyen/switch-to-adyen.controller.ts
DomBelcher: The pattern we were working with in recent work was to have one controller per route, with each controller only exporting functions for HTTP verbs for the actions on that route. It keeps the controllers nice and small and I think helps reduce cognitive load

kbottla: Update controller name and also the method name to `get` https://github.com/alphagov/pay-selfservice/compare/3b6f400de96d99b167614019122e5f7b23e44040..3cdaec7f480b347e88010db892125e83da0ba9e5

DomBelcher (approved):
Nice work 🎉

PR #4941, opened 2026-06-18
PP-15481 Correct net values for disputes and refunds

DomBelcher:
## What

PP-15481

 - correct Net Amount values displayed for refunds and disputes
   - show Net amounts as negative, in line with existing Transactions pages
 - amend display status for lost disputes from "Dispute lost to customer" to "Dispute lost to user"
 - make Cypress tests for disputes more comprehensive, with mock data based on actual Ledger responses

The `net_amount` field is only returned by Ledger for lost disputes, so this is the only case where it should be displayed. It is returned as a negative value, so there is no need to negate it in code.

Refund transactions have no `net_amount` field, so where the net amount is shown in the Transaction list, it is calculated from its `amount` field, and must be negated. Where the "Refunded amount" is shown on the Transaction details page, this is calculated from the refund summary. The `amount_refunded` field is positive, so this must also be negated.

Screenshots are taken from Cypress tests, and therefore based on mock data. This data is based on actual responses from Ledger for each of the cases, and should be correct.

Partial refund:
Screenshot 2026-06-19 at 11 15 10
Full refund:
Screenshot 2026-06-19 at 11 15 18

Dispute states:
Screenshot 2026-06-19 at 11 24 55
Screenshot 2026-06-19 at 11 25 03
Screenshot 2026-06-19 at 11 25 14
Screenshot 2026-06-19 at 11 25 34


ralphm10 (approved):
LTGM
I like the approach of splitting out the refund and dispute tests from the details ones

PR #4943, opened 2026-06-19
PP-15481 Update template name

ralphm10:
## What

PP-15481

- Update incorrect template path in all services refund controller
- Add Cypress tests covering error scenarios for refunds

### How

- navigate to a transaction that is NOT fully refunded from either service level, or all services list view
- Follow one or more of the below journeys
  - Click `Confirm refund` with nether radio button selected
  - Click `Partial refund` radio button, enter an amount higher than the remaining amount to be refunded, click `Confirm refund`
  - Click `Partial refund` radio button, click `Confirm refund`
  - Click `Partial refund` radio button, enter a string, or incorrectly formatted number (eg. `.50`), click `Confirm refund`
- You will receive a helpful error message explaining the problem, and not the generic error page with message `An error occurred` 

> 
Screenshot 2026-06-19 at 13 55 24

***

Screenshot 2026-06-19 at 13 55 44

***

Screenshot 2026-06-19 at 13 56 04

***

Screenshot 2026-06-19 at 13 56 36


src/controllers/simplified-account/home/all-service-transactions/all-service-transactions-refund.controller.ts
```diff
@@ -66,7 +66,7 @@ async function post(req: ServiceRequest<TransactionRefundBody>, res: ServiceResp
   const errors = validationResult(req)
   if (!errors.isEmpty()) {
     const formattedErrors = formatValidationErrors(errors)
-    return response(req, res, 'simplified-account/services/all-service-transactions/refund', {
+    return response(req, res, 'simplified-account/services/all-service-transactions/detail/refund', {
```
kbottla, line 69: It would be good to have a Cypress test for an unhappy path

ralphm10: I agree! 
Will add

DomBelcher (approved):
Nice 🎉

PR #4946, opened 2026-06-19
PP-15481 Make date range filters and entered dates more consistent

DomBelcher:
## What

PP-15481
improve consistency between selected date range filters and the displayed to/from dates

when using Javascript:
- choosing a date in the date picker will update the date range filter
 - if the entered dates conform to one of the possible date ranges, this will be selected
   - otherwise, it will default to `Between two dates`
   - this will update on-the-fly with client-side JS

When not using Javascript:
- selecting a valid filter will take priority over entered dates
- as with JS version, if the entered dates conform to one of the possible date ranges, this will be selected
  - this will happen when the form is submitted and the page reloads

A `jsEnabled` hidden input is added to identify if the search is performed with JS enabled or disabled, and handle the search accordingly. This allows for tuning the way dates are handled in each case.

Additional changes:
 - Extract date and time parsing code in `TransactionSearchParams` with aim of improving clarity & readability
   - this is also partially rewritten and additional comments added to improve readability
 - enforce computed dates in `TransactionSearchParams` are always valid by using default values in parsing dates & times
 - This also adds some types for JQuery and JQuery utils, and converts some client-side code to Typescript.

### DIFF
--- a/src/client-side/date-select.ts
+++ b/src/client-side/date-select.ts
@@ -61,7 +61,6 @@ function setDateFilter() {
   const toDate = (document.getElementById('toDate') as HTMLInputElement).value
 
   const period = dateRangeAsPeriod(fromDate, toDate, TRANSACTION_SEARCH_DATE_FORMAT)
-  console.log(period)
   ;(document.getElementById('dateFilter') as HTMLInputElement).value = period ?? 'custom-range'
 }
 
@@ -95,7 +94,6 @@ function bindDatePickerChangeListener() {
     $('.date-picker')
       .datepicker()
       .on('changeDate', (_) => {
-        console.log('date changed')
         setDateFilter()
       })
   }
--- a/src/client-side/date-select.ts
+++ b/src/client-side/date-select.ts
@@ -4,11 +4,22 @@ import {
   getPeriodUKDateTimeRange,
   dateRangeAsPeriod,
 } from '@utils/simplified-account/services/dashboard/datetime-utils'
-declare const $: JQueryStatic
+declare const $: JQueryStatic | undefined
 
 import { TRANSACTION_SEARCH_DATE_FORMAT } from '@utils/time/time-formats'
+import awaitJQuery from '@utils/client-side/await-jquery'
 
 function register() {
+  setJsEnabled()
+  bindDateFilterChangeListener()
+  bindIncludeTimeCheckboxListener()
+
+  awaitJQuery(() => {
+    bindDatePickerChangeListener()
+  })
+}
+
+function bindDateFilterChangeListener() {
   document.getElementById('dateFilter')?.addEventListener('change', (event) => {
     if (!(event.target instanceof HTMLSelectElement)) {
       return
@@ -22,14 +33,9 @@ function register() {
     setFromDate(dates.start)
     setEndDate(dates.end)
   })
+}
 
-  $('.date-picker')
-    .datepicker()
-    .on('changeDate', (_) => {
-      console.log('date changed')
-      setDateFilter()
-    })
-
+function bindIncludeTimeCheckboxListener() {
   document.getElementById('include-time-checkbox')?.addEventListener('change', (event) => {
     if (!(event.target instanceof HTMLInputElement)) {
       return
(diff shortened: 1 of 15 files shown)


src/client-side/date-select.ts
```diff
@@ -29,6 +49,22 @@ function register() {
...
+    jsEnabledHiddenInput.value = 'true'
+  }
+}
+
+function setDateFilter() {
+  const fromDate = (document.getElementById('fromDate') as HTMLInputElement).value
+  const toDate = (document.getElementById('toDate') as HTMLInputElement).value
+
+  const period = dateRangeAsPeriod(fromDate, toDate, TRANSACTION_SEARCH_DATE_FORMAT)
+  console.log(period)
```
ralphm10: can we remove this?

src/client-side/date-select.ts
```diff
@@ -44,13 +80,24 @@ function clearDates() {
...
+    ;(document.getElementById('toDate') as HTMLInputElement).value = date.toFormat(TRANSACTION_SEARCH_DATE_FORMAT)
+  }
+}
+
+function bindDatePickerChangeListener() {
+  if (Object.hasOwn(window, '$') && $) {
+    $('.date-picker')
+      .datepicker()
+      .on('changeDate', (_) => {
+        console.log('date changed')
```
ralphm10: and this?

src/models/transaction/TransactionSearchParams.class.ts
```diff
@@ -199,3 +186,64 @@ function parsePageNumber(pageNumber?: string) {
...
+
+const processDateAndTime = (queryParams: TransactionSearchQuery): DateTimeSearchParams => {
+  const searchParams: DateTimeSearchParams = {}
+
+  const dateRange = getPeriodUKDateTimeRange(queryParams.dateFilter as Period)
+  const isJsEnabled = queryParams.jsEnabled !== 'false'
+  const isDateFilterValidPeriod = TRANSACTION_FILTER_PERIODS.has(queryParams.dateFilter as Period)
+
+  // with JS enabled, override the selected filter if any dates are entered
+  const isJsDateFilterSearch = isJsEnabled && !queryParams.fromDate && !queryParams.toDate && isDateFilterValidPeriod
```
ralphm10: I think it might read more clearly to extract `!queryParams.fromDate && !queryParams.toDate` to a variable - `hasNoDateFilter` or something?

src/models/transaction/TransactionSearchParams.class.ts
```diff
@@ -199,3 +186,64 @@ function parsePageNumber(pageNumber?: string) {
...
+      queryParams.includeTime === 'include',
+      TimeConstants.END_OF_TODAY
+    )
+    searchParams.toDate = queryParams.includeTime === 'include' ? toDate : toDate.endOf('day')
+
+    // if the selected dates match the filter, persist the filter, otherwise set to custom-range
+    // this ensures the correct filter is displayed to the user
+    searchParams.dateFilter =
+      queryParams.fromDate === dateRange.start?.toFormat(TRANSACTION_SEARCH_DATE_FORMAT) &&
+      queryParams.toDate === dateRange.end?.toFormat(TRANSACTION_SEARCH_DATE_FORMAT)
```
ralphm10: likewise with this - might be clearer to extract to make a bit more readable? 
Something along the lines of `isDateRangeMatch`

src/utils/time/time-formats.ts
```diff
@@ -0,0 +1,5 @@
+export const TRANSACTION_SEARCH_DATE_FORMAT = 'dd/LL/yyyy'
```
ralphm10, line 1: we use this format in a few other places (such as `src/utils/simplified-account/services/dashboard/datetime-utils.ts` and `src/utils/custom-nunjucks-filters/datetime.ts`

Wondering if it is possible to rename to something more generic (`DMY_FORMAT` perhaps) and replace the usages?

ralphm10 (commented):
Overall looks great and works as expected. 

Just a few comments.

src/utils/time/time-formats.ts
DomBelcher: I was cautious about introducing a change in too many files at once, and wanted to make sure the name of this was scoped to it's current usage. I reasoned even if the same format is used elsewhere, it may be appropriate for it to have it's own definition, so they could be changed independently, although I'm a bit uncertain on that.

ralphm10: That makes sense - I don't feel strongly about this one

ralphm10 (approved):

PR #4947, opened 2026-06-22
PP-15440 Add Your provider is changing link to dashboard and navigation

kbottla:
## What
- Makes the link `Your provider is changing to Adyen` available on Dashboard and settings navigation for Stripe admin users and when the flag is enabled.

### - Dashboard

image

### - Navigation

image

### DIFF
--- a/src/utils/simplified-account/navigation/service-settings-navigation.ts
+++ b/src/utils/simplified-account/navigation/service-settings-navigation.ts
@@ -7,6 +7,7 @@ import Service from '@models/service/Service.class'
 import { LIVE } from '@models/constants/go-live-stage'
 import UserPermissions from '@models/user/permissions'
 import GatewayAccountType from '@models/gateway-account/gateway-account-type'
+import { Features } from '@root/config/features'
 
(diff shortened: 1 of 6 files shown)


src/utils/simplified-account/navigation/service-settings-navigation.ts
```diff
@@ -101,6 +102,21 @@ export = (account: GatewayAccount, service: Service, currentUrl: string, permiss
       hasPermission: UserPermissions.settings.gatewayCredentials.gatewayCredentialsUpdate,
       conditions: account.isSwitchingToProvider(STRIPE),
     })
+    .add({
+      id: 'provider-change-to-adyen',
+      altId: 'provider-change-to-adyen',
```
DomBelcher: I think you don't need the `altId` - if I remember right the `id`/`altId` are path components that the navigation checks for to determine whether to show the specific item as selected, so in most cases you only need one

kbottla: Removed altId

DomBelcher (approved):
Nice 🎉

PR #4950, opened 2026-06-22
PP-15481 Add max transaction count

ralphm10:
## What

PP-15481

- Update text shown on transaction list pages if number is greater than limit 

> 
Screenshot 2026-06-22 at 15 37 40


DomBelcher (approved):
LGTM

PR #4951, opened 2026-06-23
PP-15481 Fix CSV download display

ralphm10:
## What

PP-15481

- Fixes a bug where the CSV download link is displayed on page 2+, even when transactions are over the max limit
- Achieved through removing `page` from the list of search params which are used to determine whether or not the link should be displayed
- Add Cypress tests for this logic

>
Screenshot 2026-06-23 at 14 08 12

***

Screenshot 2026-06-23 at 14 10 28

### DIFF
--- a/src/controllers/simplified-account/services/transactions/transaction-list.controller.ts
+++ b/src/controllers/simplified-account/services/transactions/transaction-list.controller.ts
@@ -71,7 +71,11 @@ async function get(req: ServiceRequest, res: ServiceResponse) {
   const downloadQueryString = transactionSearchParams.getQueryParams().toString()
   const downloadLink = downloadQueryString.length ? `${downloadUrl}?${downloadQueryString}` : downloadUrl
   const transactionCountWithinRange = results.total > 0 && results.total <= LEDGER_TRANSACTION_COUNT_LIMIT
-  const hasQueryParams = transactionSearchParams.getQueryParams().toString().length
+
+  const userSelectedFilters = new URLSearchParams(transactionSearchParams.getQueryParams());
+  userSelectedFilters.delete("page");
+
+  const hasQueryParams = userSelectedFilters.toString().length
   const showCsvDownload =
     transactionCountWithinRange || (hasQueryParams && results.total > LEDGER_TRANSACTION_COUNT_LIMIT)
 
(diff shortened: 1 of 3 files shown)


src/controllers/simplified-account/services/transactions/transaction-list.controller.ts
```diff
@@ -71,7 +71,11 @@ async function get(req: ServiceRequest, res: ServiceResponse) {
   const downloadQueryString = transactionSearchParams.getQueryParams().toString()
   const downloadLink = downloadQueryString.length ? `${downloadUrl}?${downloadQueryString}` : downloadUrl
   const transactionCountWithinRange = results.total > 0 && results.total <= LEDGER_TRANSACTION_COUNT_LIMIT
-  const hasQueryParams = transactionSearchParams.getQueryParams().toString().length
+
+  const userSelectedFilters = new URLSearchParams(transactionSearchParams.getQueryParams())
+  userSelectedFilters.delete('page')
```
ralphm10: I don't love this variable name so open to suggestions. 
Also wondering whether we need a comment to explain why we have removed `page`?

test/cypress/integration/simplified-account/transaction/transaction-index.cy.ts
```diff
@@ -836,5 +836,100 @@ describe('Transactions index', () => {
       cy.get('#state').invoke('text').should('contain', 'Success')
       cy.get('#card-brand').invoke('text').should('contain', 'Visa')
     })
+
+    it('should not display CSV download link on subsequent pages if transaction count is over limit', () => {
```
ralphm10, line 840: wasn't sure whether these tests should go in the pagination or CSV download block

src/controllers/simplified-account/services/transactions/transaction-list.controller.ts
DomBelcher: could make it a method on `TransactionSearchParams`, something like 

```typescript
hasUserSelectedFilters () {
  const filters = this.getQueryParams()
  filters.delete('page')
  return userSelectedFilters.toString().length > 0
}
```

then do 
```typescript
const showCsvDownload = transactionCountWithinRange || (
    transactionSearchParams.hasUserSelectedFilters() &&
    results.total > LEDGER_TRANSACTION_COUNT_LIMIT
)
```

alternatively could call the method something like `isFilteredSearch`/`isFocusedSearch` maybe?

test/cypress/integration/simplified-account/transaction/transaction-index.cy.ts
DomBelcher: doesn't matter too much, I'm looking at breaking this file up a bit anyway as it's quite big

src/controllers/simplified-account/services/transactions/transaction-list.controller.ts
ralphm10: I like the suggestion

DomBelcher (approved):
LGTM

PR #4952, opened 2026-06-23
PP-15440 Add missing CSS class to email links

kbottla:
## What
- Added missing `govuk-link` class to email links


DomBelcher (approved):
LGTM

PR #4957, opened 2026-06-25
PP-15481 Fix transaction search not processing state array as csv string

DomBelcher:
## What

PP-15481

- Fix state filters not being set correctly when using pagination
- remove erroneous inclusion of old pagination template from transactions index template
  - this is inconsequential as the pagination template is enclosed in an `{% if hasPaginationLinks %}` statement, which will always be false, since `hasPaginationLinks` is not set by the new controller

### How/Why

Arrays can be sent in several ways in query strings. 

By default express uses the inbuilt node [querystring.parse](https://nodejs.org/api/querystring.html#querystringparsestr-sep-eq-options) function, which expects arrays as repeated keys. 

The `URLSearchParams` constructor accepts an object, but only as `Record<string,string>`, so if an object is passed with one of the values being a `string[]`, this is stringified using `Array.prototype.toString()`, which converts it to a comma separated list. 

This CSV cannot then be passed back to `querystring.parse` to return an Array, it will parse as the comma-separated string. 

For example, if a request is made with the query string `a=1&a=2&a=3`, this will be parsed in `req.query` as `a: ['1', '2', '3']`. If `req.query` is then passed to `new URLSearchParams()`, this will return an instance of `URLSearchParams` with `'a' => '1,2,3'`. Therefore if `urlSearchParams.toString()` is used to produce a query string for a URL, it will yield `a=1,2,3`, which will then be parsed as `a: '1,2,3'` in the subsequent query.

This causes a bug in the transactions search, where `URLSearchParams` is used to create query strings for the links in pagination. As such, when searching with several state filters selected, the pagination does not work as expected, since the pagination link will be a comma-separated list of the selected filters, instead of using repeated keys. This causes the selected state filter to default back to `All` when using the pagination links.

This is fixed by accepting inputs of `string[] | string` when parsing arrays in the search, with `string` inputs being parsed as comma-separated lists. Because none of the state filters contain commas, it is safe to do `stateFilter.split(',')`, as in the case where there is only one state filter selected, this will yield `[stateFilter]`.


ralphm10 (approved):
LGTM - I have manually tested and verified that status filters persist when navigating through pages. 

(This needs another approval as I pushed to fix the merge conflicts)

kbottla (approved):

PR #4962, opened 2026-06-29
PP-15526 Redirect to new transactions pages on feature flag

DomBelcher:
## What

PP-15526

When the `transactions` feature flag is enabled, redirect any old transactions pages to the new ones

All Cypress tests for the old transactions pages are skipped, as these routes are no longer accessible. New tests are added to test the redirects.

### DIFF
--- a/src/middleware/simplified-account/transaction-redirect.middleware.ts
+++ b/src/middleware/simplified-account/transaction-redirect.middleware.ts
@@ -1,0 +1,97 @@
+import { NextFunction, Request, Response } from 'express'
+import formatServiceAndAccountPathsFor from '@utils/simplified-account/format/format-service-and-account-paths-for'
+import { Features } from '@root/config/features'
+import formatPathFor from '@utils/replace-params-in-path'
+import { NotFoundError } from '@root/errors'
+import Ledger from '@services/clients/ledger.client'
+import { userServicesContainsGatewayAccount } from '@utils/permissions'
+import { TransactionData } from '@models/transaction/dto/Transaction.dto'
+import { getGatewayAccountById } from '@services/gateway-accounts.service'
+import Service from '@models/service/Service.class'
+import { ServiceData } from '@models/service/dto/Service.dto'
+import GatewayAccount from '@models/gateway-account/GatewayAccount.class'
+import { GatewayAccountData } from '@models/gateway-account/dto/GatewayAccount.dto'
+import createLogger from '@utils/logger'
+
+const logger = createLogger('transaction-redirect.middleware.ts')
+
+export function transactionRedirect(path: string) {
+  if (!Features.isEnabled(Features.TRANSACTIONS)) {
+    return (req: Request, res: Response, next: NextFunction) => {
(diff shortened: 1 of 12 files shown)


src/middleware/simplified-account/transaction-redirect.middleware.ts
```diff
@@ -0,0 +1,97 @@
...
+      throw new NotFoundError('Charge ID not found when attempting to redirect to all-service-transactions detail page')
+    }
+
+    let redirectPath
+    try {
+      const charge = (await Ledger.transactionWithAccountOverride(req.params.chargeId)) as TransactionData
+      if (!userServicesContainsGatewayAccount(charge.gateway_account_id, req.user)) {
+        return next(new NotFoundError('User does not have access to gateway account for transaction'))
+      }
+      const account = await getGatewayAccountById(charge.gateway_account_id as unknown as number)
```
Pippa101, line 83: Is there a way we can avoid double casting? What do we expect `gateway_account_id` to be? If it's a string representation of a number can we use `Number(charge.gateway_account_id)`

test/cypress/integration/all-service-transactions/all-service-transactions-no-autosearch.cy.js
```diff
@@ -47,56 +47,112 @@ const testTransactions = [
...
+        expect(response.headers.location).to.eq('/transactions/live/nosearch')
+      })
+    })
+  })
 
-    cy.get('.govuk-breadcrumbs').within(() => {
-      cy.get('.govuk-breadcrumbs__list-item').should('have.length', 2)
-      cy.get('.govuk-breadcrumbs__list-item').eq(1).contains('Transactions for all services')
-      cy.get('.govuk-breadcrumbs__list-item').eq(1).find('.govuk-tag').should('have.text', 'Test')
+  describe.skip('skipping old tests', () => {
```
Pippa101, line 111: Do we have a ticket for deleting these tests once the transactions feature is enabled?

DomBelcher: It'll go along with deleting all the old transactions code, there might not be a ticket for it yet but I'll make one when the time comes

src/middleware/simplified-account/transaction-redirect.middleware.ts
DomBelcher: It's all stringified in the request anyway, so it doesn't actually matter if it's a proper number or just a stringified one.

It might make more sense to have the client & service methods accept `number | string` instead, but this code shouldn't be around for too long so I just did this hack to workaround the annoyance of account IDs sometimes being returned as strings, and sometimes as numbers

Pippa101: Okay if it's getting deleted anyway that's fine

Pippa101 (dismissed):
LGTM

Pippa101 (approved):
LGTM

PR #4963, opened 2026-06-29
PP-15481 Polyfill service ID for All Service Transaction links if missing

DomBelcher:
## What

PP-15481

Old transactions may not have a `service_id` present in the Ledger database, so `service_id` may be `undefined` in some cases. It therefore cannot be used to create links to transactions, which is done in the All Service Transactions list, as the service ID must be known, since transactions cannot be searched by transaction ID alone. 

This adds a polyfill for determining the service ID from the relevant gateway account if the service ID is missing on the transaction.


Pippa101 (approved):
LGTM

PR #4964, opened 2026-06-29
PP-15481 Fix no results text, ensure permissions for CSV download

DomBelcher:
## What

PP-15481

- Fix "There are no results for the filters you used." text showing next to filters on All Service Transactions search page
  - this now shows under the search menu, as intended
- Add permission check for CSV download
  - in practice, all roles have this permission, however this check is important should this change in the future
- Remove some code in Transaction and All Service Transaction list templates which was erroneously copied from old pages

Screenshot 2026-06-29 at 17 36 23


Pippa101 (approved):
LGTM

PR #4966, opened 2026-06-30
PP-15481 Fix missing fee and net amount columns in All Service Transactions list

DomBelcher:
## What

PP-15481

The `Fee` and `Net amount` columns should show in the All Service Transactions list if any of the user's services are Stripe services, since these values exist on Stripe transactions. Services may also have switched from Stripe to Worldpay, so these columns should be shown if any of the transactions are Stripe transactions, even if the user no longer has a Stripe service (due to it switching)

- Fix a bug where `Fee` and `Net amount` columns were not showing
  - this occurred as the code for determining what payment providers a user had in `ViewMode.class.ts` was erroneously returning an empty array in all cases.
   - this happened because the callback in `Array.prototype.filter` is not required to return a `boolean`, instead it can return any truthy or falsy value. There was a `return` statement missing in the filter callback, so it was always returning `undefined`, which is falsy. 
  
- Add several new test fixtures for the `GatewayAccount` class, and all relevant classes used within it
- Restructure `EmailNotifications` classes as `EmailNotifications` with `EmailNotificationSetting` classes used within
  - make these names consistent with the DTO and fixture classes

Screenshot 2026-06-30 at 16 04 40


src/controllers/simplified-account/home/all-service-transactions/all-service-transactions.controller.ts
```diff
@@ -65,6 +63,11 @@ async function get(
     }
   }
 
+  // services could have switched from Stripe to Worldpay and have old Stripe transactions
+  const isStripe =
```
kbottla: Will this lead to inconsistent behaviour if one page of results has Stripe transactions and another page doesn't?

DomBelcher: Probably yes, although I'm not sure what we can do about it. The only options are either to do this, to never show the additional columns for services which aren't currently on Stripe, or to always show those columns if any of the relevant gateway accounts have a retired Stripe credential

DomBelcher: I don't think the `searchGatewayAccounts` endpoint in Connector returns credentials though, so it's not easy to check if there are retired Stripe credentials

kbottla (approved):

PR #4968, opened 2026-07-02
PP-15481 Fix CSV download link on payouts page & modify default Transactions search

DomBelcher:
## What

PP-15481

With the old Transactions URLs redirecting to the new ones, the CSV download link in the "Payments to your bank account" does not work, as it does not change to the new Transactions URL with the `transactions` feature flag. This fixes the link to correctly link to the new CSV download URL when the `transactions` feature flag is enabled.

This also modifies the way default Ledger search parameters are set, so that CSV downloads do not add additional search parameters, such as a `from_date` or `limit_total`. These are instead now configured by the controller setting the search parameters.

The `transaction-index` and `all-service-transaction` controllers set a default `from_date` and `to_date` for the `last-12-months` date filter period. This has been amended to make the `to_date` the end of the current day, rather than the current time. This is primarily to make Cypress testing easier, as it is not possible (with the current configuration) to stub the date in the running app used by Cypress. This is not important for the functionality, as Ledger will accept a `to_date` that is in the future. 

The code for `TransactionSearchParams` is a little clunky, and could likely be improved, but this should be left for future work.

### DIFF
--- a/src/controllers/simplified-account/home/all-service-transactions/all-service-transactions.controller.ts
+++ b/src/controllers/simplified-account/home/all-service-transactions/all-service-transactions.controller.ts
@@ -22,6 +22,7 @@ import {
   LEDGER_TRANSACTION_COUNT_LIMIT,
   MAX_TRANSACTIONS_PER_PAGE,
 } from '@controllers/simplified-account/services/transactions/constants'
+import { Period } from '@utils/simplified-account/services/dashboard/datetime-utils'
 
 async function get(
   req: AuthenticatedRequest & { viewMode: ViewMode },
@@ -49,7 +50,7 @@ async function get(
     req.query,
     true,
     MAX_TRANSACTIONS_PER_PAGE
-  )
+  ).withDefaultDateFilter(Period.LAST_12_MONTHS)
 
   let cardTypes: CardType[]
   let results: TransactionSearchResults
--- a/test/cypress/integration/payouts/payouts-list.cy.js
+++ b/test/cypress/integration/payouts/payouts-list.cy.js
@@ -32,7 +32,7 @@ describe('Payout list page', () => {
     cy.setEncryptedCookies(userExternalId)
   })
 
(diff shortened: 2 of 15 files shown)


src/controllers/simplified-account/home/all-service-transactions/all-service-transactions.controller.ts
```diff
@@ -49,7 +50,7 @@ async function get(
     req.query,
     true,
     MAX_TRANSACTIONS_PER_PAGE
-  )
+  ).withDefaultDateFilter(Period.LAST_12_MONTHS)
```
kbottla, line 53: Does this affect the transactions downloaded from the `payments into your bank account` report, which won't have data filters?

DomBelcher: That's the reason I've done this, `withDefaultDateFilter` will only set date filters if no `dateFilter`, `fromDate` or `toDate` has been set. I needed a way of having the date filters set by default in searches, but not in downloads.

To be honest, I think the whole thing would be cleaner as a proper builder, but I wasn't sure about redoing the whole thing here.

test/cypress/integration/payouts/payouts-list.cy.js
```diff
@@ -11,34 +11,38 @@ const userAndGatewayAccountStubs = [
...
+  ]),
 ]
 
 describe('Payout list page', () => {
   beforeEach(() => {
     cy.setEncryptedCookies(userExternalId)
   })
 
-  it('should correctly display payouts given a successful response from Ledger', () => {
+  it.only('should correctly display payouts given a successful response from Ledger', () => {
```
kbottla: only?

DomBelcher: Good catch, will remove.

kbottla (approved):

PR #4969, opened 2026-07-06
PP-15481 Fix missing MOTO and Fee columns in CSV downloads

DomBelcher:
## What

PP-15481

 - Fix `moto_header` and `fee_headers` params missing from requests to Ledger `/v1/transaction` search endpoint
   - this resulted in some columns being missing from CSV downloads
 - Migrate `TransactionSearchParams` to a builder-like pattern, making its use in controllers cleaner.
 - Use `ViewMode` in All Service Transactions CSV download to determine whether to include MOTO and fee headers, since all a user's gateway accounts are already retrieved by the view mode middleware
 - Add tests for  `TransactionSearchParams`

This also adds some Typescript stubs for Cypress tests which are not currently used. These were to be used in a CSV download Cypress test, however as the responsibility for correctly sending the CSV data is largely handled by Ledger, there is not much value gained from a Cypress test which stubs out the relevant Ledger request.

The Typescript stubs have been left in as they will likely be useful in the future. Uses of existing Javascript stubs could be migrated over to these at any time.

### DIFF
--- a/src/controllers/simplified-account/home/all-service-transactions/all-service-transaction-csv-download.controller.ts
+++ b/src/controllers/simplified-account/home/all-service-transactions/all-service-transaction-csv-download.controller.ts
@@ -5,25 +5,15 @@ import { downloadCsv } from '@services/transactions.service'
 import logger from '@utils/logger'
 import PaymentProviders from '@models/constants/payment-providers'
 import date from '@utils/dates'
-import { findGatewayAccountsByService } from '@services/gateway-accounts.service'
+import { ViewMode } from '@models/view-mode/ViewMode.class'
 const LOGGER = logger(__filename)
 
-async function get(req: AuthenticatedRequest, res: express.Response<unknown, { flash?: Record<string, string[]> }>) {
-  const modeFilter = req.params.modeFilter === 'test' ? 'test' : 'live'
-
-  const userServiceExternalIds = req.user.serviceRoles
-    .filter((serviceRole) => serviceRole.hasPermission('transactions-download:read'))
-    .map((serviceRole) => serviceRole.service)
-    .map((service) => service.externalId)
-
-  const gatewayAccounts = await findGatewayAccountsByService(userServiceExternalIds, modeFilter)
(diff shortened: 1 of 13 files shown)


src/controllers/simplified-account/home/all-service-transactions/all-service-transaction-csv-download.controller.ts
```diff
@@ -5,25 +5,15 @@ import { downloadCsv } from '@services/transactions.service'
...
-  const gatewayAccounts = await findGatewayAccountsByService(userServiceExternalIds, modeFilter)
-  const gatewayAccountIds = gatewayAccounts.map((gatewayAccountData) => gatewayAccountData.id)
-  const transactionSearchParams = TransactionSearchParams.fromSearchQuery(gatewayAccountIds, req.query, false)
-
-  if (gatewayAccounts.some((gatewayAccount) => gatewayAccount.paymentProvider === PaymentProviders.STRIPE)) {
-    transactionSearchParams.feeHeaders = true
-  }
-  transactionSearchParams.motoHeader = gatewayAccounts.some((gatewayAccount) => gatewayAccount.allowMoto)
+async function get(req: AuthenticatedRequest & { viewMode: ViewMode }, res: express.Response) {
+  const isMoto = Array.from(req.viewMode.gatewayAccounts.values()).some((gatewayAccount) => gatewayAccount.allowMoto)
```
kbottla, line 12: Does this filter by the `transactions-download:read` permission? Most roles should have this permission, but it would be good to explicitly filter with the relevant permissions (if not already) so it is future-proof for any roles added without download permissions.

DomBelcher: It's handled in the middleware, this controller has that permission applied, so it will filter for services with that permission

https://github.com/alphagov/pay-selfservice/blob/021d9df15f3744ecd8fc0294041fb4343fb0aa46/src/routes.js#L228-L234

kbottla (approved):
LGTM!

PR #4971, opened 2026-07-07
PP-15481 Ensure filters correctly persist from search to CSV download

DomBelcher:
## What

PP-15481

When loading the transaction search, the default `last-12-months` date filter is selected, however the CSV download link that is generated does not include any filters. This means that a CSV of all transactions is generated if the user clicks the CSV download link on first visiting the page.

This fixes this behaviour, to ensure filters are always applied to the CSV download, and that these filters will successfully recreate the same search query.

This also changes the way in which it is checked if filters are applied (for the purpose of showing/hiding the CSV download button) to simply counting the filters directly

### DIFF
--- a/src/models/transaction/TransactionSearchParams.class.ts
+++ b/src/models/transaction/TransactionSearchParams.class.ts
@@ -12,10 +12,11 @@ import {
 } from '@utils/simplified-account/services/transactions/status-filters'
 import { parseTransactionSearchDateTime } from '@utils/time/parse-date-time'
 import type { DateTime } from 'luxon'
-import { TRANSACTION_SEARCH_DATE_FORMAT } from '@utils/time/time-formats'
+import { TRANSACTION_SEARCH_DATE_FORMAT, TRANSACTION_SEARCH_TIME_FORMAT } from '@utils/time/time-formats'
 import { TimeConstants } from '@utils/time/time-constants'
+import { omit } from 'lodash'
 
-interface TransactionSearchQuery {
+export interface TransactionSearchQuery {
   cardholderName?: string
   lastDigitsCardNumber?: string
   metadataValue?: string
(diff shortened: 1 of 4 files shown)


src/models/transaction/TransactionSearchParams.class.ts
```diff
@@ -169,17 +192,13 @@ export class TransactionSearchParams {
...
+    return this.filterCount === undefined ? false : this.filterCount > 0
   }
 
   isUnfilteredAllTimeSearch() {
-    const filters = this.getQueryParams()
-    filters.delete('page')
-    filters.delete('jsEnabled')
-    return filters.size === 1 && filters.get('dateFilter') === Period.ALL_TIME
+    return this.filterCount === undefined
+      ? false
```
ralphm10: as above, can this be omitted?

src/models/transaction/TransactionSearchParams.class.ts
```diff
@@ -169,17 +192,13 @@ export class TransactionSearchParams {
   }
 
   hasUserSelectedFilters() {
-    const filters = this.getQueryParams()
-    filters.delete('page')
-    filters.delete('jsEnabled')
-    return filters.size > 0
+    return this.filterCount === undefined ? false : this.filterCount > 0
```
ralphm10: can this be simplified to `return this.filterCount > 0` - as if it was undefined it would still return false?

ralphm10 (commented):
LGTM - just a minor comment which might improve readability. 

I have verified the change and can confirm that the CSV download now includes only 12 months of transactions after first navigating to the page (as opposed to all, as previously)

src/models/transaction/TransactionSearchParams.class.ts
DomBelcher: This was just to keep the type system happen but thinking about it, it makes sense to have `filterCount` be initialised as zero, so then there's no need for it to be possibly `undefined`

ralphm10 (approved):

PR #4975, opened 2026-07-09
PP-15481 Set default date filter in Transactions search to 'all time'

DomBelcher:
## What

PP-15481

 - set the default date filter in Transactions list and All Service Transactions list to `all-time`
 - this shows as "Show all" in the dropdown list, with the from date and to date fields empty


kbottla (approved):
LGTM!

PR #4982, opened 2026-07-14
PP-15548 Add client method to get Adyen account setup

DomBelcher:
## What

PP-15548
- Add method to `ConnectorClient` to get Adyen account setup
  - method should call `GET /v1/api/service/{serviceExternalId}/account/{accountType}/adyen-setup/{credentialExternalId}`
- add Pact tests for get Adyen account setup endpoint for various scenarios
  - this uses the Pact v4 DSL, in contrast to existing Pact tests, which use the v2 DSL, however still uses the version 2 specification, since it is not possible to mix spec versions 
- enable Typescript Pact tests
- update Pact publishing to correctly tag published pacts with PR number


src/models/gateway-account/dto/AdyenAccountSetup.dto.ts
```diff
@@ -0,0 +1,10 @@
+// TODO update when we know the Adyen KYC tasks
+type AdyenAccountSetupTaskNames = never
+
+export interface AdyenAccountSetupData {
+  service_id: string
+  credential_external_id: string
+  tasks: Record<AdyenAccountSetupTaskNames, AdyenAccountSetupStatus>
```
ralphm10: In the response from connector, tasks will be in the format of:
```
"tasks": {
    "bank_account": {
      "status": "COMPLETED"
    },
    "director": {
      "status": "NOT_STARTED"
    },
    // more tasks here...
```
So I think this would need updating to `Record<AdyenAccountSetupTaskNames, Record<String, AdyenAccountSetupStatus>>`... or it's own object definition if that is cleaner

src/services/clients/pay/ConnectorClient.class.ts
```diff
@@ -169,6 +171,18 @@ class ConnectorClient extends BaseClient {
...
       },
 
+      adyenSetup: {
+        get: async (serviceExternalId: string, accountType: string, credentialExternalId: string) => {
+          const path = '/v1/api/service/{serviceExternalId}/account/{accountType}/adyen-setup/{credentialExternalId}'
+            .replace('{serviceExternalId}', encodeURIComponent(serviceExternalId))
+            .replace('{accountType}', encodeURIComponent(accountType))
+            .replace('{credentialExternalId}', encodeURIComponent(credentialExternalId))
+
+          const response = await this.get<AdyenAccountSetupData>(path, 'get stripe account onboarding progress')
```
ralphm10: update to `Adyen`

src/models/gateway-account/dto/AdyenAccountSetup.dto.ts
```diff
@@ -0,0 +1,10 @@
+// TODO update when we know the Adyen KYC tasks
+type AdyenAccountSetupTaskNames = never
```
ralphm10: these have been defined as an an enum in [`AdyenAccountSetupTask`](https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gatewayaccount/model/AdyenAccountSetupTask.java) on connector if we wanted to use the same values (although understand they may change)

src/services/clients/pay/ConnectorClient.class.ts
DomBelcher: oops

src/models/gateway-account/dto/AdyenAccountSetup.dto.ts
DomBelcher: ah my mistake, i'll fix that

src/models/gateway-account/dto/AdyenAccountSetup.dto.ts
DomBelcher: Ok, I think I misunderstood how the endpoint worked. It returns the full list of tasks with the given status, with `NOT_STARTED` as the default value if there is no row for that gateway account/task pair in the `gateway_accounts_adyen_setup` table, right?

If so, then yeah this should define all the task names here, as they will always be returned. 

It may make our lives a little more difficult with Pact tests, where we may need to update the client & server in a specific order to ensure compatibility if we have to add or remove tasks, but we can deal with that should it be necessary.

ralphm10: yes correct... if the tasks table is empty it will respond with:

```
"tasks": {
    "company_number": {
      "status": "NOT_STARTED"
    },
    "responsible_person": {
      "status": "NOT_STARTED"
    },
    "vat_number": {
      "status": "NOT_STARTED"
    },
    "director": {
      "status": "NOT_STARTED"
    },
    "government_entity_document": {
      "status": "NOT_STARTED"
    },
    "bank_account": {
      "status": "NOT_STARTED"
    },
    "organisation_details": {
      "status": "NOT_STARTED"
    }
  }
```
Or whatever the tasks are, this is populated by looping through the task enum

ralphm10 (approved):
LGTM

PR #4977, opened 2026-07-10
PP-15481 Restructure transaction directories

DomBelcher:
## What

PP-15481
- Restructure Transactions directories to group everything together
- deduplicate some code in Transactions list pages with nunjucks partials

PR #4983, opened 2026-07-16
PP-15611 Fix All Service Transactions nosearch page searching on wrong route

DomBelcher:
## What

PP-15611
Refactoring in #4977 inadvertently removed the modified search URL on the All Service Transactions nosearch page, meaning that submitting the form would send the request on the `/nosearch` URL, which does not perform the search. This meant that when a search timed out, the user would not be able to refine the search.

This fixes this bug by conditionally adding an `action` attribute to the `<form>` element if a `searchLink` property is present in the nunjucks template

This also adds Cypress tests for the nosearch page, verifying that is searches on the correct URL


ralphm10 (approved):
LGTM

PR #4984, opened 2026-07-24
PP-15407 package lock bug fix

iqbalgds:
## What

PP-**15407**

- Errors with package-lock.json
- Currently it contains an unresolved conflict, meaning the lockfile is in an inconsistent state. 
- There is a peer dependency conflict between esbuild-sass-plugin and esbuild:
- Upgrading ESBuild would fixed the issue - however, this generated some issues to be fixed:
  - Unit tests
  - TypesScript errors had to be fixed.

### How
- Make sure all unit tests and cypress tests are passing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -88,7 +88,7 @@
         "cypress": "^14.0.3",
         "cypress-axe": "^1.6.0",
         "dotenv": "16.3.1",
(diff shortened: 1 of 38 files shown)


DomBelcher (approved):
Nice work!

PR #4998, opened 2026-08-13
PP-15686 Enable Adyen PSP selection via service feature flag

christiandutton-mt:
## What

[PP-15686
](https://payments-platform.atlassian.net/browse/PP-15686)

This PR updates logic for determining which PSP can be selected as GOV.UK Pay’s Provider in the request to go live journey to be conditional on either the environment wide feature, or govuk_psp_is_adyen feature being set to enabled. This PR follows on the work done to add this feature to admin users [here](https://github.com/alphagov/pay-adminusers/pull/3342). 

- Added serviceFeatures to Service class, matching structure from admin-users 
- Updated controllers to either check for existing environment wide feature OR govuk_psp_is_adyen being set to enabled to configure whether adyen can be selected as PSP. 
- Updated njk files to use above logic
- Removed `enable_go_live_requests_for_adyen` from experimental features list in env so we just deal with service feature in testing
- Test utils and features updated
- Tests updated to now go through options as follows: 
             - Service feature is disabled in choose-how-to-process-payments - stripe displays, adyen doesn't
             - Service feature is enabled in choose-how-to-process-payments - adyen displays
             - Service feature is disabled and CHOSEN_PSP_STRIPE in agreements - stripe displays, adyen doesn't
             - Service feature is enabled and CHOSEN_PSP_ADYEN in agreements - adyen displays

### DIFF
--- a/src/models/service/Service.class.ts
+++ b/src/models/service/Service.class.ts
@@ -1,4 +1,4 @@
-import { ServiceData } from '@models/service/dto/Service.dto'
+import { ServiceData, ServiceFeatures } from '@models/service/dto/Service.dto'
 
 class Service {
   readonly id: number
@@ -25,7 +25,7 @@ class Service {
   readonly agentInitiatedMotoEnabled: boolean
   readonly defaultBillingAddressCountry?: string
   readonly takesPaymentsOverPhone: boolean
-  readonly features?: string[]
+  readonly serviceFeatures?: ServiceFeatures
 
   constructor(serviceData: ServiceData) {
     this.id = serviceData.id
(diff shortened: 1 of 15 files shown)


src/models/service/Service.class.ts
```diff
@@ -25,6 +25,7 @@ class Service {
   readonly agentInitiatedMotoEnabled: boolean
   readonly defaultBillingAddressCountry?: string
   readonly takesPaymentsOverPhone: boolean
+  readonly serviceFeatures?: ServiceFeatures
```
DomBelcher: This should always be present (I think)

DomBelcher (approved):
Nice work

PR #5001, opened 2026-08-19
PP-15765: remove npm from deployment image

AndrzejProchyra:
This PR removes npm, corepack and yarn from the final image. The following are removed:

- `/usr/local/lib/node_modules/npm` - the global npm installation
- `/usr/local/lib/node_modules/corepack` - the global corepack installation
- `/usr/local/bin/npm` - symlink to `/usr/local/lib/node_modules/npm/bin/npm-cli.js`
- `/usr/local/bin/npx` - symlink to `/usr/local/lib/node_modules/npm/bin/npx-cli.js`
- `/usr/local/bin/corepack` - symlink to `/usr/local/lib/node_modules/corepack/dist/corepack.js`
- `/opt/yarn-*` - global yarn installation(s)
- `/usr/local/bin/yarn` - symlink to `/opt/yarn-<current-version>/bin/yarn`
- `/usr/local/bin/yarnpkg` - symlink to `/opt/yarn-<current-version>/bin/yarnpkg`

Note that yarn has been [removed from the Node.js 26+ images
](https://github.com/nodejs/docker-node/issues/2264), and corepack is not included from Node 25.

Additionally:
* Uses a WORKDIRs with a more descriptive name in the "builder" stage, for clarity

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,28 @@
 FROM node:22.23.2-alpine@sha256:c610fcdfb1d5b4740dd70c284ed3cb16bb857e0f7166196e36a5501df7a3aa32 AS base
 
-WORKDIR /app
 RUN apk upgrade --no-cache \
     && apk add --no-cache tini
 
-# Upgrade npm — if updating the Node.js version, check if this
-# is still necessary and make sure it never downgrades npm
-RUN npm install -g npm@11.18.0
-
 FROM base AS builder
 
-COPY . .
+WORKDIR /build-stage
+COPY . ./
 RUN npm ci --quiet
 RUN npm run compile
 
 FROM base AS final
 
-COPY --from=builder /app/dist .
+RUN rm -rf /usr/local/lib/node_modules/npm \
+    /usr/local/lib/node_modules/corepack \
+    /usr/local/bin/npm \
+    /usr/local/bin/npx \
+    /usr/local/bin/corepack \
+    /opt/yarn-* \
+    /usr/local/bin/yarn \
+    /usr/local/bin/yarnpkg
+
+WORKDIR /app
+COPY --from=builder /build-stage/dist ./
 ENV PORT=9000
 EXPOSE 9000
 ENTRYPOINT ["tini", "--"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,8 +9,8 @@ FROM base AS builder
 RUN npm install -g npm@11.18.0
 
 WORKDIR /build-stage
-COPY . ./
 RUN npm ci --quiet
+COPY . ./
 RUN npm run compile
 
 FROM base AS final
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,23 +1,33 @@
...
 
-COPY --from=builder /app/dist .
+RUN rm -rf /usr/local/lib/node_modules/npm \
+    /usr/local/lib/node_modules/corepack \
+    /usr/local/bin/npm \
+    /usr/local/bin/npx \
+    /usr/local/bin/corepack \
+    /opt/yarn-* \
+    /usr/local/bin/yarn \
+    /usr/local/bin/yarnpkg
```
benjamb: While npm will indeed no longer be accessible in the final image with these changes, we won't benefit from the reduction in image size as they will still be present within the base layer.

If we did want that, following https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#smaller-images-without-npmyarn would be better, if a little more involved. Perhaps not really worth the effort though, it's a shame they don't publish a node-only image.

AndrzejProchyra: Yes, this is true. If we followed the Best Practices guide we would need some way of synchronising the version of Alpine used for the node image and the alpine base image. The example does this through interpolation, setting `ARG ALPINE_VERSION=3.23` and using it in both `FROM` lines. As I understand, Dependabot does not support this (although Renovate does).

Is there a better way of achieving this? It is clearly better to "copy only what you need", rather than "delete what you don't want" (which doesn't shrink the image, as you say).

benjamb: Hmm, yeah. I forget about that. We should probably track all of these dependabot pain points as we encounter them.

I'm not aware of another way of achieving this, so your approach is likely the best we can do for now.

Dockerfile
```diff
@@ -1,23 +1,33 @@
 FROM node:22.23.2-alpine@sha256:c610fcdfb1d5b4740dd70c284ed3cb16bb857e0f7166196e36a5501df7a3aa32 AS base
 
-WORKDIR /app
 RUN apk upgrade --no-cache \
     && apk add --no-cache tini
```
benjamb: We don't really need this in the builder, but it doesn't really hurt to leave it in the base.

AndrzejProchyra: You are right. But I think, on the whole, it is better to be intentional and explicit. Here we are saying "This is the *base* stage. It is important that both of the other stages have access to `tini`". And this is not the case.

AndrzejProchyra:
@benjamb It occurs to me that if the `COPY` instruction comes after `RUN npm ci`, the `npm ci` can be cached, unless `package.json` changes. But we do need `package.json` and `package-lock.json` and `.npmrc` first...

marcotranchino (approved):
LGTM 👍

PR #5003, opened 2026-08-21
PP-15743 Implement "Organisation's bank details" task skeleton

christiandutton-mt:
## What

[PP-**[15743]**](https://payments-platform.atlassian.net/browse/PP-15749)

- Added get and post controllers for bank-details and added path to routes.
- Added view in bank-details.njk matching design linked in ticket

For now as this is a skeleton the form does not submit anything to Adyen and there is no validation. That will come in later tickets. 

I will put cypress tests for this work in another PR. 

image

### DIFF
--- a/src/controllers/simplified-account/settings/switch-psp/switch-to-adyen/bank-details.controller.test.ts
+++ b/src/controllers/simplified-account/settings/switch-psp/switch-to-adyen/bank-details.controller.test.ts
@@ -4,10 +4,11 @@ import { PaymentProvider } from '@models/constants/payment-provider'
 import { UserFixture } from '@test/fixtures/user/user.fixture'
 import { ServiceFixture } from '@test/fixtures/service/service.fixture'
 import sinon from 'sinon'
-const formatServiceAndAccountPathsFor = require('@utils/simplified-account/format/format-service-and-account-paths-for')
-const paths = require('@root/paths')
+import formatServiceAndAccountPathsFor from '@utils/simplified-account/format/format-service-and-account-paths-for'
+import paths from '@root/paths'
 
 const SERVICE_EXTERNAL_ID = 'service123abc'
+const SERVICE_TYPE = 'live'
 const serviceFixture = new ServiceFixture({
   externalId: SERVICE_EXTERNAL_ID,
 })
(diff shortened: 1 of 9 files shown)


src/controllers/simplified-account/settings/switch-psp/switch-to-adyen/bank-details.controller.test.ts
DomBelcher, line 1: Could you make this a Typescript file? It can be fiddly sometimes with test files, but worth it to have everything as Typescript if possible

src/views/simplified-account/settings/switch-psp/switch-to-adyen/bank-details.njk
```diff
@@ -0,0 +1,72 @@
...
+
+{% set settingsPageTitle = 'Organisation’s bank details' %}
+
+{% block settingsContent %}
+  <form id="bank-details-form" autocomplete="off" method="post" action="{{ submitLink }}" class="govuk-!-margin-top-3">
+    <input id="csrf" name="csrfToken" type="hidden" value="{{ csrf }}" />
+    {%
+      call govukFieldset({
+        legend: {
+          text: "Organisation’s bank details",
```
christiandutton-mt, line 11: I think I need to change the ' to ’ here. Will double check in next PR with cypress tests in.

src/views/simplified-account/settings/switch-psp/switch-to-adyen/bank-details.njk
```diff
@@ -0,0 +1,72 @@
+{% extends "simplified-account/settings/settings-layout.njk" %}
+
+{% set settingsPageTitle = 'Organisation’s bank details' %}
+
+{% block settingsContent %}
+  <form id="bank-details-form" autocomplete="off" method="post" action="{{ submitLink }}" class="govuk-!-margin-top-3">
```
DomBelcher: Not important since it works either way, but you probably don't need to set the `action`, as it will POST to the current URL by default (I think)

DomBelcher (commented):
Nice work.

Could you add a screenshot to the PR description as well?

src/views/simplified-account/settings/switch-psp/switch-to-adyen/bank-details.njk
christiandutton-mt: yep you're right thanks, removed it from the controller too then!

DomBelcher (approved):
Nice work

PR #5004, opened 2026-08-24
PP-15746 Add Cypress tests for scenarios where service is not switching to Adyen

DomBelcher:
## What

PP-15746
Add Cypress tests for scenarios where service is not switching to Adyen

- Add Cypress test for service switching to a different provider
- Add Cypress test for service already live with Adyen


ralphm10 (approved):
LGTM

PR #5006, opened 2026-08-24
PP-15742 - Reason for payments - controller

iqbalgds:
## What

PP-**15742**

- Part of setting up the skeleton page for reason for payments task.
- This is just adding the controllers.
- Future PRs will implement more of the skeleton page.

### How
- Just check the new unit test is running successfully.


DomBelcher (approved):
LGTM

PR #5008, opened 2026-08-25
PP-15791 Add Local Dockerfile

DomBelcher:
## What

PP-15791

- Add local.Dockerfile for building docker images for development using pay local
- This will be picked up by the Pay CLI when running with local Selfservice


DomBelcher:
Closing as the CLI has been changed to provide a different solution

PR #5013, opened 2026-08-28
PP-15752 reason for payments - complete skeleton page

iqbalgds:
## What

PP-**15752**

image

- This PR add the files required to complete the skeleton page for 'reason for taking payments'.
- The Cypress tests are also added as part of this PR.
- Updates to the controllers:
  - Fix the paths used.
  - Move controllers to the correct folder under settings

### How
- Find the service where it is in the process of switching to Adyen
- From the task list - make sure you can access the `Reason for Taking Payments` page from there
- When you press the `confirm and submit` button, it should take you back to the task list
- Make sure Cypress tests are working.


DomBelcher (approved):
Nice work

PR #5019, opened 2026-09-02
PP-15836 add Dependabot exclude cooldown

SandorArpa:
## What

Exclude pay-js-commons from the cooldown period of dependabot updates.

N/A


foxmaltas-gds (approved):
LGTM

PR #5023, opened 2026-09-03
PP-15750 Implement persitence for responsible person tasks

ralphm10:
## What

PP-15750

- Add persistence to answers when navigating through pages
- Make answers on review page dynamic
- Update back links to be conditional
- Update unit tests
- Redirect to migration task list page if session is empty

Cypress test to be implemented in next PR


src/controllers/simplified-account/settings/switch-psp/switch-to-adyen/switch-to-adyen.controller.ts
```diff
@@ -1,9 +1,11 @@
 import { ServiceRequest, ServiceResponse } from '@utils/types/express'
 import { response } from '@utils/response'
 import { AdyenTasks } from '@models/task-workflows/AdyenTasks.class'
+import { ResponsiblePersonSession } from '../../adyen-details/responsible-person/constants'
 
 function get(req: ServiceRequest, res: ServiceResponse) {
   const adyenTasks = AdyenTasks.forProviderSwitching(req.service, req.account)
+  ResponsiblePersonSession.clear(req)
```
christiandutton-mt, line 8: If you clear here, does that mean if you are half way through the flow and view the task list that you lose progress? Could a clear work on the final submit instead?

iqbalgds:
Looks good to me - going to check if anyone else wants to review.

src/controllers/simplified-account/settings/switch-psp/switch-to-adyen/switch-to-adyen.controller.ts
ralphm10: Yes correct - if the user navigates out of the flow at any point, they lose all their answers. 
(this is the established pattern on payment links but easily changed if we want)

christiandutton-mt (approved):
LGTM!

PR #5024, opened 2026-09-04
PP-15751 Add cypress tests for service director skeleton flow

christiandutton-mt:
## What

[PP-15751](https://payments-platform.atlassian.net/browse/PP-15751)

- Amends email address to 'work email address' on details page as per updated design.
- Amend back link in address controller to accept credentialID to fix path.
- Adds cypress test for each of three pages in service director flow. 

**For now, as these are skeleton views, the check your answers page has hardcoded values which I have not tested against.** 

**The back links currently all have the functionality of going back to the previous page in the flow (so for example if you are in the address view and click back, you will always go to the details view) - in future this will be updated to work with the change buttons in the check your answers view. So if you had come from the check your answers view through the change links, pressing back in the address view takes you back to the check your answers view rather than to details. This has not been implemented yet as per this PR so has not been tested.**

[PP-15751]: https://payments-platform.atlassian.net/browse/PP-15751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ralphm10 (approved):
Nice work

PR #5026, opened 2026-09-07
PP-15549 Add Adyen account setup PATCH method

DomBelcher:
## What

PP-15549 Add Adyen account setup PATCH method

- Add method to ConnectorClient.class.ts to update Adyen account setup via PATCH request
- Add AdyenAccountSetupUpdateRequest.class.ts class to ensure correctness of request


ralphm10 (approved):
LGTM

PR #5031, opened 2026-09-09
PP-15751 Add persistence to answers and back link flow

christiandutton-mt:
## What

[PP-15751](https://payments-platform.atlassian.net/browse/PP-15751)

This PR completes the work on PP-15751 by adding persistence to answers entered throughout the flow so that they appear on the check your answers page. Also added more complicated back button logic, so that if the user clicks change on the check your answers page to return to the address/details page, pressing back would take them back to the check your answers page, not the previous page in the flow. 

- Added session data and passed through controllers.
- Added back button logic through params and passed through controllers.
- Updated unit tests. 
- Updated check-your-answers.njk to display session data.
- Updated cypress tests

[PP-15751]: https://payments-platform.atlassian.net/browse/PP-15751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/controllers/simplified-account/settings/adyen-details/service-director/constants.ts
+++ b/src/controllers/simplified-account/settings/adyen-details/service-director/constants.ts
@@ -1,0 +1,39 @@
+import { ServiceRequest } from '@utils/types/express'
+import lodash from 'lodash'
+
+const CREATE_SESSION_KEY = 'session.pageData.serviceDirector'
+
+export class ServiceDirectorSession {
+  firstName?: string
+  lastName?: string
+  dobDay?: string
+  dobMonth?: string
+  dobYear?: string
+  email?: string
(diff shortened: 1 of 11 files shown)


ralphm10 (commented):
Looks good - the only thing I think to add is to clear the session when navigating back to the task list (as with [responsible person](https://github.com/govuk-pay/pay-selfservice/blob/d6eede3b6efb7a8af0294e5c6d105e8ed58148f2/src/controllers/simplified-account/settings/switch-psp/switch-to-adyen/switch-to-adyen.controller.ts#L8) to make consistent)

ralphm10 (approved):
Good work

PR #5033, opened 2026-09-09
PP-15861 Make provider change to page visible to all Stripe services

ralphm10:
## What

PP-15861

- Make provider change to page visible to all Stripe services
- Make navigation option visible to all Stripe services
- Update Cypress tests to verify the change


DomBelcher (approved):
LGTM

PR #5036, opened 2026-09-11
PP-15877 Use data from adyen setup in connector to determine task status shown in selfservice

ralphm10:
## What

PP-15877

- Update switch to Adyen controller to call Adyen setup service
- Add connector call in Adyen setup service to retrieve task status
- Update Adyen tasks class to use this status, where previously it was hardcoded
- Display the dynamic status in the switch to Adyen view

### How

- navigate to Settings -> Payment provider -> Switch provider to Adyen now, from a service in the correct state for switching
  - this can be setup by running `pay local service <email>` from the latest version on [Pay CLI](https://github.com/govuk-pay/pay-cli), where `email` is the associated with a user you have created locally, and login to self-service with
- On the `switch-to-adyen` page, all tasks (eg. `Service director`) should have a status of `Not yet started`
- To update a task, make a PATCH request to `/v1/api/service/:serviceId/account/live/adyen-setup/:credentialExternalId` of a locally running connector istance with a body of: 
```
[
  {
    "op": "replace",
    "path": "responsible_person",
    "value": "COMPLETED"
  }
]
```
- The task status should update when you refresh your browser

[
Screenshot 2026-09-11 at 15 14 48
](url)

### DIFF
--- a/src/services/adyen-setup.service.ts
+++ b/src/services/adyen-setup.service.ts
@@ -1,3 +1,19 @@
+import ConnectorClient from '@services/clients/pay/ConnectorClient.class'
+
+const connectorClient = new ConnectorClient()
+
+const getConnectorAdyenAccountSetup = async (serviceExternalId: string, accountType: string, credentialExternalId: string) => {
+  return connectorClient.gatewayAccounts.adyenSetup.get(
+    serviceExternalId,
+    accountType,
+    credentialExternalId
+  )
+}
+
 export async function markTaskAsComplete() {
   // call client method
+}
+
+export {
+  getConnectorAdyenAccountSetup
 }
--- a/src/services/adyen-setup.service.ts
+++ b/src/services/adyen-setup.service.ts
@@ -2,18 +2,16 @@ import ConnectorClient from '@services/clients/pay/ConnectorClient.class'
 
 const connectorClient = new ConnectorClient()
 
-const getConnectorAdyenAccountSetup = async (serviceExternalId: string, accountType: string, credentialExternalId: string) => {
-  return connectorClient.gatewayAccounts.adyenSetup.get(
-    serviceExternalId,
-    accountType,
-    credentialExternalId
-  )
+const getConnectorAdyenAccountSetup = async (
+  serviceExternalId: string,
+  accountType: string,
+  credentialExternalId: string
+) => {
+  return connectorClient.gatewayAccounts.adyenSetup.get(serviceExternalId, accountType, credentialExternalId)
 }
 
 export async function markTaskAsComplete() {
   // call client method
 }
 
-export {
-  getConnectorAdyenAccountSetup
-}
+export { getConnectorAdyenAccountSetup }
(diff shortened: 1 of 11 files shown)


test/cypress/integration/simplified-account/service-settings/adyen-details/responsible-person/address.cy.ts
```diff
@@ -38,12 +39,23 @@ const serviceFixture = new ServiceFixture({
 })
 const userFixture = UserFixture.asServiceAdmin([serviceFixture], { externalId: USER_EXTERNAL_ID })
 
+const adyenAccountSetup = AdyenAccountSetupFixture.Completed({
```
DomBelcher: Very minor but maybe the setup tasks should be `NOT_STARTED` for these tests, since tasks shouldn't be repeatable - and eventually this should be enforced so these tests may fail with this state

DomBelcher (commented):
Nice work, just one suggestion

test/cypress/integration/simplified-account/service-settings/adyen-details/responsible-person/address.cy.ts
ralphm10: good shout - updated

src/services/adyen-setup.service.ts
```diff
@@ -19,3 +27,5 @@ export const markTaskAsComplete = async (
     updateRequest
   )
 }
+
+export { getConnectorAdyenAccountSetup }
```
DomBelcher: I dunno what pattern is preferable of `export const ....` or `export { ... }`, and we're not consistent across the project, but we should probably be consistent within files.

I know this is because a separate PR was merged using the other pattern, and I don't mind which you choose to use here 🙂

test/cypress/integration/simplified-account/service-settings/adyen-details/responsible-person/check-your-answers.cy.ts
```diff
@@ -39,12 +40,23 @@ const serviceFixture = new ServiceFixture({
 })
 const userFixture = UserFixture.asServiceAdmin([serviceFixture], { externalId: USER_EXTERNAL_ID })
 
+const adyenAccountSetup = AdyenAccountSetupFixture.Completed({
```
DomBelcher: Sorry, I probably should have highlighted this in all the Cypress test files - same comment applies, I think it's better for them to be `NOT_STARTED` unless there's a reason otherwise

ralphm10: no I should have checked - updated now

src/services/adyen-setup.service.ts
ralphm10: we seem to use `export { ... }` in the services so have gone with that

DomBelcher (approved):
Nice work

PR #5039, opened 2026-09-15
PP-15875: Mark Adyen task complete service file

dan-searle-mt:
## What

PP-**[15875]**

This ticket is to create an Adyen service file, in order to allow a task to be marked as complete.

> [!NOTE]
>This isn't a change to the UI, it is just adding in the service function to mark tasks as completed.


src/services/adyen-setup.service.ts
```diff
@@ -1,3 +1,28 @@
...
+
+const connectorClient = new ConnectorClient()
+
+/**
+ * Function to call the connector client, to update the task status.
+ * @param serviceExternalId
+ * @param accountType
+ * @param credentialExternalId
+ * @param task
+ */
```
DomBelcher: You can omit the JSDoc, it's still lingering in places from before we started using Typescript, but the TS types should be sufficient

src/services/adyen-setup.service.test.js
```diff
@@ -0,0 +1,119 @@
...
+const expect = chai.expect
+
+const SERVICE_EXTERNAL_ID = 'service123abc'
+const ACCOUNT_TYPE = 'live'
+const CREDENTIAL_EXTERNAL_ID = 'adyen-credential-123-abc'
+
+describe('adyen setup service', function () {
+  describe('markTaskAsComplete', function () {
+    it('should patch the given task to COMPLETED using the connector client', async function () {
+      const patchStub = sinon.stub().resolves()
```
DomBelcher: If you want this to work nicely as Typescript, you can do 
```suggestion
      const patchStub = sinon.stub<[string, string, string, AdyenAccountSetupUpdateRequest]>().resolves()
```

src/services/adyen-setup.service.test.js
```diff
@@ -0,0 +1,119 @@
...
+        return {
+          gatewayAccounts: {
+            adyenSetup: {
+              patch: patchStub,
+            },
+          },
+        }
+      }
+
+      const adyenSetupService = proxyquire('./adyen-setup.service', {
```
DomBelcher: And here if you do
```
import * as AdyenSetupService from './adyen-setup.service'
```
then 
```suggestion
      const adyenSetupService = proxyquire<typeof AdyenSetupService>('./adyen-setup.service', {
```
the types should work properly

src/services/adyen-setup.service.test.js
```diff
@@ -0,0 +1,119 @@
...
+        {
+          op: 'replace',
+          path: 'organisation_details',
+          value: 'COMPLETED',
+        },
+      ])
+    })
+
+    it('should map each task name to its connector path', async function () {
+      const cases = [
```
DomBelcher: And here:
```suggestion
      const cases: [AdyenAccountSetupTaskName, string][] = [
```

DomBelcher (commented):
Looks good, just some suggestions to make the types work nicely

DomBelcher (approved):
LGTM

PR #5047, opened 2026-09-18
PP-15820 Adyen setup - add patch call for legal terms

iqbalgds:
## PP-15820 

PP-**15820**
- Add code to make the patch request to connector to set the task as complete when the user has clicked `accept terms and conditions`.

### How
- Make sure the tests pass.


DomBelcher (approved):
LGTM

